PR #30, opened 2016-08-22
ADD check for unknown parameters in scenario file

mfeurer:


mfeurer:
@mlindauer @KEggensperger this is a first step towards generating a documentation for all options in the scenario file. It allows all kind of different spellings of a parameter (lower and upper case, with minus, with underscore, or without both), checks whether mandatory parameters are given and complains if it can't parse a parameter (prevents typos). What do you think?

mlindauer:
Ok, I changed some minor things (syntax, meta info). Imho, it can be merged into dev.

PR #43, opened 2016-09-01
FIX #40

### LINKED ISSUE #40: runcount-limit ignored
The value set for *runcount-limit* in scenario is ignored. It seems like SMAC checks the runcount value only after each *iteration* and not after each target algorithm run, therefore SMAC might do more runs than it is supposed to do.

mlindauer:
Could you (@mfeurer and @KEggensperger) please review my PR to fix #40.

### DIFF
--- a/examples/rf.py
+++ b/examples/rf.py
@@ -10,7 +10,7 @@ cmd_folder = os.path.realpath(
     os.path.abspath(os.path.split(inspect.getfile(inspect.currentframe()))[0]))
 cmd_folder = os.path.realpath(os.path.join(cmd_folder, ".."))
 if cmd_folder not in sys.path:
-    sys.path.append(cmd_folder)
+    sys.path.insert(0,cmd_folder)
 
 import numpy as np
 from sklearn.cross_validation import KFold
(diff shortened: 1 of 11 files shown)


mfeurer:
I think the budget exhausted functions is a good idea. Nevertheless, I think this PR makes the code harder to read and adds responsibility for checking budget to a place where it does not belong, namely the target algorithm (evaluator). This introduces a new check in line 239. Couldn't we instead just check the budget here? This would change the intensify method as much as this PR does, but we don't need to change the target algorithm (evaluator).

Also, I'm not sure what the changes to the examples, the RF and the runhistory are doing in this PR.

mlindauer:
I would not like to implement the budget check in the intensification, because the target algorithms are called already twice in intensification and furthermore also in the initial design. If we will have a more complex initial design in the future (I have plans for this), a budget check is also really important there. Therefore, I believe it is the better solution to do the budget check in the target algorithm class (where we have to implement it only once).

The change of the runhistory was necessary because the TA returns a ABORT run if the budget is exhausted and we don't want to aggregate the performance of ABORT runs. I will add a comment in the runhistory class.

The RF change was simply due to a annoying warning of numpy which I wanted to fix. 

Cheers,
Marius

mfeurer:
Okay, +1.

Issue #39, opened 2016-08-30
Output and trajectory files don't match


KEggensperger:
When running rosenbrock.py I get the output shown below. The output suggests that there are at least 4 incumbents (not even listing the initial incumbent), but the trajectory file only lists three (including the incumbent). Which one is correct?

```
Default Value: 120048.00
/home/eggenspk/virtualEnvs/virtual3SMAC/lib/python3.4/site-packages/smac-0.1.0-py3.4.egg/smac/smbo/acquisition.py:173: RuntimeWarning: divide by zero encountered in true_divide
INFO:intensifier:Challenger (69492.7391) is better than incumbent (120048.0000) on 1 runs.
INFO:intensifier:Changing incumbent to challenger: Configuration:
  x1, Value: -4.656907303634218
  x2, Value: -3
  x3, Value: 2
  x4, Value: -2

INFO:intensifier:Challenger (677.3422) is better than incumbent (69492.7391) on 1 runs.
INFO:intensifier:Changing incumbent to challenger: Configuration:
  x1, Value: -0.8297799529742598
  x2, Value: 2
  x3, Value: 2
  x4, Value: 3

INFO:intensifier:Updated estimated performance of incumbent on 1 runs: 677.3422
INFO:intensifier:Updated estimated performance of incumbent on 1 runs: 677.3422
INFO:intensifier:Updated estimated performance of incumbent on 1 runs: 677.3422
INFO:intensifier:Challenger (511.8286) is better than incumbent (677.3422) on 1 runs.
INFO:intensifier:Changing incumbent to challenger: Configuration:
  x1, Value: -0.23478815616176085
  x2, Value: 0
  x3, Value: -2
  x4, Value: 5

INFO:intensifier:Challenger (410.7729) is better than incumbent (511.8286) on 1 runs.
INFO:intensifier:Changing incumbent to challenger: Configuration:
  x1, Value: 0.17154408126871612
  x2, Value: 0
  x3, Value: -2
  x4, Value: 4

INFO:intensifier:Updated estimated performance of incumbent on 1 runs: 410.7729
INFO:intensifier:Updated estimated performance of incumbent on 1 runs: 410.7729
INFO:intensifier:Updated estimated performance of incumbent on 1 runs: 410.7729
INFO:intensifier:Updated estimated performance of incumbent on 1 runs: 410.7729
INFO:intensifier:Updated estimated performance of incumbent on 1 runs: 410.7729
INFO:Stats:##########################################################
[...]

Issue #38, opened 2016-08-30
Performance of initial incumbent


KEggensperger:
The trajectory (file) always starts with the first evaluated configuration (which is the default) with a cost of 999999999, but it never updates or adds a new entry with the real initial incumbent performance after evaluating it.

Example from rosenbrock.py:

```
"CPU Time Used","Estimated Training Performance","Wallclock Time","Incumbent ID","Automatic Configurator (CPU) Time","Configuration..."
0.004036, 999999999.000000, 0.012154, 1, 0.008118, x1='5.0', x2='5', x3='5', x4='5'
0.034413, 677.342200, 0.387002, 3, 0.352589, x1='-0.8297799529742598', x2='2', x3='2', x4='3'
0.210829, 410.772936, 2.789726, 5, 2.578898, x1='0.17154408126871612', x2='0', x3='-2', x4='4'
```

mlindauer:
Easy to fix. 
Should I open a PR?

KEggensperger:
Yes and thanks.

PR #45, opened 2016-09-06
Fix/#38/#39

mlindauer:
Fixing #38 and #39.
@mfeurer , @KEggensperger  please review

### DIFF
--- a/smac/smbo/smbo.py
+++ b/smac/smbo/smbo.py
@@ -319,17 +319,17 @@ class SMBO(BaseSolver):
                 self.stats.get_remaining_ta_budget(),
                 self.stats.get_remaining_ta_runs()))
 
-            if self.stats.get_remaing_time_budget() < 0 or \
-                    self.stats.get_remaining_ta_budget() < 0 or \
-                    self.stats.get_remaining_ta_runs() < 0:
-                break
-
             if self.stats.inc_changed > inc_id:
                 self.trajLogger.add_entry(train_perf=inc_perf,
                                           incumbent_id=self.stats.inc_changed,
                                           incumbent=self.incumbent)
                 inc_id = self.stats.inc_changed
-                
+
+            if self.stats.get_remaing_time_budget() < 0 or \
+                    self.stats.get_remaining_ta_budget() < 0 or \
+                    self.stats.get_remaining_ta_runs() < 0:
+                break
+
             self.stats.print_stats(debug_out=True)
 
         return self.incumbent
--- a/smac/smbo/smbo.py
+++ b/smac/smbo/smbo.py
@@ -100,7 +100,7 @@ class SMBO(BaseSolver):
             self.stats = stats
         else:
             self.stats = Stats(scenario)
-
+            
         self.runhistory = RunHistory()
 
         self.logger = logging.getLogger("smbo")
@@ -120,6 +120,8 @@ class SMBO(BaseSolver):
 
         self.scenario = scenario
         self.config_space = scenario.cs
+        self.traj_logger = TrajLogger(
+            output_dir=self.scenario.output_dir, stats=self.stats)
 
         self.types = get_types(self.config_space, scenario.feature_array)
         if model is None:
@@ -149,6 +151,7 @@ class SMBO(BaseSolver):
 
         self.inten = Intensifier(executor=self.executor,
                                  stats=self.stats,
+                                 traj_logger = self.traj_logger,
                                  instances=self.scenario.train_insts,
                                  cutoff=self.scenario.cutoff,
                                  deterministic=self.scenario.deterministic,
@@ -196,9 +199,6 @@ class SMBO(BaseSolver):
         else:
             raise ValueError('Unknown run objective: %s. Should be either '
                              'quality or runtime.' % self.scenario.run_obj)
-
-        self.trajLogger = TrajLogger(
-            output_dir=self.scenario.output_dir, stats=self.stats)
 
     def run_initial_design(self):
         '''
@@ -248,7 +248,7 @@ class SMBO(BaseSolver):
 
         self.stats.inc_changed += 1  # first incumbent
 
-        self.trajLogger.add_entry(train_perf=default_perf,
+        self.traj_logger.add_entry(train_perf=default_perf,
                                   incumbent_id=self.stats.inc_changed,
                                   incumbent=self.incumbent)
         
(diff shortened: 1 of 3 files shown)


KEggensperger:
When running rosenbrock I still don't get the true initial incumbent performance (default performance=120048.00). I think the first two entries should look like that:

```
{"evaluations": 0, "cost": 999999999, "cpu_time": 0, "total_cpu_time": null, "wallclock_time": 0, "incumbent": ["x1='5.0'", "x2='5'", "x3='5'", "x4='5'"]}
{"evaluations": 1, "cost":120048.00 , "cpu_time": X, "total_cpu_time": null, "wallclock_time": X, "incumbent": ["x1='5.0'", "x2='5'", "x3='5'", "x4='5'"]}
```

I would insert the first entry when initializing SMBO and the second after the initial design. What do you think?

smac/smbo/smbo.py
```diff
@@ -120,6 +120,8 @@ def __init__(self, scenario, tae_runner=None, acquisition_function=None,
 
         self.scenario = scenario
         self.config_space = scenario.cs
+        self.traj_logger = TrajLogger(
+            output_dir=self.scenario.output_dir, stats=self.stats)
```
KEggensperger: Do the first entry to the trajectory file here

KEggensperger:
This PR does not fix #39, right?

mlindauer:
Hi,

if I'm running the rosenbrock example, I get the following output in `traj_aclib2.json`

```
{"cost": 120048.0, "incumbent": ["x1='5.0'", "x2='5'", "x3='5'", "x4='5'"], "evaluations": 1, "cpu_time": 0.0015752315521240234, "total_cpu_time": null, "wallclock_time": 0.01628422737121582}
{"cost": 69492.73909731719, "incumbent": ["x1='-4.656907303634218'", "x2='-3'", "x3='2'", "x4='-2'"], "evaluations": 2, "cpu_time": 0.003468036651611328, "total_cpu_time": null, "wallclock_time": 0.20449256896972656}
```

Why do you still get the erroneous output?

> I would insert the first entry when initializing SMBO

No. Maybe we want to use an initial design that does not start with the default configuration. So, the first entry should be in the initial design.

mlindauer:
> This PR does not fix #39, right?

Yes, it does.

KEggensperger:
Yes you are right.
But now there is no entry at time zero, so we don't have an incumbent in case the first configuration (during the initial design) runs longer than the threshold?
@mfeurer: Is that what we want? SMAC2 works differently, doesn't it?

mfeurer:
1. we should decide whether we add something to the trajectory at timestep zero. It would make sense to only add a configuration which actually ran. As @KEggensperger mentioned, this would break backwards compability
2. Did the original SMAC update the trajectory every time it ran the incumbent on a new instance? If yes, there should be a new entry to the trajectory in line 172.

mlindauer:
> As @KEggensperger mentioned, this would break backwards compability

I don't see how this breaks backwards compability. smac-validate still works on the trajectory files of SMAC3. Could you please explain it.

> Did the original SMAC update the trajectory every time it ran the incumbent on a new instance? 

I don't know whether SMAC2 updates the trajectory file after each new incumbent run, but indeed, the same incumbent is listed more than once in the SMAC2's trajectory files with updated performance values.
The question is whether we also want to have the same behavior in SMAC3. As above mentioned, I don't see how the new implementation would break anything.
Furthermore, I like to avoid too much I/O accesses if we have really fast algorithm runs.
Last but not least, if we have no (or only one) instance, the estimated performance is already accurate after one run. If we have real algorithm configuration with a lot of instances, the estimated training performance is hard to use and not comparable with other SMAC runs, because each SMAC run used another instance subset to estimate the training performance.

mlindauer:
No replies?

mfeurer:
Yes, the workers would still run. But wouldn't the semantics of the trajectory file change? If there is no reason to change the semantics of the trajectory file, I would SMAC3 to behave as similar to SMAC2 as possible.

Regarding IO. In a scenario with short runs, how many instances do you normally have and how long does a short run take? Also, how often does the incumbent usually change?

mlindauer:
> But wouldn't the semantics of the trajectory file change?

Is their a defined semantic of the trajectory file?
We use it normally only to get the incumbent at time step t
since the estimated training performance is incomparable anyway.
Do you use the estimated training performance in the trajectory file?

> Regarding IO. In a scenario with short runs, how many instances do you normally have and how long does a short run take?

There is no normal case for us. But as an example Spear on SWGCP: 1000 instances (plus spear is a  stochastic algorithm such that we can multiply it by infinite number of random seeds). A short run can be less than 0.1 seconds.

Cheers,
Marius

mfeurer:
The only reference I know is a preliminary ACLIB specification, which reads like one should update every time the estimated performance changes:
    
    The configurator has to output a trajectory file in the following format:
    <cpu time used>, <wallclock time>, <number of evaluations>, <estimated performance>, <incumbent configuration>

mlindauer:
It's funny that you argue with the AClib specification I wrote.
I don't know which version you have. But at least the current version has a footnote.

`The configurator has to output a trajectory file\footnote{A trajectory file consists of one line per new incumbent found by the configurator.} in the following JSON format`

There is later a further note what to do at the end of the configuration process:

`TIP: At the end of configuration, an entry in the trajectory file should be outputted past or at the limit of configuration, so that it is clear from examination of the file that the procedure completed successfully (e.g., there is an entry where the cpu time limit exceeded the requested limit).`

mfeurer:
It's the only definition of trajectory I know of. I'm referring to a draft from 27.10.2014. There is no newer specification on the ACLIB website, so I assumed this is still the current version.

mlindauer:
There was never an official/final version of this specification and even SMAC2 does not fully implements it. But here is an internal link:

https://bitbucket.org/fawcettc/aclib/src/4012624be3f445910f2779d4ecc6417b468ab00a/ac_format/?at=aclib-2.0

mfeurer:
Okay, I will have a look at this next week.

mfeurer:
Okay, that sound very reasonable and I'm okay with this. The only question for me is now whether we want to assume that the first entry is at time 0.0, or whether we want to change this. Since we may have different initial designs is the future, how about dropping this? @KEggensperger would this break anything on your side?

KEggensperger:
Approved

mlindauer:
Hi,

as discussed today, I added an entry at the beginning of the initial design such that we have something without any runs. The time spent for  the algorithms is 0 at that moment, however the wallclock time of SMAC is of course not exactly 0 (but very small).

Can we merge now?

Best,
Marius

mfeurer:
Approved

PR #47, opened 2016-09-14
FIX #46, also fixes the following issues:

### LINKED ISSUE #46: Allow to return additional run information from a python function target algorithm
instead of returning an empty dictionary right now. How about:

    rval = obj(config)
    if isinstance(rval, tuple):
        value = rval[0]
        additional_run_info  rval[1]
    else:
        value = rval

mfeurer:
1. limit wallclock time instead of cpu time. I was not able to write
   a unit test which limits cpu time, and I don't see a reason why
   one would only want to limit cpu time
2. allow to enforce a memory limit
3. consider the algorithm run to have failed if the return value is
   None
4. add unittests

### DIFF
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -120,6 +120,7 @@ class Scenario(object):
         self.add_argument(name='overall_obj', help=None, default='par10')
         self.add_argument(name='cutoff_time', help=None, default=999999999,
                           dest='cutoff', callback=lambda arg: float(arg))
+        self.add_argument(name='memory_limit', help=None)
         self.add_argument(name='tuner-timeout', help=None, default=numpy.inf,
                           dest='algo_runs_timelimit',
                           callback=lambda arg: float(arg))
--- a/smac/smbo/smbo.py
+++ b/smac/smbo/smbo.py
@@ -149,10 +149,6 @@ class SMBO(BaseSolver):
                                             par_factor=scenario.par_factor)
         else:
             self.executor = tae_runner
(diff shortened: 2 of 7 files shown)


mfeurer:
@mlindauer @KEggensperger what do you think?

Besides that, I'm confused where the conflict comes from.

mlindauer:
Hi,

The memory limit will not work. 
After applying the PR for #40, we added the new function `start` in ExecuteTARun. The `start` function calls the `run` function of the heirs of ExecuteTARun (i.e., `start` wraps `run`). The `start` function is indeed important since it handles logging and entries in the stats. 
However, `start` does not have an argument for the memory limit and I'm also hesitant to add this argument in `start` since a memory limit is not supported by the other TA heirs.

Cheers,
Marius

mfeurer:
Then I should maybe remove it from this PR and add it to all TAs?

mlindauer:
How do you want to add it all TAs? Unfortunately, the memlimit of the pynisher is not working for the other TAs.

mfeurer:
And the others don't assume that the runsolver is installed, right?

mlindauer:
yes

mfeurer:
Okay, I found the source of the merge conflict. I forgot to pull the latest changes from the development branch. I did a rebase of my two commits.

mfeurer:
@mlindauer okay, I made the relevant updates. Tests are working again after @KEggensperger updated the travis-ci configuration file.

KEggensperger:
Approved

PR #51, opened 2016-09-16
FIX #50

### LINKED ISSUE #50: Variance prediction across Instances should not take instance hardness distribution into account
Right now, our variance prediction marginalized across instances (used for EI) also takes into account the instance hardness distribution (by using the law of total variance). However, as discussed with Stefan and Katha, this is actually wrong here. 
We only want to use the mean of the variance (without the variance of the means).

I will create a PR for this.

mlindauer:
FIX for #50

### DIFF
--- a/smac/epm/rf_with_instances.py
+++ b/smac/epm/rf_with_instances.py
@@ -204,7 +204,9 @@ class RandomForestWithInstances(object):
             X_ = np.hstack(
                 (np.tile(x, (n_instances, 1)), self.instance_features))
             means, vars = self.predict(X_)
-            var_x = np.mean(vars) + np.var(means)
+            # use only mean of variance and not the variance of the mean here
+            # since we don't want to reason about the instance hardness distribution
+            var_x = np.mean(vars) # + np.var(means)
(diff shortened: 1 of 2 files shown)


mfeurer:
APPROVED

mfeurer:
As mentioned in the related issue, I think this is the correct way of computing variances and this PR makes sense. We can merge this when all tests on travis work again.

mlindauer:
ok, unit test should be fixed again

mfeurer:
I now used the github approval feature. Not sure if we still need pullapprove?

KEggensperger:
Yes, we do still need pullapprove if we want approvals to be mandatory. The github approval is optional for non protected branches, like our development branch.  See here: https://help.github.com/articles/enabling-required-reviews-for-pull-requests/.

Is there any other advantage of github approve over pullapprove? If not I would propose to continue using pullapprove.

KEggensperger:
approved

mlindauer:
ok, I will merge this PR and you can discuss the use of pullapprove offline, ok?

PR #54, opened 2016-09-19
Documentation

AndreBiedenkapp:
I added another example  to the already existing ones.
I added the first documentation.
It contains a landing page, installation instructions, a manual that describes how to use SMAC directly with python or as before, the license and contact information.


mlindauer:
Hi,

I made a brief pass on the documentation. I think it is a very good starting point.
Before merging it to the development branch, here my wishlist:

- to use it the github pages, the html-files should live in docs/ (in the master branch later)
- could we move all documentation files into a separated folder? For non-Python-experts, folders such as "source" or files such as "Makefile" looks like the real stuff and not as docu files.

Cheers,
Marius

mfeurer:
Hi,

I think neither the development branch nor the master branch should contain any form of built files, which include documentation html. Documentation should be either built by the user on her system or by us using a CI-like service like circle-ci or readthedocs.org (although the latter one has issues and I wouldn't use it any more for projects involving packages like numpy).

If we want to use github pages, the documentation should live in a separate branch called gh-pages. You can see an example in the auto-sklearn repository. It uses circle-ci to rebuild the documentation each time a user pushes to the master or development branch.

Let me know if I can help here.

Cheers,
Matthias

mlindauer:
> Documentation should be either built by the user on her system

I don't know why the user should have the burden to build the documentation!? 

> If we want to use github pages, the documentation should live in a separate branch called gh-pages

github pages can now also in live in other places; from the github docu:

`You can configure GitHub Pages to publish your site's source files from master, gh-pages, or a /docs folder on your master branch for Project Pages and other Pages sites that meet certain criteria.`

> by us using a CI-like service like circle-ci 

If automation of the process of the documentation building would be indeed nice. Maybe you can show Andre how to setup this up (or give him some pointers where he should look).

mfeurer:
I didn't mean the user should build the documentation, I just stated that either she or we are responsible. Having a service which does that would make it easy for us to do so. Still, I'm opposed to having documentation in the master branch as it would clutter both pull requests and the commit history (if we use a service like circle-ci).

@AndreBiedenkapp you can just drop by and I can explain you the example in auto-sklearn.

mlindauer:
My point is simply that if I as a user download a package (or clone a repo), I would like to have a chance to use it offline (including the documentation without any necessary packages or tools to build the documentation first).

Is this possible with your workflow?

mfeurer:
In my opinion, if you clone a repo you should build the documentation yourself (or clone the gh-pages, which are produced by the proposed workflow). If you download a package from pypi, it usually doesn't contain any documentation (at least I wouldn't know of any example which does). Also, I'm not aware of any major python packages which has the built documentation in master/development branch.

There are certainly ways to build the documentation as a PDF for offline usage and the workflow would allow to produce these, but I wouldn't know where to store them.

Issue #48, opened 2016-09-14
Automatically set stats object in TAE when passing it to SMAC


mfeurer:
Would make this code easier:
```
stats = Stats()
SMBO(tae_runner=ExecuteTAFunc(lambda: 1, stats=stats),
     scenario=self.scenario,
     stats=stats)
```

mlindauer:
sorry, I don't get what you want here.

mfeurer:
I need to create a stats object and pass it to the TAE and the SMBO object. Instead, it would be great if the SMBO object could inject the stats object it creates into the TAE. Is there any use case where the user wants to create the stats object?

mlindauer:
right now, the stats argument of smbo is optional. if not set, smbo will generate a stats object. 

If the argument of the tae is not given, smbo will generate a tae object and will pass the stats object to the tae. 

So, you would like to keep both arguments, but smbo should always update the stats object in the tae?

mfeurer:
Sorry, I wasn't clear about what I propose. I propose to remove both stats arguments, and add a private function to the TAE which is called by SMBO and injects the stats. The SMBO object creates a stats object. I currently don't see a reason why a user wants to provide a custom stats object.

mlindauer:
I agree that the SMBO object could inject the stats object to the TAE.
However, why should this a private method? By definition, I would expect that a private method cannot be called by another object.

Please create a PR and we can discuss the details. OK?

mfeurer:
Okay, will do so once #47 is merged.

mfeurer:
Would be covered by #55

mfeurer:
I'll check that again on Tuesday.

mfeurer:
Covered by #55, now also tested. I'm closing this now.

PR #60, opened 2016-09-22
Fix#55

mlindauer:
Hi,

restructured the SMAC code; the main changes is that I added facades to abstract away some complexity. Right now, I have two facades

* standard SMAC over command line
* a scipy similar facade (see examples/rosenbrock2.py for an example to use this new facade)

@mfeurer and @KEggensperger : could you two please have a look at the changes and whether you agree on them.

I'm aware that a lot of unit tests are probably broken right now.  I would like to have first some feedback on the changes before I start to fix the unit tests.

Best,
Marius


mlindauer:
two further comments/change requests

* do we need max_iter as part of the optimize function? right now every budget lives in the stats object and is defined in the scenario. furthermore, as far as I know nobody uses the number of iterations
* why is the aggregation function (previously objective) an argument of the intensifiy method? Do we call the method with different aggregation functions? or should this be an argument of the constructor?

mfeurer:
Hi,

I think this PR is a good starting point to discuss a major refactoring of the SMAC codebase. I like most of the ideas, but propose that we do the whole refactoring in small steps. Making only small changes to the codebase reduces the danger of adding bugs by mistake, makes reviewing the code easier and only 'blocks' small parts of the code base. I propose that you split this PR into multiple small PRs:

1. Creation of a SMAC object as in c880cc20dab5a86065de2ce57ed0b4f15a755834
2. Add a module for the initial design
3. Add a module for intensification
4. Add an interface which makes SMAC look like scipy.optimize

and we add the changes one by one. What do you think?

And now the answers to your questions:

1. No, I don't think we need the maximum number of iterations. If we needed that feature, it should at least live in the scenario file. So far it confused everybody I talked to.
2. I thought that being able to have different aggregation functions would help make SMAC more flexible, for example one could only aggregate a subset of instances. So far I didn't use it and think it's obsolete (especially given that one would have to change a lot in the rest of intensify to make it work).

mlindauer:
Hi,

> I propose that you split this PR into multiple small PRs:

I see your point of avoiding newly introduced bugs. However, we are already awfully slow in changing SMAC. If we split the PR and the refactoring in small pieces, it will cost us even more momentum. Excluding the weekend, I waited 3 days to get feedback. If I wait for feedback for each mini PR, the refactoring of SMAC will not be done by xmas. 
Furthermore, it already costs me a lot of time to continues with SMAC after a break for several days. So my time is also not used optimally. 

Therefore, I vote for more larger PRs and not for more smaller PRs.

Best,
Marius

mfeurer:
Okay, if you think so. Then give me some time to work through the PR.

mfeurer:
General:

* I like the style of the new SMAC object. It's simple and does only a single thing.

Here are some suggestions on what I think should be changed in this PR on an abstract level:

* We should not use the name facade in any package that will be accessed by the user. I think we should add the SMAC class into the smbo package.
* I don't think the class FuncSMAC is what we're aiming for. It's already easier to use than the full SMAC class with all it's options, but we could make it look even more like scipy.optimize.fmin() (http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin.html#scipy.optimize.fmin). For the beginning, what about making it a function and put it into a module smac.optimize and call the function fmin()?
* I think the name `DefaultDesign` is ambiguous. As a novice user I would think this is the initial design which is run by default, and not the initial design which runs a default configuration.

And some more in-depth comments:

* We don't need `from __future__ import` any more
* The abstract initial design should raise `NotImplementedError()` instead of `raise("Not implemented")`
* typo in runhistory.py (stoar)
* in SMBO, why is the optimization routine for the acquisition function called `acp_optimizer`? What does `acp` stand for?
* the costs for a configuration are no longer updated in the run history? Is this on purpose? And what mechanism makes this obsolete?

These are some general suggestions of mine. I will have a close look once we have agreed on what the package should look like in the end.

mlindauer:
Hi,

Thank you for the thorough feedback.

> We should not use the name facade in any package that will be accessed by the user. I think we should add the SMAC class into the smbo package.

I don't know whether the package name "facade" is a problem. Imho, this name is very descriptive. I don't agree that SMAC should live in the smbo package because SMAC is more than SMBO (e.g., additionally intensification)

 > I don't think the class FuncSMAC is what we're aiming for. It's already easier to use than the full SMAC class with all it's options, but we could make it look even more like scipy.optimize.fmin() 

I designed FuncSMAC to be similar to scipy.optimize.fmin, except that I like classes more than functions. I'm fine with to change it in a future PR again.

> I think the name DefaultDesign is ambiguous. As a novice user I would think this is the initial design which is run by default, and not the initial design which runs a default configuration.

agreed and fixed

> We don't need from __future__ import any more

fixed

> The abstract initial design should raise NotImplementedError() instead of raise("Not implemented")

fixed 

> typo in runhistory.py (stoar)

fixed

> in SMBO, why is the optimization routine for the acquisition function called acp_optimizer? What does acp stand for?

it should be acq for acquisition function. fixed

> the costs for a configuration are no longer updated in the run history? Is this on purpose? And what mechanism makes this obsolete?

The costs of a configuration is updated now always after a new run was added to the runhistory; far more consistent and we cannot forget it in other parts of the code.

If there are no further change requests for the PR, I would start to fix the unit tests and examples. 
@KEggensperger Do you agree with the changes?

Best,
Marius

mlindauer:
Hi, 

I fixed the unit tests and a bug in  the facade.
Before merging, I would like to remove the `cs` argument in the `RFRImputator` constructor, if you agree. Of course, if you insist, I will create a separate PR for this issue.

Best,
Marius

mfeurer:
I will not be able to review code for imputation. The rest of the code looks good to me (and fixes #48).

smac/tae/execute_func.py
```diff
@@ -29,6 +29,25 @@ class ExecuteTAFunc(ExecuteTARun):
     par_factor: int
         Penalized average runtime factor. Only used when `run_obj='runtime'`
     """
+    
+    def __init__(self, ta, stats=None, runhistory=None, run_obj="quality", par_factor=1):
+        """
```
mfeurer: We don't need a docstring here, this will be covered by the class docstring.

mlindauer:
I merged the dev into this branch, to avoid the conflict resolving in the other direction. All examples and unit tests are working now. Imho, we could merge this PR now.

examples/rosenbrock.py
```diff
@@ -41,51 +34,15 @@ def rosenbrock_4d(cfg, seed):
...
-finally:
-    smbo.stats.print_stats()
-print("Final Incumbent: %s" %(smbo.incumbent))
 
-inc_value = taf.run( smbo.incumbent )[1]
-print("Optimized Value: %.2f" %(inc_value))
+smac = FuncSMAC(func=rosenbrock_4d,
+                 x0 = [-1,0,1,0],
+                 upper_bounds= [1,1,1,1],
+                 lower_bounds= [-1,-1,-1,-1],
```
KEggensperger: Description of rosenbrock says: "All input parameters are continuous, but we will pretent that x2, x3, and x4 can only take integral values." Is that still true? If yes, I don't see where that is defined and the example needs more documentation.

smac/tae/execute_func.py
mlindauer: fixed

examples/rosenbrock.py
mlindauer: Thanks. fixed

smac/intensification/intensification.py
```diff
@@ -75,14 +74,14 @@ def __init__(self, executor, stats, traj_logger, instances=None,
...
-        self.tae = executor
+        self.tae_runner = tae_runner
 
         self.Adaptive_Capping_Slackfactor = 1.2
 
         if self.run_limit < 1:
             raise ValueError("run_limit must be > 1")
 
-    def intensify(self, challengers, incumbent, run_history, objective,
+    def intensify(self, challengers, incumbent, run_history, aggreagte_func,
```
KEggensperger: typo

mlindauer: thanks. fixed

test/test_smbo/test_smbo.py
```diff
@@ -76,34 +77,34 @@ def test_init_EIPS_as_arguments(self):
                 ['cost', 'runtime'], types)
             eips = EIPS(umrfwi)
             rh2EPM = RunHistory2EPM4EIPS(self.scenario, 2)
-            smbo = SMBO(self.scenario, model=umrfwi, acquisition_function=eips,
-                        runhistory2epm=rh2EPM)
+            smbo = smbo = SMAC(self.scenario, model=umrfwi, acquisition_function=eips,
```
KEggensperger: I suppose that is also a typo?

KEggensperger:
If unittests pass, this PR looks good to me.

test/test_smbo/test_smbo.py
mlindauer: fixed

mfeurer:
Looks good to me, let's get that merged.

PR #73, opened 2016-10-05
ADD unittest for issue #48

mfeurer:
This unittest checks that when passing a target algorithm object to the SMAC class, that the SMAC class actually injects a stats object.


mfeurer:
UPDATE: this now also checks for the injection of the runhistory.

PR #75, opened 2016-10-06
Fix#41

mlindauer:
* extend api of random epm 
* add new facade
* add unit test to ROAR facade
* add hidden cmd line argument

### DIFF
--- a/examples/leadingones.py
+++ b/examples/leadingones.py
@@ -71,7 +71,6 @@ smac = SMAC(scenario=scenario, rng=np.random.RandomState(42),
 try:
     incumbent = smac.optimize()
 finally:
-    smac.stats.print_stats()
     incumbent = smac.solver.incumbent
 
 inc_value = taf.run(incumbent)[1]
--- a/examples/rf.py
+++ b/examples/rf.py
@@ -1,5 +1,7 @@
 # TODO: remove really ugly boilerplate
 import logging
+import os
+import inspect
 
 import numpy as np
 from sklearn.cross_validation import KFold
@@ -122,7 +124,6 @@ smac = SMAC(scenario=scenario, rng=np.random.RandomState(42),
 try:
     incumbent = smac.optimize()
 finally:
-    smac.stats.print_stats()
     incumbent = smac.solver.incumbent
 
 inc_value = taf.run(incumbent)[1]
--- a/smac/epm/random_epm.py
+++ b/smac/epm/random_epm.py
@@ -15,16 +15,16 @@ __version__ = "0.0.1"
 class RandomEpm(object):
     '''implement an epm, which returns only random values'''
 
-    def __init__(self, seed):
+    def __init__(self, rng):
         '''
         initialize random number generator and logger
 
         Parameters
         ----------
-        seed : int
+        rng : np.random.RandomState
         '''
-        self.logger = logging.getLogger("random_epm")
-        self.rng = np.random.RandomState(seed)
+        self.logger = logging.getLogger("RandomEpm")
+        self.rng = rng
 
     def train(self, X, Y, **kwargs):
         '''
@@ -44,7 +44,7 @@ class RandomEpm(object):
         if not isinstance(Y, np.ndarray):
             raise NotImplementedError("Y has to be of type np.ndarray")
 
-        self.logger.debug("Fit model to data")
+        self.logger.debug("(Pseudo) Fit model to data")
 
     def predict(self, X):
         '''
@@ -64,4 +64,42 @@ class RandomEpm(object):
         '''
         if not isinstance(X, np.ndarray):
             raise NotImplementedError("X has to be of type np.ndarray")
+        return self.rng.rand(len(X), 1), self.rng.rand(len(X), 1)
+    
+    def _predict(self, x):
+        """Predict mean and variance for given x.
+
+        Parameters
+        ----------
+        x : np.ndarray of shape = [n_features (config + instance features), ]
+
+        Returns
+        -------
+        mean : float
+            Predictive mean
+        var : float
+            Predictive variance
+        """
+        mean, var = self.rf.predict(x)
+
+        return self.rng.rand(), self.rng.rand()
+
+    def predict_marginalized_over_instances(self, X):
+        """Predict mean and variance marginalized over all instances.
+
+        Returns the predictive mean and variance marginalised over all
+        instances for a set of configurations.
+
+        Parameters
+        ----------
+        X : np.ndarray of shape = [n_features (config), ]
+
+        Returns
+        -------
+        means : np.ndarray of shape = [n_samples, 1]
+            Predictive mean
+        vars : np.ndarray  of shape = [n_samples, 1]
+            Predictive variance
+        """
+
         return self.rng.rand(len(X), 1), self.rng.rand(len(X), 1)
--- a/smac/epm/random_epm.py
+++ b/smac/epm/random_epm.py
@@ -11,7 +11,7 @@ __email__ = "eggenspk@cs.uni-freiburg.de"
 __version__ = "0.0.1"
 
 
-class RandomEpm(AbstractEPM):
+class RandomEPM(AbstractEPM):
     '''implement an epm, which returns only random values'''
 
     def __init__(self, rng):
@@ -22,12 +22,12 @@ class RandomEpm(AbstractEPM):
         ----------
         rng : np.random.RandomState
         '''
-        self.logger = logging.getLogger("RandomEpm")
+        self.logger = logging.getLogger("RandomEPM")
         self.rng = rng
 
     def train(self, X, Y, **kwargs):
         '''
-        Trains the random forest on X and Y.
+        Pseudo training on X and Y.
 
         Parameters
         ----------
--- a/smac/epm/random_epm.py
+++ b/smac/epm/random_epm.py
@@ -66,24 +66,6 @@ class RandomEpm(object):
             raise NotImplementedError("X has to be of type np.ndarray")
         return self.rng.rand(len(X), 1), self.rng.rand(len(X), 1)
     
-    def _predict(self, x):
-        """Predict mean and variance for given x.
-
-        Parameters
-        ----------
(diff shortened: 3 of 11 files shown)


examples/leadingones.py
```diff
@@ -71,7 +71,6 @@ def leading_ones(cfg, seed):
 try:
     incumbent = smac.optimize()
 finally:
-    smac.stats.print_stats()
```
mfeurer: Why are these statements removed in this PR?

examples/rf.py
```diff
@@ -122,7 +124,6 @@ def rfr(cfg, seed):
 try:
     incumbent = smac.optimize()
 finally:
-    smac.stats.print_stats()
```
mfeurer: Same here.

smac/epm/random_epm.py
```diff
@@ -64,4 +64,42 @@ def predict(self, X):
         '''
         if not isinstance(X, np.ndarray):
             raise NotImplementedError("X has to be of type np.ndarray")
+        return self.rng.rand(len(X), 1), self.rng.rand(len(X), 1)
+    
+    def _predict(self, x):
```
mfeurer: I think the inclusion of this method is a copy-paste error?

smac/facade/roar_facade.py
```diff
@@ -0,0 +1,130 @@
...
+                 runhistory: RunHistory=None,
+                 intensifier: Intensifier=None,
+                 acquisition_function: AbstractAcquisitionFunction=None,
+                 model=None,
+                 runhistory2epm: AbstractRunHistory2EPM=None,
+                 initial_design: InitialDesign=None,
+                 stats: Stats=None,
+                 rng: np.random.RandomState=None):
+        '''
+        Facade to use ROAR default mode
```
mfeurer: What is the 'ROAR default mode'? I only know the 'ROAR' mode.

smac/facade/roar_facade.py
```diff
@@ -0,0 +1,130 @@
...
+            run(self, config)
+            If not set, it will be initialized with the tae.ExecuteTARunOld()
+        runhistory: RunHistory
+            runhistory to store all algorithm runs
+        intensifier: Intensifier
+            intensification object to issue a racing to decide the current incumbent
+        acquisition_function : AcquisitionFunction
+            Object that implements the AbstractAcquisitionFunction. Will use
+            EI if not set.
+        model : object
```
mfeurer: I think the model shouldn't be an argument to the ROAR class

smac/facade/roar_facade.py
```diff
@@ -0,0 +1,130 @@
...
+
+class ROAR(SMAC):
+
+    def __init__(self,
+                 scenario: Scenario,
+                 tae_runner: ExecuteTARun=None,
+                 runhistory: RunHistory=None,
+                 intensifier: Intensifier=None,
+                 acquisition_function: AbstractAcquisitionFunction=None,
+                 model=None,
```
mfeurer: I think the model shouldn't be an argument to the ROAR class.

smac/facade/roar_facade.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class ROAR(SMAC):
+
+    def __init__(self,
+                 scenario: Scenario,
+                 tae_runner: ExecuteTARun=None,
+                 runhistory: RunHistory=None,
+                 intensifier: Intensifier=None,
+                 acquisition_function: AbstractAcquisitionFunction=None,
```
mfeurer: I don't think an acquisition function is necessary because we don't use a model.

smac/facade/roar_facade.py
```diff
@@ -0,0 +1,130 @@
...
+        else:
+            raise TypeError('Unknown type %s for argument rng. Only accepts '
+                            'None, int or np.random.RandomState' % str(type(rng)))
+
+        # initial EPM
+        if model is None:
+            #use random predictions to simulate random sampling of configurations
+            model = RandomEpm(rng=rng)
+
+        # initial conversion of runhistory into EPM data
```
mfeurer: Why is this copied here? Can't this part of code be reused from the SMAC class?

smac/facade/roar_facade.py
```diff
@@ -0,0 +1,130 @@
...
+        initial_design: InitialDesign
+            initial sampling design
+        stats: Stats
+            optional stats object
+        rng: np.random.RandomState
+            Random number generator
+        '''
+        self.logger = logging.getLogger("ROAR")
+
+        # initialize random number generator
```
mfeurer: This code duplication hints that setting the rng should be a private method in the SMAC class which can then be called in both constructors.

smac/facade/roar_facade.py
```diff
@@ -0,0 +1,130 @@
...
+            num_run = rng
+            rng = np.random.RandomState(seed=rng)
+        elif isinstance(rng, np.random.RandomState):
+            num_run = rng.randint(1234567980)
+            rng = rng
+        else:
+            raise TypeError('Unknown type %s for argument rng. Only accepts '
+                            'None, int or np.random.RandomState' % str(type(rng)))
+
+        # initial EPM
```
mfeurer: In case ROAR accepts a model, we should in all cases replace it here an issue a warning.

smac/epm/random_epm.py
mlindauer: removed

examples/leadingones.py
mlindauer: Because I wanted to merge a working PR without obvious bugs.

examples/rf.py
mlindauer: same as above

smac/facade/roar_facade.py
mlindauer: fixed

smac/facade/roar_facade.py
mlindauer: fixed

smac/facade/roar_facade.py
mlindauer: fixed

smac/facade/roar_facade.py
mlindauer: obsolet

smac/facade/roar_facade.py
mlindauer: because we can always disable the imputation of censored data.
We could remove the if-statement and always use RunHistory2EPM4Cost since the data is not used anyway.

examples/leadingones.py
mfeurer: Sorry, I don't see the bug here

smac/facade/roar_facade.py
mfeurer: If the data is not used anyway, wouldn't it be better to use a random imputer to reduce the running time of the ROAR algorithm?

mlindauer: right now, we use no imputer at all, which should be most time-efficient variant, right? We could replace the runhistory2EPM by a random class. However, this function is typically not the bottle neck. 
To save time, we would need to remove the local search and the sorting of the random configuration based on the acq function. However, this would require direct changes in the smbo class which I would like to avoid.

smac/facade/roar_facade.py
mlindauer: but we need the rng for the random model!?

smac/facade/roar_facade.py
KEggensperger: Is th docstring also already updated?

smac/facade/roar_facade.py
KEggensperger: Agree, but maybe this should be a different PR

smac/facade/roar_facade.py
mlindauer: fixed now.

KEggensperger:
I guess we now want the base_epm code, as we use two different epms?

smac/facade/roar_facade.py
mlindauer: Something has to be called by SMBO. We could add a nearly empty RunHistory2EPM class. However, more code could lead to more to maintain. I wanted to minimize the changes as much as possible

mlindauer:
> I guess we now want the base_epm code, as we use two different epms?

to define the api? or why?

KEggensperger:
Yes, but also to make the integration of new models simpler. Before we used only one epm so this was not necessary.

mlindauer:
Should this be in the same PR or should we open a new one?

mfeurer:
New PR please :)

mlindauer:
Ok, are there any open issues for this PR or can we merge?

mfeurer:
Here are the open todos:

- [ ] base class of EPM
- [ ] anwer my first question
- [ ] discuss the use of runhistory 2 EPM. I still find it weird that the user can pass this to ROAR. Couldn't we just remove the argument and pass `None` to the SMAC constructor?

examples/leadingones.py
mlindauer: The optimize function of SMAC looks like

```
        try:
            incumbent = self.solver.run()
        finally:
            self.solver.stats.print_stats()
            self.logger.info("Final Incumbent: %s" % (self.solver.incumbent))
```

So, this would duplicate the output of the stats.

mlindauer:
> base class of EPM

see commit

> anwer my first question

see answer

> discuss the use of runhistory 2 EPM. I still find it weird that the user can pass this to ROAR. Couldn't we just remove the argument and pass None to the SMAC constructor?

I agree that the argument in the ROAR facade is useless. Therefore, I removed it.
To avoid that SMAC will create a runhistory2epm with censored data imputation, I have to pass something to SMAC.

mfeurer:
Awesome, we're getting close to be able to merge this. Could you please update the docstrings in the abstract EPM and change the author of the abstract EPM package? Then I also think this PR is ready to be merged.

mlindauer:
> Could you please correct the docstrings in the abstract EPM 

done. I hope I haven't overlooked anything.

> and change the author of the abstract EPM package? 

funny to be the author for a bit of copy and pasta.

PR #76, opened 2016-10-06
FIX #72, config space now mandatory for scenario

### LINKED ISSUE #72: Make configuration space a mandatory object
Currently, it is possible to create a scenario object without passing a configuration space object to the scenario. SMAC then fails later on. Will create a PR soonish.

mfeurer:
This PR makes sure that either the argument `cs` or `paramfile` is passed to the scenario object.

### DIFF
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 import os
 import sys
 import logging
@@ -52,6 +53,7 @@ class Scenario(object):
             scenario.update(cmd_args)
 
         self._arguments = {}
+        self._groups = defaultdict(set)
         self._add_arguments()
 
         # Parse arguments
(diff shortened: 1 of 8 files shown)


smac/scenario/scenario.py
```diff
@@ -64,13 +66,48 @@ def __init__(self, scenario, cmd_args=None):
             raise ValueError('Could not parse the following arguments: %s' %
                              str(list(scenario.keys())))
 
+        for group, potential_members in self._groups.items():
```
KEggensperger: If an argument is also set as required this check is not influenced and the code might crash before?
This should be clarified in the docstring: if an argument is part of the required group one should not mark it as required.

smac/scenario/scenario.py
```diff
@@ -64,13 +66,48 @@ def __init__(self, scenario, cmd_args=None):
...
+            If given, the callback will be called when the argument is
+            parsed. Useful for custom casting/typechecking.
+        default : object, optional
+            Default value if the argument is not given. Default to ``None``.
+        dest : str
+            Assign the argument to scenario object by this name.
+        required : bool
+            If True, the scenario will raise an error if the argument is not
+            given.
+        required_group : str
```
KEggensperger: The name is misleading. What about mutually_exclusive_group?

smac/scenario/scenario.py
mfeurer: I added a check for that and adapted the docstring.

smac/scenario/scenario.py
mfeurer: I changed the name of the argument.

PR #78, opened 2016-10-07
ADD allow passing callable as target algorithm

### LINKED ISSUE #49: Allow passing target algorithm directly to SMAC
instead of wrapping it into a class manually. What do you think?

mfeurer:
Allows passing a callable to the SMAC or ROAR class. The SMAC class will then wrap the callable in the necessary ExecuteTaRun class. Fixes #49

### DIFF
--- a/smac/facade/smac_facade.py
+++ b/smac/facade/smac_facade.py
@@ -125,6 +125,11 @@ class SMAC(object):
                                          run_obj=scenario.run_obj,
                                          runhistory=runhistory,
                                          par_factor=scenario.par_factor)
+        elif not isinstance(tae_runner, ExecuteTARun):
+            raise TypeError("Argument 'tae_runner' is %s, but must be "
+                            "either a callable or an instance of ExecuteTaRun."
+                            % type(tae_runner))
+
         # inject stats if necessary
         if tae_runner.stats is None:
             tae_runner.stats = self.stats
--- a/smac/facade/smac_facade.py
+++ b/smac/facade/smac_facade.py
@@ -4,6 +4,7 @@ import numpy as np
 
 from smac.tae.execute_ta_run import ExecuteTARun
 from smac.tae.execute_ta_run_old import ExecuteTARunOld
+from smac.tae.execute_func import ExecuteTAFunc
 from smac.tae.execute_ta_run import StatusType
 from smac.stats.stats import Stats
 from smac.scenario.scenario import Scenario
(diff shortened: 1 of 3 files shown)


smac/facade/smac_facade.py
```diff
@@ -32,7 +33,9 @@ class SMAC(object):
 
     def __init__(self,
                  scenario: Scenario,
-                 tae_runner: ExecuteTARun=None,
+                 # TODO: once we drop python3.4 add type hint
```
mlindauer: I don't like such TODOs which will maybe never resolved. Either cut it or do it right away.

smac/facade/smac_facade.py
```diff
@@ -111,7 +113,13 @@ def __init__(self,
...
 
         # initialize tae_runner
-        if tae_runner is None:
+        if tae_runner is not None and callable(tae_runner):
+            tae_runner = ExecuteTAFunc(ta=tae_runner,
+                                       stats=self.stats,
+                                       run_obj=scenario.run_obj,
+                                       runhistory=runhistory,
+                                       par_factor=scenario.par_factor)
+        elif tae_runner is None:
```
mlindauer: Do we want to raise an exception in case both if statements fail?

smac/facade/smac_facade.py
mfeurer: Typing is not available in python3.4, so I can't do it right away. You're sure I should not leave it in there?

smac/facade/smac_facade.py
mfeurer: If both statements fail it must be an instance of ExecuteTaRun, right?

smac/facade/smac_facade.py
mlindauer: same discussion as before: I vote for Python 3.5 support ;-)

mfeurer: We support python3.5 (and python3.4).

smac/facade/smac_facade.py
mfeurer: What would the exception then be about?

mfeurer: I just added that check.

mlindauer: the exception should also mention that a passed None will lead to a TAE based on the call in the scenario.

smac/facade/smac_facade.py
mlindauer: Ok, let's be more concrete: I vote for Python 3.5 support only. The only larger project that depends on SMAC3 is auto-sklearn right now, right?  And you are going to drop Python 2.7 soon -- so you could go for Python 3.5 straight away.

Furthermore, I would expect that either we will drop Python 3.4 support soon or we keep it for a long time. 
And I don't see a reason for a Python 3.4 support right now. We decided to go for Python 3 in SMAC3 because it is the default in Ubuntu 16.04, but in fact it is Python 3.5 and not 3.4 in there.

Last but not least, could you please also fix the readme and remove the sentence: 

> We expect that it should be also compatible with Python 2.7.

smac/facade/smac_facade.py
mfeurer: Okay, I just updated the exception.

PR #84, opened 2016-10-13
ADD first draft of scipy.optimize.fmin_l_bfgs_b-like interface

### LINKED ISSUE #82: Func facade is not a drop-in replacement for current optimization packages
The newly added [facade for direct optimization](https://github.com/automl/SMAC3/blob/development/smac/facade/func_facade.py) is only partially compatible with common minimization algorithms such as the ones in scipy.minimize, direct, cma-es etc... The two most striking differences I found is the invocation, which should be a function instead of a class, and the second is the signature of the target algorithm. I have worked on a small example on how a function `fmin_smac` could look like. It is a refactoring of the current FuncSMAC class:

```
def fmin_smac(func: callable,
              x0: list,
              bounds: list,
              maxfun: int=-1,
              maxtime: int=-1,
              rng: np.random.RandomState=None):
    """Minimize a function func using the SMAC algorithm.

    This method is a convenience wrapper for the SMAC class.

    Parameters
    ----------
    func : callable f(x)
        Function to minimize.
    x0 : list
        Initial guess/default configuration.
    bounds : list
        ``(min, max)`` pairs for each element in ``x``, defining the bound on
        that parameters.
    maxiter : int, optional
        Maximum runtime in seconds.
    maxfun : int, optional
        Maximum number of function evaluations.
    rng : np.random.RandomState, optional
            Random number generator used by SMAC.

    Returns
    -------
    x : list
        Estimated position of the minimum.
    f : float
        Value of `func` at the minimum.
    r : :class:`smac.runhistory.runhistory.RunHistory`
        Information on the SMAC run.
    """

    aggregate_func = average_cost
    # initialize empty runhistory; needed to pass to TA
    runhistory = RunHistory(aggregate_func=aggregate_func)

    # create configuration space
    cs = ConfigurationSpace()
    for idx, (lower_bound, upper_bound) in enumerate(bounds):
        parameter = UniformFloatHyperparameter(name="x%d" % (idx + 1),
                                               lower=lower_bound,
                                               upper=upper_bound,
                                               default=x0[idx])
        cs.add_hyperparameter(parameter)

    # create scenario
    scenario_dict = {"run_obj": "quality",  # we optimize quality
                     "cs": cs,  # configuration space
                     "deterministic": "true",
                     }
    if maxfun > -1:
        scenario_dict["runcount_limit"] = maxfun
    if maxtime > - 1:
        scenario_dict["wallclock_limit"] = maxtime
    scenario = Scenario(scenario_dict)

    smac = SMAC(scenario=scenario, tae_runner=func, rng=rng)
    smac.logger = logging.getLogger("fmin_smac")
    incumbent = smac.optimize()

    config_id = smac.solver.runhistory.config_ids[incumbent]
    run_key = RunKey(config_id, None, 0)
    incumbent_performance = smac.solver.runhistory.data[run_key]
    incumbent = [incumbent['x%d' % (idx + 1)] for idx in range(len(bounds))]
    return incumbent, incumbent_performance.cost, \
           smac.solver.runhistory
```

When I tried to replace `scipy.optimize.fmin_l_bfgs_b` with fmin_smac I figured that the target algorithms actually have a quite different signature. While SMAC expects `func(dict, seed)`, scipy.optimize expects `func(array-like)`. To make using the code as easy as possible, I suggest that we also add a target algorithm executer for this case, which does nothing except calling the target algorithm with an array-like and without a seed .

mfeurer:
Fixes #82

### DIFF
--- a/smac/facade/func_facade.py
+++ b/smac/facade/func_facade.py
@@ -7,7 +7,7 @@ from smac.scenario.scenario import Scenario
 from smac.configspace import ConfigurationSpace
 from smac.smbo.objective import average_cost
 from smac.runhistory.runhistory import RunKey
-from smac.tae.execute_func import ExecuteTAFunc4FMIN
+from smac.tae.execute_func import ExecuteTAFuncArray
 
 from ConfigSpace.hyperparameters import UniformFloatHyperparameter
 
@@ -64,7 +64,7 @@ def fmin_smac(func: callable,
         cs.add_hyperparameter(parameter)
 
     # Create target algorithm runner
-    ta = ExecuteTAFunc4FMIN(ta=func)
+    ta = ExecuteTAFuncArray(ta=func)
 
     # create scenario
     scenario_dict = {"run_obj": "quality",  # we optimize quality
--- a/smac/facade/func_facade.py
+++ b/smac/facade/func_facade.py
@@ -6,7 +6,7 @@ from smac.facade.smac_facade import SMAC
 from smac.scenario.scenario import Scenario
 from smac.configspace import ConfigurationSpace
 from smac.smbo.objective import average_cost
-from smac.runhistory.runhistory import RunHistory, RunKey
+from smac.runhistory.runhistory import RunKey
 from smac.tae.execute_func import ExecuteTAFunc4FMIN
 
 from ConfigSpace.hyperparameters import UniformFloatHyperparameter
@@ -35,7 +35,7 @@ def fmin_smac(func: callable,
     bounds : list
         ``(min, max)`` pairs for each element in ``x``, defining the bound on
         that parameters.
-    maxiter : int, optional
+    maxtime : int, optional
         Maximum runtime in seconds.
     maxfun : int, optional
         Maximum number of function evaluations.
@@ -53,8 +53,6 @@ def fmin_smac(func: callable,
     """
 
     aggregate_func = average_cost
-    # initialize empty runhistory; needed to pass to TA
-    runhistory = RunHistory(aggregate_func=aggregate_func)
 
     # create configuration space
     cs = ConfigurationSpace()
@@ -73,9 +71,9 @@ def fmin_smac(func: callable,
                      "cs": cs,  # configuration space
                      "deterministic": "true",
                      }
-    if maxfun > -1:
+    if maxfun > 0:
         scenario_dict["runcount_limit"] = maxfun
-    if maxtime > - 1:
+    if maxtime > 0:
         scenario_dict["wallclock_limit"] = maxtime
     scenario = Scenario(scenario_dict)
 
--- a/smac/facade/func_facade.py
+++ b/smac/facade/func_facade.py
@@ -4,84 +4,89 @@ import numpy as np
 
 from smac.facade.smac_facade import SMAC
 from smac.scenario.scenario import Scenario
-from smac.stats.stats import Stats
 from smac.configspace import ConfigurationSpace
-from smac.tae.execute_func import ExecuteTAFunc
 from smac.smbo.objective import average_cost
-from smac.runhistory.runhistory import RunHistory
+from smac.runhistory.runhistory import RunHistory, RunKey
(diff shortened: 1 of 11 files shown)


smac/facade/func_facade.py
```diff
@@ -4,84 +4,89 @@
...
-                         tae_runner=taf,
-                         stats=self.stats,
-                         runhistory=runhistory,
-                         rng=rng)
+    # create scenario
+    scenario_dict = {"run_obj": "quality",  # we optimize quality
+                     "cs": cs,  # configuration space
+                     "deterministic": "true",
+                     }
+    if maxfun > -1:
```
mlindauer: this (and the next if) should be > 0, right?

smac/facade/func_facade.py
```diff
@@ -4,84 +4,89 @@
...
-        cs = ConfigurationSpace()
-        for indx in range(0, len(x0)):
-            parameter = UniformFloatHyperparameter(name="x%d" % (indx+1),
-                                                   lower=lower_bounds[indx],
-                                                   upper=upper_bounds[indx],
-                                                   default=x0[indx])
-            cs.add_hyperparameter(parameter)
+    aggregate_func = average_cost
+    # initialize empty runhistory; needed to pass to TA
+    runhistory = RunHistory(aggregate_func=aggregate_func)
```
mlindauer: You never you this runhistory object. Either pass it to SMAC() or cut it.

smac/tae/execute_func.py
```diff
@@ -109,3 +95,79 @@ def run(self, config, instance=None,
...
+            result = rval[0]
+            additional_run_info = rval[1]
+        else:
+            result = rval
+            additional_run_info = {}
+
+        return result, additional_run_info
+
+
+class ExecuteTAFunc4FMIN(AbstractTAFunc):
```
mlindauer: the name "4FMIN" is not very descriptive. Please change the name to something that describes what is the difference and not for what we use it.

smac/facade/func_facade.py
mfeurer: My bad, will change this.

smac/tae/execute_func.py
mfeurer: If I knew a better name I would immediately change this. What about ExecuteTaFuncSimple?

mfeurer: Then we should rename the other executor to ExecuteTaFuncDict and make sure they both accept the same interface. While simple is also not really descriptive, it hints that there might be more than just a change in the input argument.

mlindauer: I would prefer the first option (i.e., also changing the other executor).

mfeurer: Okay, I'll add a draft to the general comments on this PR.

mfeurer:
If we want the two target algorithm runners look the same, but only differ in the way they pass data to the target algorithm we need to consider the following change: We need to allow target algorithms to accept a single argument, also in the current ExecuteTAFunc. We could check the number of arguments the target algorithm accepts in the constructor. If it's only a single argument, it's the configuration. If there are two arguments, we should define that argument number two is always the seed, i.e., it shouldn't be possible to pass an instance to the target algorithm but not the seed. Thus, I propose that ExecuteTaFuncArray and ExecuteTAFuncDict accept target algorithms with the following signature:

```
def target_algorithm(config: list) -> typing.Union[float, typing.Tuple[float, typing.Any]]:
```

or 

```
def target_algorithm(config: list, seed: int) -> typing.Union[float, typing.Tuple[float, typing.Any]]:
```

or

```
def target_algorithm(config: list, instance: str, seed: int) -> typing.Union[float, typing.Tuple[float, typing.Any]]:
```

Hopefully, this finishes this PR.

mlindauer:
I agree but shouldn't it be in the last signature `target_algorithm (config, seed, instance)`?

mfeurer:
I don't mind the order. I'll update the PR with the order you proposed.

Issue #98, opened 2016-10-24
RNG for ConfigSpace 


KEggensperger:
When using a PCS-file we seed the ConfigCpace with '42', no matter what SMAC's rng is:
https://github.com/automl/SMAC3/blob/master/smac/scenario/scenario.py#L262

This results in drawing the same random configuration for different SMAC runs with different seeds?
https://github.com/automl/SMAC3/blob/master/smac/smbo/smbo.py#L225

I propose to draw a random integer with SMAC's rng and reset the rng in the ConfigSpace object with this integer in each "facade"

Issue #22, opened 2016-08-17
Start from a random configuration


mlindauer:
Hi,

for comparing against other systems, it is sometime necessary to start SMAC from a random configuration instead of from a default configuration.

Best,
Marius

PR #99, opened 2016-10-24
Fix#22

KEggensperger:
Solves issue #22 by adding initial_incumbent as an optional input to the scenario file.
By default this is set to random and ignored in case an initial design is set. 

Needs #98 to be solved.

### DIFF
--- a/smac/initial_design/random_configuration_design.py
+++ b/smac/initial_design/random_configuration_design.py
@@ -1,0 +1,96 @@
+import sys
+
+import numpy as np
+
+from smac.initial_design.initial_design import InitialDesign
+from smac.tae.execute_ta_run import ExecuteTARun
+from smac.stats.stats import Stats
+from smac.utils.io.traj_logging import TrajLogger
+from smac.scenario.scenario import Scenario
+from smac.tae.execute_ta_run import StatusType
+from smac.runhistory.runhistory import RunHistory
+from smac.utils import constants
+
+__author__ = "Katharina Eggensperger"
+__copyright__ = "Copyright 2016, ML4AAD"
+__license__ = "3-clause BSD"
+
+
+class RandomConfiguration(InitialDesign):
+
+    def __init__(self,
+                 tae_runner: ExecuteTARun,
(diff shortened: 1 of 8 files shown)


smac/initial_design/random_configuration_design.py
```diff
@@ -0,0 +1,96 @@
...
+            runs the initial design by calling the target algorithm
+            and adding new entries to the trajectory logger
+
+            Returns
+            -------
+            incumbent: Configuration()
+                initial incumbent configuration
+        '''
+
+        random_conf = self.scenario.cs.sample_configuration(1)
```
mfeurer: Don't pass a size here, or make it a keyword argument.

smac/scenario/scenario.py
```diff
@@ -190,6 +190,9 @@ def _add_arguments(self):
                           dest='train_insts')
         self.add_argument(name='test_instances', default=[[None]], help=None,
                           dest='test_insts')
+        self.add_argument(name='initial_incumbent', default="DEFAULT",
+                          help=None, dest='initial_incumbent',
+                          callback=lambda arg: arg if arg in ['DEFAULT', 'RANDOM'] else None)
```
mfeurer: This could pose a problem if a user provides the argument `DEFAULt`, it will be replaced by `None`, and the user will get an error message that the facade doesn't know what initial design None is. Also, I think we should make this case-insensitive.

smac/initial_design/random_configuration_design.py
```diff
@@ -0,0 +1,96 @@
...
+from smac.tae.execute_ta_run import StatusType
+from smac.runhistory.runhistory import RunHistory
+from smac.utils import constants
+
+__author__ = "Katharina Eggensperger"
+__copyright__ = "Copyright 2016, ML4AAD"
+__license__ = "3-clause BSD"
+
+
+class RandomConfiguration(InitialDesign):
```
mfeurer: This class is mostly a copy of the `DefaultConfiguration`, I think this should be generalized.

Issue #85, opened 2016-10-13
SMAC crashes when trying to abort when the first run crashes


mfeurer:
In `default_configuration_design`, if the status of the first run is CRASHED or ABORT, SMAC tries to exit, but crashs itself in the following code:

```
if status in [StatusType.CRASHED or StatusType.ABORT]:
    self.logger.critical("First run crashed -- Abort")
    sys.exit(1)
```

Reasons are a missing import of `sys` and a missing logger.

mlindauer:
already fixed

PR #101, opened 2016-10-25
FIX#98 resetting ConfigSpace random number generator

KEggensperger:
This PR draws a random integer with MAXINT as its maximum to reset the rng in the ConfigSpace object in the SMAC facade, which is used by all other facades.

It also replaces '1234567980' with MAXINT as the maximum random integer to seed the random forest.


smac/facade/smac_facade.py
```diff
@@ -93,6 +94,9 @@ def __init__(self,
         # initial random number generator
         num_run, rng = self._get_rng(rng=rng)
 
+        # reset random number generator in config space
+        scenario.cs.random = np.random.RandomState(rng.randint(MAXINT))
```
mfeurer: I think you should either do `scenario.cs.seed(rng.randint(MAXINT))` or `scenario.cs.random = rng`, but not create a new random state here.

KEggensperger: fixed

smac/initial_design/single_config_initial_design.py
```diff
@@ -0,0 +1,108 @@
...
+
+        status, cost, runtime, additional_info = self.tae_runner.start(
+            initial_incumbent,
+            instance=rand_inst,
+            cutoff=self.scenario.cutoff,
+            seed=initial_seed,
+            instance_specific=self.scenario.instance_specific.get(rand_inst, "0"))
+
+        if status in [StatusType.CRASHED or StatusType.ABORT]:
+            self.logger.critical("First run crashed -- Abort")
```
mfeurer: You could fix #85 here if you want ;)

test/test_facade/test_roar_facade.py
```diff
@@ -22,3 +24,41 @@ def test_inject_stats_and_runhistory_object_to_TAE(self):
         ROAR(tae_runner=ta, scenario=self.scenario)
         self.assertIsInstance(ta.stats, Stats)
         self.assertIsInstance(ta.runhistory, RunHistory)
+
+    def test_check_random_states(self):
```
mfeurer: I think these tests suffice in the smac facade and don't have to be duplicated.

smac/scenario/scenario.py
```diff
@@ -190,6 +189,10 @@ def _add_arguments(self):
                           dest='train_insts')
         self.add_argument(name='test_instances', default=[[None]], help=None,
                           dest='test_insts')
+        self.add_argument(name='initial_incumbent', default="DEFAULT",
+                          help=None, dest='initial_incumbent',
+                          callback=lambda arg: str.upper(arg.strip()) if
+                          str.upper(arg.strip()) in ['DEFAULT', 'RANDOM'] else None)
```
mfeurer: Hm, although we should have a choice here, a hotfix would be raising an exception in the else statement. Would that make sense?

smac/initial_design/single_config_initial_design.py
KEggensperger: Should be already fixed

test/test_facade/test_roar_facade.py
KEggensperger: agree

smac/scenario/scenario.py
KEggensperger: agree, but that is not a valid statement

PR #107, opened 2016-11-04
FIX #104 Make the Scenario objects pickable

### LINKED ISSUE #104: Scenario can't be pickled
Currently, the scenario can't be pickled due to lambda functions in the callback argument of the arguments in the scenario: https://github.com/automl/SMAC3/blob/master/smac/scenario/scenario.py#L159

Removing these should make the SMAC object pickleable again.

Ayaro:


smac/scenario/scenario.py
```diff
@@ -18,6 +18,14 @@
 __version__ = "0.0.2"
 
 
+def _is_truthy(arg):
+    return arg in ["1", "true", True]
+
+def _parse_initial_incumbent(arg):
+    arg = str.upper(arg.strip())
+    return arg if arg in ['DEFAULT', 'RANDOM'] else None
```
mlindauer: If we want to add a "choice" argument (as in argparse) in another parameter, we have to write every time a new function?

mfeurer: No, I don't think so. This is currently a fix to allow the scenario to be picklable again. Once we add a choice argument, we can remove this function again.

mlindauer: and why don't we add a choice argument now?

mfeurer: Good point. Should we try to use the argparser instead of custom code?

mfeurer: I think it should be possible. @Ayaro could you please try this and submit an update to this PR?

mfeurer:
@mlindauer @KEggensperger ready for review.

smac/scenario/scenario.py
```diff
@@ -121,18 +126,21 @@ def add_argument(self, name, help, callback=None, default=None,
...
-                                 'callback': callback}
+                                 'callback': callback,
+                                 'choice': choice}
 
         if mutually_exclusive_group:
             self._groups[mutually_exclusive_group].add(name)
 
     def _parse_argument(self, name, scenario, help, callback=None, default=None,
-                        dest=None, required=False):
+                        dest=None, required=False, choice=None):
```
mlindauer: could we please add a docstring for `_parse_argument`

mfeurer: Docstring is there. Do you approve now?

PR #108, opened 2016-11-15
Warmstart

mlindauer:
ADD methods to warmstart SMAC and speed up imputation of censored data

* ADD utils.merge_foreign_data to load runhistory data from a different scenarios with the same pcs, feature space and running time cutoff
* ADD PCA on instance features
* MAINT vectorize truncnorm computation
* MAINT reduce number of imputation iterations to 2 (as in SMAC2)
* MAINT increase epsilon of SLS (as in SMAC2)

### DIFF
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -67,6 +67,7 @@ class Scenario(object):
             arg_name, arg_value = self._parse_argument(key, scenario, **value)
             parsed_arguments[arg_name] = arg_value
 
+        
         if len(scenario) != 0:
             raise ValueError('Could not parse the following arguments: %s' %
                              str(list(scenario.keys())))
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -6,6 +6,7 @@ import numpy
 import shlex
 import time
 import datetime
+import copy
 
 from sklearn.decomposition import PCA
 from sklearn.preprocessing import MinMaxScaler
@@ -49,7 +50,7 @@ class Scenario(object):
             self.logger.info("Reading scenario file: %s" % (scenario_fn))
             scenario = self.in_reader.read_scenario_file(scenario_fn)
         elif type(scenario) is dict:
-            pass
+            scenario = copy.copy(scenario)
         else:
             raise TypeError(
                 "Wrong type of scenario (str or dict are supported)")
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -259,6 +259,7 @@ class Scenario(object):
             for inst_ in self.train_insts:
                 self.feature_array.append(self.feature_dict[inst_])
             self.feature_array = numpy.array(self.feature_array)
+            self.n_features = self.feature_array.shape[1]
             
             # reduce dimensionality of features of larger than PCA_DIM
             if self.feature_array.shape[1] > self.PCA_DIM:
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -8,6 +8,7 @@ import time
 import datetime
 
 from sklearn.decomposition import PCA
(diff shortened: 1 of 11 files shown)


smac/scenario/scenario.py
```diff
@@ -250,12 +255,24 @@ def extract_instance_specific(instance_list):
...
-                self.feature_dict[list(self.feature_dict.keys())[0]])
             self.feature_array = []
             for inst_ in self.train_insts:
                 self.feature_array.append(self.feature_dict[inst_])
             self.feature_array = numpy.array(self.feature_array)
+            
+            # reduce dimensionality of features of larger than PCA_DIM
+            if self.feature_array.shape[1] > self.PCA_DIM:
+                X = self.feature_array
+                # scale features
```
mfeurer: From my understanding of [the code](https://github.com/scikit-learn/scikit-learn/blob/a5ab948/sklearn/decomposition/pca.py#L293), PCA does the scaling itself. Otherwise, I think you should use the MinMaxScaler form scikit-learn.

smac/scenario/scenario.py
```diff
@@ -250,12 +255,24 @@ def extract_instance_specific(instance_list):
...
+                X = (X - X.min()) / (X.max() - X.min())
+                X = numpy.nan_to_num(X) # if features with max == min
+                #PCA
+                pca = PCA(n_components=self.PCA_DIM)
+                self.feature_array = pca.fit_transform(X)
+                self.n_features = self.PCA_DIM
```
mfeurer: This will only be set if the number of features is higher than the number of PCA dimensions. What happens in the other case?

smac/smbo/local_search.py
```diff
@@ -16,7 +16,7 @@
 class LocalSearch(object):
 
     def __init__(self, acquisition_function, config_space,
-                 epsilon=0.000001, max_iterations=None, rng=None):
+                 epsilon=0.00001, max_iterations=None, rng=None):
```
mfeurer: This is an increase in epsilon, while the PR comment states that this PR lowers epsilon.

mfeurer:
Okay, could you please also take care of the decreased coverage?

KEggensperger:
I fixed the unittest by 
a) requiring scikit-learn (instaed of sklearn)
b) requiring the package in setup.py (I have no idea what/who uses/needs requirements.txt)

smac/scenario/scenario.py
mlindauer: I haven't found the scaling in the PCA code; I found only centering around the mean which is not sufficient. Could you please point me to the exact code line for the scaling?
The MinMaxScaler of sklearn is also only doing the exact same stuff as I do in one line -- we save lines for imports, init of object and fit_transform call.

smac/scenario/scenario.py
mlindauer: Good catch. Thanks. Fixed now

mlindauer:
@KEggensperger : Thank you very much for the fix. 
Why do we list the requirements in `requirements.txt` and `setup.py`? I would say that such redundancies should be avoided.

smac/scenario/scenario.py
mfeurer: This is a valid argument, but on the other hand we would rely on well-tested code. Inspecting the code again, did you actually check that it works as intended? Calling `X.max()` returns the maximum of the whole array, so this code scales between the maximum of all features and the minimum of all features. This would only cost us a single line, namely the import, or no line if we do `X = MinMaxScaler().fit_transform(X)`.

mlindauer: fair enough. fixed

mlindauer:
@mfeurer I added a unit test for the new `merge_foreign_data` function. the coverage will probably nevertheless decrease since I wrote also methods to do it based on files (and not on internal objects) and I don't see the point in testing such a file-based function.

PR #110, opened 2016-11-17
FIX #109

### LINKED ISSUE #109: Use batch predictions instead of loop over predictions

mlindauer:
* use batch_predictions()
* remove unncessary unit tests

### DIFF
--- a/smac/epm/rf_with_instances.py
+++ b/smac/epm/rf_with_instances.py
@@ -138,31 +138,9 @@ class RandomForestWithInstances(AbstractEPM):
             raise ValueError('Rows in X should have %d entries but have %d!' %
                              (self.types.shape[0], X.shape[1]))
 
-        means = np.ndarray((X.shape[0], 1))
-        vars = np.ndarray((X.shape[0], 1))
-        for i, x in enumerate(X):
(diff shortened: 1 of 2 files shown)


smac/epm/rf_with_instances.py
```diff
@@ -138,31 +138,9 @@ def predict(self, X):
...
-            Predictive mean
-        var : float
-            Predictive variance
-        """
-        mean, var = self.rf.predict(x)
-
-        return mean, var
+        means, vars = self.rf.batch_predictions(X)
+        
+        return np.reshape(means,[means.shape[0], 1]), np.reshape(vars, [vars.shape[0], 1])
```
mfeurer: What about `means.reshape((-1, 1))` and `vars.reshape((-1, 1))`?

mlindauer: Why is this better?

mfeurer: I think it's easier to read. By using -1 instead of the shape attribute it's clear that this axis will be as large as needed in order to fit all attributes.

mlindauer: fixed

PR #113, opened 2016-11-18
MAINT extend runhistory unit test for picklability

mfeurer:
This PR adds a unit test to ensure that the runhistory (with simple data) can be pickled.


mfeurer:
@mlindauer @KEggensperger this is ready for inspection. It simply provides a unit test to ensure that the runhistory can be pickled.

PR #116, opened 2016-11-22
ADD initial design with multiple configuration read from previous

mlindauer:
* ADD methods to read trajectories from disc (in aclib2 json format)
* ADD initial design based on multiple configurations using intensification to determine initial incumbent
* ADD adaption of SMAC facade to support the new initial design
* FIX adaptive capping slack use -- I know this should not be in this PR, sorry!
* MAINT move check "first run crashed" from initial design into TAE 
* FIX maximal wc limit by pynisher to (2**16)-1
* FIX single config initial design adds also the run to the runhistory
* MAINT argument to disable traj logging

@mfeurer @KEggensperger Could you please review this PR. If you approve the changes, I will add further unit tests.


smac/initial_design/multi_config_initial_design.py
```diff
@@ -0,0 +1,90 @@
...
+            -------
+            incumbent: Configuration()
+                initial incumbent configuration
+        '''
+        
+        configs = self.get_configs()
+        inc, inc_perf = self.intensifier.intensify(challengers=configs[1:], 
+                                              incumbent=configs[0], 
+                                              run_history=self.runhistory, 
+                                              aggregate_func=self.aggregate_func)
```
KEggensperger: No timebound? Intensification should also respect the limits defined in the scenario.

smac/initial_design/multi_config_initial_design.py
```diff
@@ -0,0 +1,90 @@
...
+            runs the initial design given the configurations from self.get_configs
+            
+            Returns
+            -------
+            incumbent: Configuration()
+                initial incumbent configuration
+        '''
+        
+        configs = self.get_configs()
+        inc, inc_perf = self.intensifier.intensify(challengers=configs[1:],
```
KEggensperger: Intensify will not add an initial incumbent to the trajectory, will it? So this should be done here.

smac/facade/smac_facade.py
```diff
@@ -150,12 +158,43 @@ def __init__(self,
...
+                                      cutoff=scenario.cutoff,
+                                      deterministic=scenario.deterministic,
+                                      run_obj_time=scenario.run_obj == "runtime",
+                                      instance_specifics=scenario.instance_specific)
+
         # initial design
-        if initial_design is None:
+        if initial_design is not None and initial_configurations is not None:
+            raise ValueError(
+                "Either use initial_design or initial_configuration; but not both")
```
KEggensperger: "initial_configuration" -> "initial_configurations"

smac/initial_design/multi_config_initial_design.py
```diff
@@ -0,0 +1,90 @@
...
+                         stats=stats,
+                         traj_logger=traj_logger,
+                         runhistory=runhistory,
+                         rng=rng)
+        
+        self.get_configs = get_configs
+        self.intensifier = intensifier
+        self.aggregate_func = aggregate_func
+
+    def run(self) -> Configuration:
```
KEggensperger: Other initial designs terminate if the first run crashes. Is there any way to ensure this here too?

smac/smac_cli.py
```diff
@@ -58,13 +59,26 @@ def main_cli(self):
...
                         cs=scen.cs,
                         aggregate_func=aggregate_func)
-
+            
+        initial_configs = None
+        if args_.warmstart_trajectory:
+            initial_configs = [scen.cs.get_default_configuration()]
+            for traj_fn in args_.warmstart_trajectory:
+                trajectory = TrajLogger.read_traj_aclib_format(fn=traj_fn, cs=scen.cs)
+                initial_configs.append(trajectory[-1]["incumbent"])
```
KEggensperger: If this option only reads in the final incumbent of a trajectory is "warmstart_trajectory" the right name for it? Shouldn't it rather be "warmstart_finalIncumbent", "warmstart_Incumbents" or read in the complete trajectory?

smac/initial_design/multi_config_initial_design.py
mlindauer: Yes, intensify adds entries to the trajectory.

smac/initial_design/multi_config_initial_design.py
mlindauer: The timebound is related to the overhead of the model training. 
To respect the scenario budgets, the TAE is aware of the budgets.

smac/initial_design/multi_config_initial_design.py
mlindauer: good question. right now, I have no idea how to ensure this here.

smac/initial_design/multi_config_initial_design.py
KEggensperger: Okay, but it does not add the initial incumbent at time 0, which should be configs[0:]. Intensify can then add the next incumbent.

smac/initial_design/multi_config_initial_design.py
mlindauer: I moved this check into the execute_ta_run such that there is only one place that check this.

smac/initial_design/multi_config_initial_design.py
mlindauer: thanks. fixed

smac/facade/smac_facade.py
mlindauer: thanks. fixed

smac/smac_cli.py
mlindauer: thanks. fixed

smac/utils/io/traj_logging.py
```diff
@@ -146,3 +153,66 @@ def _add_in_aclib_format(self, train_perf, incumbent_id, incumbent,
...
+                configuration as a list of "str='str'"
+            cs: ConfigurationSpace
+                Configuration Space to translate dict object into Confiuration object
+        '''
+        config_dict = {}
+        for param in config_list:
+            k,v = param.split("=")
+            v = v.strip("'")
+            hp = cs.get_hyperparameter(k)
+            if isinstance(hp, FloatHyperparameter):
```
mfeurer: We also need a check for Constants which have a float/int value. Maybe this should rather be handled in the ConfigSpace?

smac/initial_design/multi_config_initial_design.py
```diff
@@ -0,0 +1,113 @@
...
+        '''
+        configs = self.get_configs()
+
+        self.traj_logger.add_entry(train_perf=2**31,
+                                   incumbent_id=1,
+                                   incumbent=configs[0])
+
+        if len(set(configs)) > 1:
+            # at least two configuration differ
+            # otherwise intensify will not run any configuration
```
mlindauer: yes. if challenger and incumbent are identical, itensify skip this challenger (and also won't perform any new run of the incumbent)

smac/tae/execute_ta_run.py
```diff
@@ -122,6 +122,11 @@ def start(self, config, instance,
                                                           seed=seed,
                                                           instance_specific=instance_specific,
                                                           **additional_arguments)
+        
+        if self.stats.ta_runs == 0 and status in [StatusType.CRASHED, StatusType.ABORT]:
+            self.logger.critical("First run crashed -- Abort")
+            sys.exit(1)
```
mlindauer: Thanks. fixed

smac/utils/constants.py
```diff
@@ -1,2 +1,3 @@
 MAXINT = 2 ** 31 - 1
-MINIMAL_COST_FOR_LOG = 0.00001 
\ No newline at end of file
+MINIMAL_COST_FOR_LOG = 0.00001 
+MAX_CUTOFF = 65535
```
mlindauer: probably the PR text is wrong. uint16 should be at most 2**16-1

smac/utils/io/traj_logging.py
```diff
@@ -30,29 +31,34 @@ def __init__(self, output_dir, stats):
...
-
-        self.old_traj_fn = os.path.join(output_dir, "traj_old.csv")
-        if not os.path.isfile(self.old_traj_fn):
-            with open(self.old_traj_fn, "w") as fp:
-                fp.write(
-                    '"CPU Time Used","Estimated Training Performance","Wallclock Time","Incumbent ID","Automatic Configurator (CPU) Time","Configuration..."\n')
-
-        self.aclib_traj_fn = os.path.join(output_dir, "traj_aclib2.json")
+        if output_dir is None:
+            self.logger.warn("No output directory for trajectory logging specified -- trajectory will not be logged.")
```
mlindauer: my feeling would be "warning", but I'm also fine with info. fixed.

smac/utils/io/traj_logging.py
mlindauer: It would be great if SMAC does not need to implement the logic to convert types of parameter values, but the config space could fix all types since it is more aware of all its hyperparameter types. 
This would also prevent that we will miss to update this code in the future when we would add a new hyperparameter type in the configspace.

smac/facade/smac_facade.py
```diff
@@ -43,9 +47,10 @@ def __init__(self,
                  runhistory: RunHistory=None,
                  intensifier: Intensifier=None,
                  acquisition_function: AbstractAcquisitionFunction=None,
-                 model:AbstractEPM=None,
+                 model: AbstractEPM=None,
                  runhistory2epm: AbstractRunHistory2EPM=None,
                  initial_design: InitialDesign=None,
+                 initial_configurations: typing.List[Configuration]=None,
```
mlindauer: This was also my initial feeling, but the initial designs need a lot of objects which will be created in the SMAC facade. If I would create the initial design object in the smac_cli, I would need to duplicate a lot of code from the SMAC facade. Since I would like to avoid this, I would prefer to keep it this way.

smac/initial_design/multi_config_initial_design.py
```diff
@@ -0,0 +1,113 @@
...
+class MultiConfigInitialDesign(InitialDesign):
+
+    def __init__(self,
+                 tae_runner: ExecuteTARun,
+                 scenario: Scenario,
+                 stats: Stats,
+                 traj_logger: TrajLogger,
+                 runhistory: RunHistory,
+                 rng: np.random.RandomState,
+                 get_configs: typing.Callable,
```
mlindauer: 1. This is motivated by the code in the single_config_intial_design, where we have the `_select_configuration` method to get the configuration. So, it is more consistent to do it a similar way.
2. If we want to later implement another initial design with an internal function using the ConfigSpace, we could simply implement a corresponding function.

smac/facade/smac_facade.py
mfeurer: Agreed.

smac/initial_design/multi_config_initial_design.py
mfeurer: I see, could you please make the comment more elaborate as this is (at least to me) not intuitive.

smac/tae/execute_func.py
```diff
@@ -79,6 +79,9 @@ def run(self, config, instance=None,
         arguments = {'logger': logging.getLogger("pynisher"),
                      'wall_time_in_s': cutoff,
                      'mem_in_mb': memory_limit}
+        
+        if cutoff is not None and cutoff > MAX_CUTOFF:
+            self.logger.warn("Pynisher may only limit the cutoff for up to %d" %(MAX_CUTOFF))
```
mfeurer: I see. But we should avoid any behaviour that might differ too much on different systems. Therefore, we should either check the limit dynamically, or fail if a user tries to hit the bar

smac/utils/io/traj_logging.py
mfeurer: Agreed. On the other hand, the trajectory is stored as a json. Type information should not get lost when using the json format. It looks odd to me that we need this at all.

mfeurer: Agreed. On the other hand, the trajectory is stored as a json. Type information should not get lost when using the json format. It looks odd to me that we need this at all.

mlindauer: the trajectory format follows the format spec for AClib2. I would really like to avoid to start the discussion about this AClib2 format again.
If we believe that this format is suboptimal, we could also dump the runhistory in a further "SMAC3 format".

smac/tae/execute_func.py
mlindauer: So, what is your proposed solution?

```
if cutoff is not None and cutoff > MAX_CUTOFF:
            raise ValueError("Pynisher may only limit the cutoff for up to %d" %(MAX_CUTOFF))
```

??

smac/utils/io/traj_logging.py
mfeurer: Agreed. You could add this issue for ACLIB 2.1 or ACLIB 3 ;)

smac/initial_design/multi_config_initial_design.py
mlindauer: ok, you convinced me. I fixed this: now the argument is a list

smac/tae/execute_func.py
mfeurer: I do not have one. Remove this piece of code and create a separate issue?

mlindauer: Ok, I removed it. But I wonder who will fix this in the future...

mlindauer:
Ok, I think I solved all the issues raised by @mfeurer . We could now merge, right?

PR #119, opened 2016-11-24
Ci/disable test on osx

mfeurer:
This PR disables the memory limit test for OSX. On the other hand, it adds a check which would notify us if the memory limit test for OSX would mysteriously start working in the future. Furthermore, it removes the `allow_failure` flag from the travis yaml, as we now have (nearly) full OSX compability.


mfeurer:
@KEggensperger @mlindauer we now have a working OSX test :) Please review.

PR #121, opened 2016-11-25
Fix/#118

mlindauer:
@mfeurer please check this PR. I would like to remove the aggregate function in another PR since it will be otherwise mess up this PR.

* ADD two new runhistory functions: compute_all_costs() and incremental_update_cost
* FIX merge_foreign_data updated accordingly
* add unit tests

### DIFF
--- a/smac/runhistory/runhistory.py
+++ b/smac/runhistory/runhistory.py
@@ -45,9 +45,10 @@ class RunHistory(object):
         self.ids_config = {}  # id -> config
         self._n_id = 0
 
-        self.cost_per_config = {} # config_id -> cost
-        self.runs_per_config = {} # config_id -> number of runs
-        
+        self.cost_per_config = {}  # config_id -> cost
+        # runs_per_config is necessary for computing the moving average
+        self.runs_per_config = {}  # config_id -> number of runs
+
         self.aggregate_func = aggregate_func
 
     def add(self, config, cost, time,
@@ -97,7 +98,7 @@ class RunHistory(object):
             store the performance of a configuration across the instances in self.cost_perf_config
             and also updates self.runs_per_config;
             uses self.aggregate_func
-            
+
             Arguments
             --------
             config: Configuration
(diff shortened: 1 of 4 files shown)


smac/runhistory/runhistory.py
```diff
@@ -45,6 +46,7 @@ def __init__(self, aggregate_func):
         self._n_id = 0
 
         self.cost_per_config = {} # config_id -> cost
+        self.runs_per_config = {} # config_id -> number of runs
```
mfeurer: Could you please add a comment that this is necessary to calculate the moving average?

smac/runhistory/runhistory.py
```diff
@@ -103,6 +107,50 @@ def update_cost(self, config):
...
+            Arguments
+            --------
+            config: Configuration
+                configuration to update cost based on all runs in runhistory
+            cost: float
+                cost of new run of config
+        '''
+        
+        config_id = self.config_ids[config]
+        n_runs = self.runs_per_config.get(config_id,0)
```
mfeurer: Please add whitespaces after commas.

smac/runhistory/runhistory.py
```diff
@@ -103,6 +107,50 @@ def update_cost(self, config):
...
+        '''
+            computes the cost of all configurations from scratch
+            and overwrites self.cost_perf_config and self.runs_per_config accordingly;
+            Since we iterate over the runhistory for every configuration,
+            this function can be quite expensive.
+            
+            Arguments
+            ---------
+            instances: typing.List[str]
+                list of instances; if given, cost is only computed with wrt to this instance set
```
mfeurer: with and wrt are redundant

smac/runhistory/runhistory.py
mlindauer: Thanks. Fixed

smac/runhistory/runhistory.py
mlindauer: Thanks. Fixed

smac/runhistory/runhistory.py
mlindauer: Thanks. I applied an automatic pep8 conversion now :-)

PR #127, opened 2016-11-29
FIX pass memory limit to target algorithm executor

### LINKED ISSUE #126: memory limit not passed to the target algorithm runner
I have found that we don't pass the memory limit to a target algorithm runner. This results in the memory limit being ignored without the user noticing this (and the computer starting to SWAP). I'll submit a PR in a few minutes.

mfeurer:
Fixes #126


mlindauer:
actually, I would prefer if the memory limit is directly passed to the TAE constructor and not passed through intensify and initial_design. At least given our current implementation, I don't see a reason why the inital design and intensify have to be aware of a memory limit.

mfeurer:
I can see the argument for the intensification procedure. But why do we then pass the cutoff to the initial design? The same holds true here as well.

mlindauer:
cutoff has to be argument of the run() method of a TAE since we modify it in intensify. Passing it as constructor argument would imply that we pass the cutoff twice which does not make sense imho. Therefore, all methods (including the initial design) that have to call the TAE have to be aware of the cutoff.

mfeurer:
I just had a look at the code again. You basically argue that we should add the memory limit to the constructor and not pass it dynamically? Would we then only add the memory limit to the constructor of each TAE class which allows for a memory limit or to all of them?

mlindauer:
> You basically argue that we should add the memory limit to the constructor and not pass it dynamically?

Yes

> Would we then only add the memory limit to the constructor of each TAE class which allows for a memory limit or to all of them?

Since only the pynisher-related TAEs can use a memory limit right now, I would pass it only to these TAE classes.

mfeurer:
@mlindauer Okay, I did the updates. Currently, I think about removing the memory limit argument from the run method completely? What do you think about this?

examples/rf.py
```diff
@@ -110,21 +110,18 @@ def rfr(cfg, seed):
...
 smac = SMAC(scenario=scenario, rng=np.random.RandomState(42),
-            tae_runner=taf)
+            tae_runner=rfr)
 try:
     incumbent = smac.optimize()
 finally:
     incumbent = smac.solver.incumbent
 
-inc_value = taf.run(incumbent)[1]
+inc_value = rfr(incumbent, 1)
```
mlindauer: If you run the rfr directly without a TAE class,
the call will have not limits. Therefore, the `inc_value` will not be necessarily the same as SMAC has observed it.
I think this will not be an issue for the example, but it will be a bad example for others.

mlindauer:
> Currently, I think about removing the memory limit argument from the run method completely? What do you think about this?

yes, I agree that this should be removed.

mfeurer:
Okay, I updated the PR accordingly. The way I have to obtain the target algorithm in the example suggests that we add a getter function to the SMAC object (in a new PR), what do you think about this?

mlindauer:
> Okay, I updated the PR accordingly. The way I have to obtain the target algorithm in the example suggests that we add a getter function to the SMAC object (in a new PR), what do you think about this?

Sounds like a good idea

PR #129, opened 2016-12-01
Minimize SMAC overhead

mlindauer:
the title of the branch is a bit misleading. My main goal was to speed up the local search but I also changed other stuff to minimize the overhead of SMAC. Profiling SMAC on branin "with instances" (see aclib2), SMAC spends now more time in the subprocess than in its own functions (for the first 2000 function evaluations), even tough branin is such a cheap function.

* initialize local search with best configs from random search
* reduce number of trees in RF to 10 (as in SMAC2)
* run at least 2 configurations per intensify call (as in SMAC2)
* adaptively adjust number of local search iterations based on intensify stats
* more efficient runhistory data structure to get runs for a given

### DIFF
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -270,11 +270,11 @@ class Intensifier(object):
                     # challenger is not worse, continue
                     N = 2 * N
 
-            if chall_indx >= 1 and num_run > self.run_limit:
+            if chall_indx > 1 and num_run > self.run_limit:
                 self.logger.debug(
                     "Maximum #runs for intensification reached")
                 break
-            elif chall_indx >= 1 and time.time() - self.start_time - time_bound >= 0:
+            elif chall_indx > 1 and time.time() - self.start_time - time_bound >= 0:
                 self.logger.debug("Timelimit for intensification reached ("
                                   "used: %f sec, available: %f sec)" % (
                                       time.time() - self.start_time, time_bound))
@@ -285,5 +285,7 @@ class Intensifier(object):
         inc_perf = aggregate_func(incumbent, run_history, inc_runs)
         self.logger.info("Updated estimated performance of incumbent on %d runs: %.4f" % (
             len(inc_runs), inc_perf))
+        
+        self.stats.update_average_configs_per_intensify(n_configs=chall_indx)
 
         return incumbent, inc_perf
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -125,7 +125,6 @@ class Intensifier(object):
                     "Challenger was the same as the current incumbent; Skipping challenger")
                 continue
             
-            chall_indx += 1
             self.logger.debug("Intensify on %s", challenger)
             if hasattr(challenger, 'origin'):
                 self.logger.debug(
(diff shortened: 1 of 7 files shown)


smac/intensification/intensification.py
```diff
@@ -270,11 +270,11 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
                     # challenger is not worse, continue
                     N = 2 * N
 
-            if chall_indx >= 1 and num_run > self.run_limit:
+            if chall_indx > 1 and num_run > self.run_limit:
```
mfeurer: We should consider counting the challenger index differently to make sure that we run two new configurations. Currently, it can happen that a challenger is identical to the incumbent or that we pick a challenger for which all runs are completed. Doing so would counteract the goal of this change.

smac/runhistory/runhistory.py
```diff
@@ -88,8 +94,13 @@ def add(self, config, cost, time,
 
         k = RunKey(config_id, instance_id, seed)
         v = RunValue(cost, time, status, additional_info)
-
         self.data[k] = v
+        
+        #also add to fast data structure
+        is_k = InstSeedKey(instance_id,seed)
```
mfeurer: Could you please run your pep8-scripts on this file ? ;)

smac/runhistory/runhistory.py
```diff
@@ -88,8 +94,13 @@ def add(self, config, cost, time,
...
+        self._configid_to_inst_seed[config_id] = self._configid_to_inst_seed.get(config_id,{})
+        self._configid_to_inst_seed[config_id][is_k] = v
```
mfeurer: What's the reason for storing the value here?

smac/intensification/intensification.py
mlindauer: Thanks. Fixed.

smac/runhistory/runhistory.py
mlindauer: Done.

smac/runhistory/runhistory.py
mlindauer: I needed a bit to figure out what you mean, but I think it should be fixed now. If not, please explain in more depth the issue here.

smac/intensification/intensification.py
```diff
@@ -188,6 +194,10 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
 
                 # Line 10
                 missing_runs = list(inc_inst_seeds - chall_inst_seeds)
+                
+                if (not first_run) and missing_runs:
+                    first_run = True
+                    chall_indx += 1
```
mfeurer: I think this should be further down where the algorithm is actually called. Right now it could happen that two challengers have already been run on all instances and thus not considered. In that cases we want to continue iterating the challengers until we find one that was not yet evaluated, rights?

mlindauer: Sorry, I don't get it. 
If `to_run` is not empty, we will have new runs.
`to_run` is a truncated list of `missing_runs`. I don't see a case in which `missing_runs` is not empty but `to_run` is empty since `N` is at least 1.
So, if `missing_runs` is not empty, we will have at least one run of the challenger.

However, there could an issue with adaptive capping. If we picked the challenger for the second time, adaptive capping could prevent us to have further runs on this challenger.

Therefore, I moved the code further down.

PR #132, opened 2016-12-06
ADD reading of pcs file with new format

mlindauer:


mlindauer:
to fix the unit tests, we would need a new pypi release of `ConfigSpace`

mfeurer:
I will push the ConfigSpace to pypi either on Friday or when I'm back in Germany. I only wanted to know whether the new pcs format can be used within SMAC, which seems to be the case.

mfeurer:
Did you check that SMAC accepts the new ordinal hyperparameter? I think you need to adapt at least [this piece of code](https://github.com/automl/SMAC3/blob/master/smac/utils/util_funcs.py).

mlindauer:
Unfortunately, I don't have any "real-world" example with ordinal parameters yet.
I created a simple example and indeed the code you mentioned failed.
I fixed it. 
I assume ordinals will be handled in the same way as floats and integer parameters?

btw: why is this function hidden in `utils`. Imho this should live in `smac.configspace`

mfeurer:
I think we should handle ordinals as integer hyperparameters. Could you please check whether my latest fixes to the ConfigSpace make this PR work?

The function is hidden in `utils` because you put it there (see history) ;) I think this is should be part of SMAC, because it is very specific to the RF that it wants a vector of integers to represent the hyperparameter types.

mlindauer:
I posted the new error again in the ConfigSpace issue tracker.

mlindauer:
@mfeurer After we have a new configspace version on pypi, we could merge this PR.

Issue #133, opened 2016-12-08
get_runs_for_config should return only self made runs


mlindauer:
Since intensify uses the function `get_runs_for_config()` it will get all runs for a given configuration independent of its origin (self made or loaded with wamstarting). This can cause several issues, such as

1. the incumbent has runs on instances that are not part of the current scenario
1. the challenger has more runs than the incumbent

Both problems will crash SMAC.
I will provide a fix where the method `RunHistory.add()` gets an argument to not add runs to the "fast" data structure `_configid_to_inst_seed` which is used by `get_runs_for_config()`.

PR #134, opened 2016-12-08
Fix/#133

mlindauer:
Fix for #133 
and adding feature to deactivate output_dir which can be annoying and confusing if merge_foreign_data is used.

Downside of this fix is that the externally loaded runhistory data is only used for training the EPM.  Previously, I hoped that we could also use the loaded data to have more evidence in `intensify` if the instance set is the same, but I don't see how this can be implemented in a consistent way using our current `intensify`.

### DIFF
--- a/smac/runhistory/runhistory.py
+++ b/smac/runhistory/runhistory.py
@@ -1,6 +1,6 @@
 import collections
 import json
-import numpy
+import numpy as np
 import typing
 
 from smac.configspace import Configuration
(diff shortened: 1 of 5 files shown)


mlindauer:
>  why it is confusing when merge_foreign_data is used.

sorry, for this missing explanation.
Each time, we parse a scenario, an output directory is created. Since we have to parse several scenario files, if we use `merge_foreign_data`, more than one output directory is created and also shown in the stdout logs.

mfeurer:
A pass of the pep8 tool, and then it's fine ;)

Issue #19, opened 2016-07-20
Limit maxR


KEggensperger:
Originally reported by: **Matthias Feurer (Bitbucket: [mfeurer](https://bitbucket.org/mfeurer), GitHub: [mfeurer](https://github.com/mfeurer))**

---

Currently, maxR is set to 2000. This might be useful for algorithm configuration, but it isn't for machine learning with five-fold cross-validation. We need to add this as an argument.

---
- Bitbucket: https://bitbucket.org/aadfreiburg/smac3/issue/19

mfeurer:
Since #138, there is also a minR flag which needs to be added.

PR #138, opened 2016-12-19
ADD minR

mfeurer:
The minR argument to intensify allows us to perform random search in the presence of instances by setting minR = maxR.

### DIFF
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -28,8 +28,8 @@ class Intensifier(object):
     '''
 
     def __init__(self, tae_runner, stats, traj_logger, rng, instances=None,
-                 instance_specifics={}, cutoff=MAX_CUTOFF, deterministic=False,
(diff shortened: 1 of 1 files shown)


smac/intensification/intensification.py
```diff
@@ -134,51 +137,59 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
...
+                            config=incumbent,
+                            instance=next_instance,
+                            seed=next_seed,
+                            cutoff=self.cutoff,
+                            instance_specific=self.instance_specifics.get(next_instance, "0"))
+
+                        num_run += 1
+                    else:
+                        self.logger.debug(
+                            "No further instance-seed pairs for incumbent available.")
```
mlindauer: I would expect a `break` here.

mfeurer: Good catch! Thanks. Fixed.

mfeurer:
I just added your request for adding minR to the scenario file to issue #19 and will work on this after this PR is merged.

PR #140, opened 2016-12-19
FIX #139

### LINKED ISSUE #139: SingleConfigInitialDesign adds run twice to the runhistory
As the TAE adds the runs to the runhistory, the SingleConfigInitialDesign should no longer add a run to the runhistory.

mfeurer:
Fixes issue #139


mfeurer:
I just figured out that the initial design doesn't need the runhistory any more. Should I also remove it in this PR?

mlindauer:
yes

Issue #125, opened 2016-11-25
SMAC crashes if first run of solution_quality scenario is a timeout


mlindauer:
If the first run of SMAC3 is a timeout and we optimize for solution quality (cost),
runhistory2epm does not provide any data point for the RF 
and the RF will crash with an empty X matrix.

```
  File "/work/ws/nemo/fr_tl1023-aclib-0/aclib2/configurators/smac3/smac/epm/rf_with_instances.py", line 114, in train
    data = pyrfr.regression.numpy_data_container(self.X, y, self.types)
  File "pyrfr/regression_source.pyx", line 125, in pyrfr.regression.numpy_data_container.__cinit__ (pyrfr/regression.cpp:3575)
IndexError: Out of bounds on buffer access (axis 0)
```

As specified in smac_facade, we use only "successful" for soluation quality scenarios:

```
            elif scenario.run_obj == 'quality':
                runhistory2epm = RunHistory2EPM4Cost\
                    (scenario=scenario, num_params=num_params,
                     success_states=[StatusType.SUCCESS, ],
                     impute_censored_data=False, impute_state=None)
```

I think that a neat solution would be if SMAC could also handle "unsuccessful" runs (TIMEOUT, CRASH) in solution quality scenarios.
For example, we could try to force the user to provide an imputation value of unsuccessful runs in solution quality scenarios.

btw: the issue occurs on the svm_mnist scenario in AClib2.

mlindauer:
@mfeurer told me that we should look in the paper "Bayesian Optimization with Unknown Constraints" to solve this issue in a clean way.
However, I would prefer to have an interim solution in the meantime.

mfeurer:
Sorry, my answer was incorrect. If the first run is a timeout, SMAC should do random search until it finds a feasible solution (only then it makes sense to model something). I will go ahead and implement this.

Issue #131, opened 2016-12-06
Crash if first run is memory out


mlindauer:
SMAC crashes if the first run of the TA is memory out.
Right now, the first run has to either a successful run or a "real" timeout (time > cutoff). 
Even if the memout is encoded as a timeout, SMAC will crash because `runhistory2epm` will only pass "real timeouts" to the EPM.

mfeurer:
This is more or less a duplicate of #125, would be fixed by my proposed solution.

PR #145, opened 2016-12-20
Fix issues #125 and #131

mfeurer:
SMAC does not longer crash if the first run was a memout or a timout. More general, if an empty array X is passed to choose_next, a random set of configurations is returned.

Fixes issues #125 and #131.

### DIFF
--- a/smac/smbo/smbo.py
+++ b/smac/smbo/smbo.py
@@ -176,7 +176,9 @@ class SMBO(BaseSolver):
             List of 2020 suggested configurations to evaluate.
         """
         if X.shape[0] == 0:
-            return [x[1] for x in self._get_next_by_random_search()]
+            # Only return a single point to avoid an overly high number of
+            # random search iterations
+            return [x[1] for x in self._get_next_by_random_search(num_points=1)]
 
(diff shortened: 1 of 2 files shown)


smac/smbo/smbo.py
```diff
@@ -175,6 +175,9 @@ def choose_next(self, X, Y,
         list
             List of 2020 suggested configurations to evaluate.
         """
+        if X.shape[0] == 0:
+            return [x[1] for x in self._get_next_by_random_search()]
```
KEggensperger: Could you please specify num_points to improve readability?

mfeurer:
I just added the requested keyword argument to the call. Could you please check again @KEggensperger

mlindauer:
Shouldn't have X not at least two points such that we can learn something from the EPM?
With only one point in X we predict only a constant.

mfeurer:
Isn't SMAC doing exactly that by default? Running the initial design to obtain a single point, then using the random forest to suggest a new point? I think this is a more general issue that we need to address.

mlindauer:
Fair enough...

Issue #81, opened 2016-10-12
Add mapping from status type to human readable text


mfeurer:
Currently, the status is an integer between 1 and 5. It cannot be easily displayed to the user without explaining it. It would be helpful to add a mapping back to a human readable string, so that I can do:

```
>>> print(mapping[status])
Memout
```

In the course of implementing this, we could also start using the [Enum](https://docs.python.org/3/library/enum.html#enum.Enum) class which was added in python3.4.

mlindauer:
Yes, very good idea to use Enum

shukon:
Enum affects the JSON-serialization. It is implemented in #151, using a customized JSON-Encoder. I assume, that the JSON-files are only ever read in by smac (more precise: the runhistory)?
mapping is realized by "status.name", which returns the status in human-readable form (-> "SUCCESS").

PR #151, opened 2017-01-17
CHANGE StatusType to Enum

shukon:
Implementing issue #81, adding human-readable mapping of StatusTypes. The originally wished for

    \>\>\> print(mapping[status])

is realized through [Enum](https://docs.python.org/3/library/enum.html#enum.Enum) as

    \>\>\> print(status.name)
 
JSON-serialization of runhistories is affected: status is now saved by customized JSON-Encoder.
This might affect the possibility of reading in the JSON-file from another language/without the object_hook.

### DIFF
--- a/smac/tae/execute_ta_run.py
+++ b/smac/tae/execute_ta_run.py
@@ -1,6 +1,7 @@
 import sys
 import logging
 import math
+from enum import Enum
 
 import numpy as np
 
@@ -12,7 +13,7 @@ __email__ = "lindauer@cs.uni-freiburg.de"
 __version__ = "0.0.1"
 
 
-class StatusType(object):
+class StatusType(Enum):
 
     """
         class to define numbers for status types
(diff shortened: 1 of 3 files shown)


smac/tae/execute_ta_run.py
```diff
@@ -23,6 +24,17 @@ class to define numbers for status types
...
     MEMOUT = 5
 
+    def enum_hook(obj):
+        """
+        hook function passed to json-deserializer as "object_hook"
+        """
+        if "__enum__" in obj:
+            # object is marked as enum
+            name, member = obj["__enum__"].split(".")
+            return getattr(globals()[name], member)
```
mfeurer: We only want to parse a single enum, all other enums should not be parsed. Thus, I think this should be more restrictive.

shukon: Integrated a simple check for the name of the enum. Should a TypeError be raised if another enum than StatusType is detected?

`if name == "StatusType": return getattr(globals()[name], member)`
`else: return obj`

PR #152, opened 2017-01-18
Test intensify

mlindauer:
See commit messages for what I have done.
Overall, I could not find any obvious bugs in intensify.
Please check carefully whether you find anything.


mlindauer:
one comment: I did not understand why `Intensifier()` was allowed to get instances=None. We fix this in the constructor by 
```
        if instances is None:
            instances = []
```

However, we would never add a run for an incumbent because of:

```
    available_insts = (self.instances - inc_inst)
[...]
    if available_insts:
```

So, I would say that `Intensifier` needs at least one instance as an input.

smac/intensification/intensification.py
```diff
@@ -307,7 +163,242 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
...
+                either challenger or incumbent
+        '''
+        # at least one run of challenger
+        # to increase chall_indx counter
+        first_run = False
+
+        # Line 8
+        N = max(1, self.minR)
+        
+        inc_inst_seeds = set(run_history.get_runs_for_config(incumbent))
```
mfeurer: This was a lambda expression before. I guess it's not needed?

smac/intensification/intensification.py
```diff
@@ -307,7 +163,242 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
...
+                self.logger.debug("Add run of challenger")
+                status, cost, dur, res = self.tae_runner.start(
+                    config=challenger,
+                    instance=instance,
+                    seed=seed,
+                    cutoff=cutoff,
+                    instance_specific=self.instance_specifics.get(instance, "0"))
+                self._num_run += 1
+
+                if status == StatusType.ABORT:
```
mfeurer: This is new, isn't it? But if the signal is abort, shouldn't SMAC stop completely?

smac/intensification/intensification.py
```diff
@@ -118,185 +124,35 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
...
-                    num_run += 1
-
-                # we cannot use inst_seed_pairs here since we could have less runs
-                # for the challenger due to the adaptive capping
-                chall_inst_seeds = set(map(lambda x: (
-                    x.instance, x.seed), run_history.get_runs_for_config(challenger)))
-                
-                if len(chall_inst_seeds) < len(inst_seed_pairs):
-                    #challenger got not enough runs because of exhausted config budget
-                    self.logger.debug("No (or not enough) valid runs for challenger.")
```
mfeurer: It looks like this statement didn't make it to the new code.

smac/intensification/intensification.py
```diff
@@ -307,7 +163,242 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
...
+
+        # Line 15
+        if chal_perf > inc_perf and len(chall_runs) >= self.minR:
+            # Incumbent beats challenger
+            self.logger.debug("Incumbent (%.4f) is better than challenger (%.4f) on %d runs." % (
+                inc_perf, chal_perf, len(chall_runs)))
+            return incumbent
+
+        # Line 16
+        if len(chall_runs) == len(inc_runs):
```
mfeurer: This isn't what is written in the docstring. The docstring states 'at least as many', thus it should be a `>=`.

smac/intensification/intensification.py
```diff
@@ -307,7 +163,242 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
...
+
+            inst_seed_pairs = list(inc_inst_seeds - set(missing_runs))
+
+            inc_sum_cost = sum_cost(config=incumbent, instance_seed_pairs=inst_seed_pairs,
+                                    run_history=run_history)
+
+            # Line 12
+            for instance, seed in to_run:
+                # Run challenger on all <config,seed> to run
+                if self.run_obj_time:
```
mfeurer: I think this should be an extra function, because it is very specific for runtime optimization and interrupts the reading flow for the racing algorithm.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+                    seed=None,
+                    additional_info=None)
+
+        conf = intensifier._compare_configs(incumbent=self.config1,
+                                            challenger=self.config2,
+                                            run_history=self.rh,
+                                            aggregate_func=average_cost)
+
+        # challenger has enough runs and is better
+        self.assertTrue(conf == self.config2, "conf: %s" % (conf))
```
mfeurer: How about `self.assertEqual(conf, self.conf2)`? Or wasn't the output in case of inequality nice to read?

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+                    seed=None,
+                    additional_info=None)
+
+        conf = intensifier._compare_configs(incumbent=self.config1,
+                                            challenger=self.config2,
+                                            run_history=self.rh,
+                                            aggregate_func=average_cost)
+
+        # challenger worse than inc
+        self.assertTrue(conf == self.config1, "conf: %s" % (conf))
```
mfeurer: Same here.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+                    seed=None,
+                    additional_info=None)
+
+        conf = intensifier._compare_configs(incumbent=self.config1,
+                                            challenger=self.config2,
+                                            run_history=self.rh,
+                                            aggregate_func=average_cost)
+
+        # challenger worse than inc
+        self.assertTrue(conf is None, "conf: %s" % (conf))
```
mfeurer: How about `assertIsNone(conf)`?

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def test_race_challenger(self):
+        '''
+           test _race_challenger without adaptive capping
+        '''
+
+        def target(x):
+            time.sleep(min(1, (x['a'] + 1) / 1000.))
```
mfeurer: Why is this sleep necessary?

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+                    status=StatusType.SUCCESS, instance_id=1,
+                    seed=None,
+                    additional_info=None)
+
+        inc = intensifier._race_challenger(challenger=self.config2,
+                                           incumbent=self.config1,
+                                           run_history=self.rh,
+                                           aggregate_func=average_cost)
+
+        self.assertTrue(inc == self.config2)
```
mfeurer: Same as above.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+
+        # config2 should have a timeout (due to adaptive capping)
+        # and config1 should still be the incumbent
+        inc = intensifier._race_challenger(challenger=self.config2,
+                                           incumbent=self.config1,
+                                           run_history=self.rh,
+                                           aggregate_func=average_cost)
+
+        # self.assertTrue(False)
+        self.assertTrue(inc == self.config1)
```
mfeurer: Same as above.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+        self.assertTrue(
+            self.rh.get_cost(self.config2) < 2, self.rh.get_cost(self.config2))
```
mfeurer: Why not `assertLess(self.rh.get_cost(self.config2), 2)`?

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+
+        # self.assertTrue(False)
+        self.assertTrue(inc == self.config2)
+        self.assertTrue(
+            self.rh.get_cost(self.config2) ==1, self.rh.get_cost(self.config2))
+        
+
+        # get data for config2 to check that the correct run was performed
+        runs = self.rh.get_runs_for_config(self.config2)
+        self.assertTrue(len(runs)==10)
```
mfeurer: Same as above.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+
+        # tie on first instances and then challenger should always win
+        # and be returned as inc
+        inc = intensifier._race_challenger(challenger=self.config2,
+                                           incumbent=self.config1,
+                                           run_history=self.rh,
+                                           aggregate_func=average_cost)
+
+        # self.assertTrue(False)
+        self.assertTrue(inc == self.config2)
```
mfeurer: Same as above.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+        self.assertTrue(
+            self.rh.get_cost(self.config2) ==1, self.rh.get_cost(self.config2))
```
mfeurer: Same as above.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+        
+        # get data for config2 to check that the correct run was performed
+        runs = self.rh.get_runs_for_config(self.config2)
+        self.assertTrue(len(runs)==10)
+        
+        seeds = sorted([r.seed for r in runs])
+        self.assertTrue(seeds == list(range(10)), seeds)
```
mfeurer: These asserts can be simplified as well.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+            deterministic=True)
+
+        intensifier._add_inc_run(incumbent=self.config1, run_history=self.rh)
+        self.assertTrue(len(self.rh.data) == 1, self.rh.data)
+        
+        # since we assume deterministic=1, 
+        # the second call should not add any more runs 
+        # given only one instance
+        intensifier._add_inc_run(incumbent=self.config1, run_history=self.rh)
+        self.assertTrue(len(self.rh.data) == 1, self.rh.data)
```
mfeurer: Same as above.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,351 @@
...
+        
+        intensifier._add_inc_run(incumbent=self.config1, run_history=self.rh)
+        self.assertTrue(len(self.rh.data) == 2, self.rh.data)
+        runs = self.rh.get_runs_for_config(config=self.config1)
+        # exactly one run on each instance
+        self.assertTrue(1 in [runs[0].instance, runs[1].instance])
+        self.assertTrue(2 in [runs[0].instance, runs[1].instance])
+        
+        intensifier._add_inc_run(incumbent=self.config1, run_history=self.rh)
+        self.assertTrue(len(self.rh.data) == 3, self.rh.data)
```
mfeurer: asserts here can be simplified as well.

smac/intensification/intensification.py
```diff
@@ -65,7 +64,8 @@ def __init__(self, tae_runner, stats, traj_logger, rng, instances=None,
         if instances is None:
             instances = []
         self.instances = set(instances)
-        self.instance_specifics = {} if instance_specifics is None else instance_specifics
+        self.instance_specifics = {
+        } if instance_specifics is None else instance_specifics
```
mlindauer: fixed

smac/intensification/intensification.py
mlindauer: Since we directly return the incumbent now if we had an aborted run,
we do not need this case anymore.

smac/intensification/intensification.py
mlindauer: correct.

smac/intensification/intensification.py
mlindauer: Done

smac/intensification/intensification.py
mlindauer: Yes, it is new and makes the code later easier.
Right now, the ABORT only stops the intensification of the current challenger. 

How, do we want to exit if SMAC saw an ABORT?
Directly, sys.exit(42)?

smac/intensification/intensification.py
mlindauer: Thanks. fixed

mlindauer:
I fixed all things but one -- left a comment above.

> I wonder what the 20% of untested code are, though.

Look into https://coveralls.io/builds/9740163 and you will know.

smac/intensification/intensification.py
mfeurer: One would like to shutdown gracefully I assume. Maybe stop intensification and add a check for ABORT at the beginning or end of the SMBO loop? The first of these solutions would also handle the case that the first run was aborted.

test/test_intensify/test_intensify.py
```diff
@@ -0,0 +1,396 @@
...
+
+        intensifier._add_inc_run(incumbent=self.config1, run_history=self.rh)
+        self.assertEqual(len(self.rh.data), 1, self.rh.data)
+        
+        intensifier._add_inc_run(incumbent=self.config1, run_history=self.rh)
+        self.assertEqual(len(self.rh.data), 2, self.rh.data)
+        runs = self.rh.get_runs_for_config(config=self.config1)
+        # exactly one run on each instance
+        self.assertTrue(1 in [runs[0].instance, runs[1].instance])
+        self.assertTrue(2 in [runs[0].instance, runs[1].instance])
```
mfeurer: Sorry, I didn't see this yesterday. How about `self.assertIn()`?

smac/intensification/intensification.py
mlindauer: The question is how to stop intensification?
Maybe raising an exception which is caught in the smbo loop?

mfeurer: That would be great. We could add an `OptimizationAbortException` and catch that in the main loop. This would also make the initial design code easier.

mlindauer: hmm... here is the issue; maybe the incumbent was changed to the first challenger, if the abort  happens because of the second challenger, we would still need to return the changed incumbent.

smac/intensification/intensification.py
```diff
@@ -307,7 +165,284 @@ def intensify(self, challengers, incumbent, run_history, aggregate_func,
...
+            # cost used by incumbent for going over all runs in inst_seed_pairs
+            inc_sum_cost = sum_cost(config=incumbent, instance_seed_pairs=inst_seed_pairs,
+                                    run_history=run_history)
+
+            # Line 12
+            for instance, seed in to_run:
+                # Run challenger on all <config,seed> to run
+                if self.run_obj_time:
+                    # adaptive capping
+                    cutoff = self._adaptive_capping(challenger=challenger, incumbent=incumbent,
```
KEggensperger: The name _adaptive_capping suggests that this method does more than just returning a number, but is doesn't. Naming could be improved for readability.

smac/intensification/intensification.py
KEggensperger: Not sure, whether this is already fixed, but now we are checking twice whether self.run_obj_time is True: Here and in _adaptive_capping in 369. Once should be enough?

smac/intensification/intensification.py
mfeurer: Would we need to? I thought abort is a sign to immediately stop the configuration process in order to make the user fix a bug. Is it required that the data generated during the configuration procedure is usable?

mlindauer: We also use ABORT if our configuration budget is empty and we tried to run a TAE.

mfeurer: Then we'd could add a set of exceptions which are raised when SMAC is supposed to finish, one for time up, one for abort and catch them at the respective places?

mlindauer:
@KEggensperger thanks. both issues fixed.

@mfeurer Could we open another issue and PR later to introduce the exceptions? I would like to merge and close this PR soon.

mfeurer:
Sure, go ahead; but there seems to be a test error now...

mlindauer:
Everything green. Can I merge this now?

mfeurer:
No :p

PR #157, opened 2017-01-20
add landscape conf

KEggensperger:
Adds a landscape configuration file. Checks are now with python3 but also with increased strictness and enforcing full pep8. Thus, the score stays the same.


mfeurer:
Hm, why doesn't landscape check this?

KEggensperger:
Now landscape does run.

PR #158, opened 2017-01-21
Perform local search on configurations from previous runs

jtuyls:
Perform local search on configurations from previous runs as it is described in the SMAC paper.

### DIFF
--- a/smac/smbo/smbo.py
+++ b/smac/smbo/smbo.py
@@ -206,7 +206,7 @@ class SMBO(BaseSolver):
 
         # initiate local search with best configurations from previous runs
         configs_previous_runs = self.runhistory.get_all_configs()
-        configs_previous_runs_sorted = self._get_acq_values_sorted(configs_previous_runs)
+        configs_previous_runs_sorted = self._sort_configs_by_acq_value(configs_previous_runs)
         num_configs_local_search = min(len(configs_previous_runs_sorted), num_configurations_by_local_search)
         next_configs_by_local_search = \
             self._get_next_by_local_search(
@@ -310,7 +310,7 @@ class SMBO(BaseSolver):
 
         return configs_acq
 
-    def _get_acq_values_sorted(self, configs):
+    def _sort_configs_by_acq_value(self, configs):
         """ Sort the given configurations by acquisition value
 
         Parameters
--- a/smac/smbo/smbo.py
+++ b/smac/smbo/smbo.py
@@ -204,13 +204,14 @@ class SMBO(BaseSolver):
             else:
                 num_configurations_by_local_search = 10
 
-        # initial SLS by incumbent +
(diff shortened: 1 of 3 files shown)


smac/smbo/smbo.py
```diff
@@ -204,13 +204,14 @@ def choose_next(self, X, Y,
             else:
                 num_configurations_by_local_search = 10
 
-        # initial SLS by incumbent +
-        # best configuration from next_configs_by_random_search_sorted
+        # initiate local search with best configurations from previous runs
+        configs_previous_runs = self.runhistory.get_all_configs()
+        configs_previous_runs_sorted = self._get_acq_values_sorted(configs_previous_runs)
```
mlindauer: Why not also adding the random configs? Why only the configs from the runhistory?

mfeurer: The idea here is to reproduce the behavior of the original SMAC implementation. I can see your point that it might be good to also start the local search from random points, but we should first get the original implementation up and tested.

smac/smbo/smbo.py
```diff
@@ -204,13 +204,14 @@ def choose_next(self, X, Y,
...
+        num_configs_local_search = min(len(configs_previous_runs_sorted), num_configurations_by_local_search)
         next_configs_by_local_search = \
             self._get_next_by_local_search(
-                [self.incumbent] +
                 list(map(lambda x: x[1],
-                         next_configs_by_random_search_sorted[:num_configurations_by_local_search - 1])))
+                         configs_previous_runs_sorted[:num_configs_local_search])))
```
mfeurer: @mlindauer why did you use `[:num_configurations_by_local_search - 1]`?

smac/smbo/smbo.py
```diff
@@ -308,3 +309,36 @@ def _get_next_by_local_search(self, init_points=typing.List[Configuration]):
         configs_acq.sort(reverse=True, key=lambda x: x[0])
 
         return configs_acq
+
+    def _get_acq_values_sorted(self, configs):
```
mfeurer: I think a more descriptive name of this method would be `_sort_configs_by_acq_value)`.

smac/smbo/smbo.py
mlindauer: I don't know whether I agree. Our goal never was to have a perfect reimplementation.

smac/smbo/smbo.py
mlindauer: Good question. I cannot remember.
If you remove the +1 in Line 203, you maybe can remove the -1 here

jtuyls: Because the incumbent was also added. So the number of configurations for local search was 10, you did local search on the incumbent + 9 random configurations.

Issue #160, opened 2017-01-24
Introduce option to not abort on first crash


mfeurer:
Aborting on the first run will make it hard to run random search with the ROAR algorithm because randomly drawing a configuration which's hyperparameter don't work will make the whole optimization run crash -> solution: disable aborting with a flag in the scenario file/object

mlindauer:
I thought the first configuration in the initial design is the default configuration and it would be weird if this already crashes. 
Furthermore, I would say that your configuration space is not so well designed, if the probability is so high to sample crashing configurations often.

Nevertheless, I do not object such an option.

mfeurer:
Sorry, I meant full random search with a random initial design. I hope it makes more sense then.

mlindauer:
@shukon Please take care of this issue.

PR #162, opened 2017-01-31
Fix/#155

shukon:
Implementing issue#155, adding custom exceptions for TAE Abort- and Crashed-results.

### DIFF
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -156,7 +156,9 @@ class Intensifier(object):
                                                   aggregate_func=aggregate_func)
             except BudgetExhaustedException:
                 # We return incumbent, SMBO stops due to its own budget checks
-                return incumbent
+                inc_runs = run_history.get_runs_for_config(incumbent)
+                inc_perf = aggregate_func(incumbent, run_history, inc_runs)
+                return incumbent, inc_perf
             except TAEAbortException:
                 # We set intensifier.incumbent_on_abort to incumbent to retrieve
                 # it from SMBO and raise
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -316,11 +316,11 @@ class Intensifier(object):
 
                 self.logger.debug("Add run of challenger")
                 status, cost, dur, res = self.tae_runner.start(
-                   config=challenger,
-                   instance=instance,
-                   seed=seed,
-                   cutoff=cutoff,
-                   instance_specific=self.instance_specifics.get(instance, "0"))
+                    config=challenger,
+                    instance=instance,
+                    seed=seed,
+                    cutoff=cutoff,
+                    instance_specific=self.instance_specifics.get(instance, "0"))
                 self._num_run += 1
 
             new_incumbent = self._compare_configs(
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -153,8 +153,7 @@ class Intensifier(object):
                                                   aggregate_func=aggregate_func)
             except BudgetExhaustedException:
                 # We return incumbent, SMBO stops due to its own budget checks
-                inc_runs = run_history.get_runs_for_config(incumbent)
-                inc_perf = aggregate_func(incumbent, run_history, inc_runs)
+                inc_perf = run_history.get_cost(incumbent)
                 return incumbent, inc_perf
 
             if self._chall_indx > 1 and self._num_run > self.run_limit:
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -14,7 +14,7 @@ from smac.utils.constants import MAXINT, MAX_CUTOFF
 from smac.configspace import Configuration
 from smac.runhistory.runhistory import RunHistory
 from smac.tae.execute_ta_run import StatusType
-from smac.tae.execute_ta_run import TAEAbortException, BudgetExhaustedException
+from smac.tae.execute_ta_run import BudgetExhaustedException
 
 __author__ = "Katharina Eggensperger, Marius Lindauer"
 __copyright__ = "Copyright 2017, ML4AAD"
@@ -74,9 +74,6 @@ class Intensifier(object):
         self.maxR = maxR
         self.minR = minR
         self.rs = rng
-
-        # retrieve incumbent in case of interrupted intensification
-        self.incumbent_on_abort = None
 
         # scenario info
         self.cutoff = cutoff
@@ -159,11 +156,6 @@ class Intensifier(object):
                 inc_runs = run_history.get_runs_for_config(incumbent)
                 inc_perf = aggregate_func(incumbent, run_history, inc_runs)
                 return incumbent, inc_perf
-            except TAEAbortException:
-                # We set intensifier.incumbent_on_abort to incumbent to retrieve
-                # it from SMBO and raise
-                self.incumbent_on_abort = incumbent
-                raise
 
             if self._chall_indx > 1 and self._num_run > self.run_limit:
                 self.logger.debug(
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -14,6 +14,7 @@ from smac.utils.constants import MAXINT, MAX_CUTOFF
 from smac.configspace import Configuration
 from smac.runhistory.runhistory import RunHistory
 from smac.tae.execute_ta_run import StatusType
+from smac.tae.execute_ta_run import TAEAbortException, BudgetExhaustedException
 
 __author__ = "Katharina Eggensperger, Marius Lindauer"
 __copyright__ = "Copyright 2017, ML4AAD"
@@ -73,6 +74,9 @@ class Intensifier(object):
         self.maxR = maxR
         self.minR = minR
         self.rs = rng
+
+        # retrieve incumbent in case of interrupted intensification
+        self.incumbent_on_error = None
 
         # scenario info
         self.cutoff = cutoff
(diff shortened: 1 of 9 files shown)


smac/tae/execute_ta_run.py
```diff
@@ -140,6 +148,8 @@ def start(self, config, instance,
 
         self.logger.debug("Return: Status: %d, cost: %f, time. %f, additional: %s" % (
             status.value, cost, runtime, str(additional_info)))
+        print("Return: Status: %d, cost: %f, time. %f, additional: %s" % (
```
mlindauer: Why have you added a print here?

smac/smbo/smbo.py
```diff
@@ -126,12 +127,16 @@ def run(self):
...
-                run_history=self.runhistory,
-                aggregate_func=self.aggregate_func,
-                time_bound=max(0.01, time_spend))
+            try: self.incumbent, inc_perf = self.intensifier.intensify(
+                  challengers=challengers,
+                  incumbent=self.incumbent,
+                  run_history=self.runhistory,
+                  aggregate_func=self.aggregate_func,
+                  time_bound=max(0.01, time_spend))
+            except (TAEAbortException, BudgetExhaustedException):
```
mlindauer: `intensify` races more than one challenger against the incumbent
and we always want to return the latest incumbent.
Here a potential problem:
If the first challenger beats the incumbent, and on the second challenge we would get a `BudgetExhaustedException`,
we would not return the newest incumbent (i.e., the first challenger) with your implementation.

Either these exceptions will be handled within intensify(),
or at least Intensifier stores the current incumbent always in an object variable,
such that we can return `self.intensifier.incumbent`

shukon: Is there any disadvantage of storing the incumbent in `self.intensifier.incumbent`? If handling the exceptions in intensification, it is more complicated to abort the smbo-loop.

smac/tae/execute_ta_run.py
shukon: Unintentional, removed.

smac/smbo/smbo.py
mlindauer: > Is there any disadvantage of storing the incumbent in self.intensifier.incumbent?

not as far as I know

shukon: I chose the minimal-invasive solution and introduced an intensifier.incumbent_on_error that is set through exception-handling. Also added another try/except around the initial-run in smbo.

shukon: Actually I suppose the `BudgetExhaustedException` does not need to be handed up to smbo, since there already is a check for budget there. Also changed the handling in intensify to be more intuitive.

mlindauer:
@shukon please ensure that the coverage of our unit tests do not decrease!

smac/smbo/smbo.py
```diff
@@ -102,7 +103,11 @@ def run(self):
             The best found configuration
         '''
         self.stats.start_timing()
-        self.incumbent = self.initial_design.run()
+        try:
+            self.incumbent = self.initial_design.run()
+        except TAEAbortException:
```
mfeurer: This exception changes the current behaviour of SMAC as it does not indicate a failure if the first run aims to abort the evaluation. Thus, I think this try/except is not necessary at this point and should be only added in a PR to fix #160 (or this PR should also fix #160).

smac/tae/execute_ta_run.py
```diff
@@ -124,9 +129,12 @@ def start(self, config, instance,
                                                           seed=seed,
                                                           instance_specific=instance_specific)
 
-        if self.stats.ta_runs == 0 and status in [StatusType.CRASHED, StatusType.ABORT]:
+        if self.stats.ta_runs == 0 and status == StatusType.CRASHED:
             self.logger.critical("First run crashed -- Abort")
```
mfeurer: This should now rather be the message of the exception than a call to the logger.

smac/tae/execute_ta_run.py
```diff
@@ -124,9 +129,12 @@ def start(self, config, instance,
...
-            sys.exit(1)
+            raise TAEAbortException()
+        if status == StatusType.ABORT:
+            self.logger.error("Target algorithm status ABORT - SMAC will exit.")
```
mfeurer: Also, I think this should be the exception message and not a call to the logger.

smac/smbo/smbo.py
```diff
@@ -126,12 +131,15 @@ def run(self):
...
-                run_history=self.runhistory,
-                aggregate_func=self.aggregate_func,
-                time_bound=max(0.01, time_spend))
+            try: self.incumbent, inc_perf = self.intensifier.intensify(
+                  challengers=challengers,
+                  incumbent=self.incumbent,
+                  run_history=self.runhistory,
+                  aggregate_func=self.aggregate_func,
+                  time_bound=max(0.01, time_spend))
+            except TAEAbortException:
```
mfeurer: This except prevents someone who uses SMAC from python from figuring out that something went wrong in the code. Do we want this?

smac/tae/execute_ta_run.py
```diff
@@ -113,7 +118,7 @@ def start(self, config, instance,
         if self.stats.is_budget_exhausted():
             self.logger.debug(
                 "Skip target algorithm run due to exhausted configuration budget")
```
mfeurer: I think this should be the message of the exception rather than a call to the logger.

smac/smbo/smbo.py
shukon: What would be the alternative? Issuing a failed SMAC-run/ raising an exception? I used 
>  The SMAC object should always return an incumbent. (Issue/#155)

as a guideline.

smac/smbo/smbo.py
shukon: This try/except-statement is simply aiming to ensure that an incumbent is always returned. The options to not abort on first-run-crash would rather fit into the tae-runner, I think.
Related to below: What should the expected failure on ABORTs look like?

shukon:
Not sure what is going on with the builds, sorry. Any way to have a look in the logs?

mfeurer:
Click on the link details, then you get to travis-ci. There you see all four/six builds that are happening. When you click one of those, you get the details.

shukon:
Ok, thanks. Also: If first run crashes/ aborts in the initial-design-run, then an `TAEAbortException` is now raised. Not sure if this is expected behaviour, in comparison to ABORTs in the later course of the optimization, where we will not raise, but return incumbent. Need some guidelines on that.

mlindauer:
Let me discuss this with Matthias in the course of the week.

mlindauer:
So, we came to the following conclusion:
the TAEAbortException should only be caught in `smac_cli` such that the command line user gets a message that the TA signaled an abort and the last incumbent can be found in the trajectory file.
This also implies that we do not catch the exception in `SMBO`

shukon:
Should be working now. Message to user by `self.logger.error(message)` - is that fine?

smac/intensification/intensification.py
```diff
@@ -141,14 +142,20 @@ def intensify(self, challengers: typing.List[Configuration],
...
+
+                # Lines 8-17
+                incumbent = self._race_challenger(challenger=challenger,
+                                                  incumbent=incumbent,
+                                                  run_history=run_history,
+                                                  aggregate_func=aggregate_func)
+            except BudgetExhaustedException:
+                # We return incumbent, SMBO stops due to its own budget checks
+                inc_runs = run_history.get_runs_for_config(incumbent)
+                inc_perf = aggregate_func(incumbent, run_history, inc_runs)
```
mlindauer: you can use the function `get_cost()` instead of these two lines

smac/tae/execute_ta_run.py
```diff
@@ -35,6 +35,18 @@ def enum_hook(obj):
                 return getattr(globals()[name], member)
         return obj
 
+class BudgetExhaustedException(Exception):
+    """ Exception indicating that time- or memory-budgets are exhausted. """
+    pass
+
+class TAEAbortException(Exception):
+    """ Exception indicating that the TAE suggests an ABORT of SMAC. """
```
mfeurer: @mlindauer is this correct? It can be the TAE (in case of the first run crashing, but there's a special exception for that), but mostly this is supposed to be sent by the target algorithm, right? Also, I'm not sure if the user should be confronted to the abbreviation TAE.

smac/tae/execute_ta_run.py
```diff
@@ -35,6 +35,18 @@ def enum_hook(obj):
...
+    pass
+
+class TAEFirstRunCrashedException(TAEAbortException):
```
mfeurer: Hm, could we rename this to `FirstRunCrashedException`? I think this is easier for a user.

smac/smbo/smbo.py
```diff
@@ -98,7 +100,15 @@ def run(self):
...
-        self.incumbent = self.initial_design.run()
+        try:
+            self.incumbent = self.initial_design.run()
+        except TAEFirstRunCrashedException:
+            if self.scenario.abort_on_first_run_crash:
+                # We log to make sure the user knows about the reason for abort
+                self.logger.error("First run crashed, abort. (To prevent this,"
+                                  "toggle the 'abort_on_first_run_crash'-optio"
+                                  "n!)")
+                raise TAEAbortException("First run crashed, abort.")
```
mfeurer: Why not raising the original exception?

smac/smac_cli.py
```diff
@@ -59,29 +60,32 @@ def main_cli(self):
...
-                rng=np.random.RandomState(args_.seed), 
-                runhistory=rh, 
+                scenario=scen,
+                rng=np.random.RandomState(args_.seed),
+                runhistory=rh,
                 initial_configurations=initial_configs)
         try:
             optimizer.optimize()
-
+        except TAEAbortException:
```
mfeurer: Does this also catch the exception that the first run crashed? If yes, the error message would be incorrect (as the tae and not the ta signaled to abort).

smac/tae/execute_ta_run.py
shukon: Corrected the doc-string. Would `AbortException` be enough? Or maybe `TargetAlgorithmAbortException`?

smac/tae/execute_ta_run.py
shukon: Done.

smac/smbo/smbo.py
shukon: Now raising FirstRunCrashedException, logging in smac_cli.

smac/smac_cli.py
shukon: Now logging exception-message (which is either the abort through TA or First Run Crashed)

shukon:
Implemented suggestions. I kept the name `TAEAbortException` for now, would `TargetAlgorithmAbortException` be better? Generally the exception-message should be what concerns the user, shouldn't it?

PR #163, opened 2017-02-04
Add log-function to scenario (issue/#161)

shukon:
Adds a `Scenario._write()`, that is executed automatically upon creating a scenario and writes a scenario.txt to the output_dir.
Note that this only supports objects, that can be read in from a text-file anyway, no `ConfigurationSpace`-objects and also no lists of instances. This is, so that the input_reader can be used for loading it again, as requested in issue/#161.
A new file with a list of available options in the scenario.txt is introduced in smac/utils/scenario_options.py.


smac/scenario/scenario.py
```diff
@@ -345,6 +359,40 @@ def extract_instance_specific(instance_list):
...
+            ''' returns the correct value to write into a file '''
+            if key == 'ta' : return " ".join(value)
+            else: return value
+
+        # Create output-dir if necessary
+        if not os.path.isdir(self.output_dir):
+            self.logger.debug("Output directory does not exist! Will be created.")
+            try:
+                os.makedirs(self.output_dir)
+            except OSError:
```
mlindauer: don't catch the exception but it can be raised to the user with an appropriated message.

shukon: So there still is a try/except-clause now, but only to re-raise with appropriate error-msg.

shukon:
The scenario.write() now writes the pcs-file to the output-dir. Is it possible that we do not have a pcs-file-path (i.e. that the ConfigSpace is defined directly in python?)?

smac/scenario/scenario.py
```diff
@@ -345,6 +358,57 @@ def extract_instance_specific(instance_list):
...
 
+    def _write(self):
+        """Write scenario to file in a format that can be easily retrieved using
+        the input_reader. Will overwrite if file already exists.
+
+        Sideeffect: creates output-directory if it doesn't exist.
+        """
+        self.logger.debug("Writing scenario-file to {}.".format(self.output_dir))
+        # Reverse mapping of option-names to internal -> external
+        options_int2ext = {v: k for k, v in self.options_ext2int.items()}
```
mfeurer: Does a scenario file written for logging purposes has to be backwards compatible?

smac/scenario/scenario.py
```diff
@@ -345,6 +358,57 @@ def extract_instance_specific(instance_list):
...
+            ''' returns the correct value to write into a file '''
+            if key == 'pcs_fn':
+                # Write PCS to output_dir and change filepath
+                if value != None:
+                    old_path = os.path.splitext(os.path.split(value)[1])[0]
+                else:
+                    old_path = "configspace"
+                new_path = os.path.join(self.output_dir, old_path+".pcs")
+                with open(new_path, "w") as f:
+                    print(value)
```
mfeurer: Hm, what's that print doing here?

smac/scenario/scenario.py
mfeurer: This would fail if the output directory already exists. Is this intended behaviours?

shukon: I took inspiration from traj_logger: https://github.com/automl/SMAC3/blob/master/smac/utils/io/traj_logging.py#L46-L52

shukon:
I agree with the problems of the design. I will attempt to explain the new solution:
The scenario_options.py has been removed. I was confused by the "name" and "dest" -fields of the `scenario._arguments`, but I have now adapted this in the `_write`-function. We are processing all options to write them to a scenario-file, as we would by hand. This way we do not need any extra reloading-function. I still feel that "name" and "dest" instead of one all-purpose name per option is unnecessarily complicated.
The `ConfigSpace` is now written to the `output_dir` and the `pcs_fn`-path is changed accordingly in the logged scenario. I would appreciate a new review, I also believe that the implementation is much simpler and more intuitive now.

PR #170, opened 2017-02-11
Speed up conditional hyperparameter imputation

mfeurer:
Instead of using a method for this, convert the configuration
into an array and replace missing values with -1.

### DIFF
--- a/smac/smbo/local_search.py
+++ b/smac/smbo/local_search.py
@@ -71,9 +71,9 @@ class LocalSearch(object):
         """
         incumbent = start_point
         # Compute the acquisition value of the incumbent
-        incumbent_ = incumbent.get_array()
-        incumbent_[~np.isfinite(incumbent_)] = -1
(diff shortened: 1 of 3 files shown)


mlindauer:
Let's first check whether Frank agrees or has some concerns.

smac/smbo/local_search.py
```diff
@@ -71,10 +71,9 @@ def maximize(self, start_point, *args):
         """
         incumbent = start_point
         # Compute the acquisition value of the incumbent
-        incumbent_ = impute_inactive_values(incumbent)
-        acq_val_incumbent = self.acquisition_function(
-                                            incumbent_.get_array(),
-                                            *args)
+        incumbent_ = incumbent.get_array()
```
KEggensperger: Distinguishing between two variables with only "_" can cause a lot of confusion or needs at least a comment in the code.

PR #174, opened 2017-02-13
Maint/faster one exchange neighborhood

mfeurer:


smac/smbo/smbo.py
```diff
@@ -307,13 +303,10 @@ def _sort_configs_by_acq_value(self, configs):
...
-        imputed_configs = map(ConfigSpace.util.impute_inactive_values,
-                              configs)
-        imputed_configs = [x.get_array()
-                           for x in imputed_configs]
-        imputed_configs = np.array(imputed_configs,
-                                   dtype=np.float64)
-        acq_values = self.acquisition_func(imputed_configs)
+        config_array = np.array([x.get_array() for x in configs], dtype=np.float64)
+        # This imputes inactive values!
+        config_array[~np.isfinite(config_array)] = -1
```
mlindauer: shouldn't this imputation be in a method such that we cannot forget to do it everywhere?

mfeurer: This was actually introduced in PR #170 and I'm not sure why github does not update this PR. I can update this though (with yet a new pull request).

mfeurer: Now the PR is updated ;)

PR #176, opened 2017-02-13
MAINT refactor imputation of inactive hyperparameters into function

mfeurer:
Refactor imputation into it's own function @mlindauer

### DIFF
--- a/smac/configspace/__init__.py
+++ b/smac/configspace/__init__.py
@@ -1,3 +1,4 @@
 from ConfigSpace import ConfigurationSpace, Configuration
 from ConfigSpace.io import pcs
-from ConfigSpace.util import impute_inactive_values, get_random_neighbor, get_one_exchange_neighbourhood
+from ConfigSpace.util import get_random_neighbor, get_one_exchange_neighbourhood
+from smac.configspace.util import impute_inactive_hyperparameters
--- a/smac/configspace/__init__.py
+++ b/smac/configspace/__init__.py
@@ -1,4 +1,4 @@
 from ConfigSpace import ConfigurationSpace, Configuration
 from ConfigSpace.io import pcs
 from ConfigSpace.util import get_random_neighbor, get_one_exchange_neighbourhood
-from smac.configspace.util import impute_inactive_hyperparameters
+from smac.configspace.util import convert_configurations_to_array
--- a/smac/configspace/util.py
+++ b/smac/configspace/util.py
@@ -6,6 +6,20 @@ from smac.configspace import Configuration
 
 
 def impute_inactive_hyperparameters(configs: List[Configuration]) -> np.ndarray:
+    """Impute inactive hyperparameters in configurations with their default.
+
+    Necessary to apply an EPM to the data.
+
+    Parameters
+    ----------
+    configs : List[Configuration]
+        List of configuration objects.
+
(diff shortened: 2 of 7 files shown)


smac/configspace/__init__.py
```diff
@@ -1,3 +1,4 @@
 from ConfigSpace import ConfigurationSpace, Configuration
 from ConfigSpace.io import pcs
-from ConfigSpace.util import impute_inactive_values, get_random_neighbor, get_one_exchange_neighbourhood
\ No newline at end of file
+from ConfigSpace.util import get_random_neighbor, get_one_exchange_neighbourhood
+from smac.configspace.util import impute_inactive_hyperparameters
```
mlindauer: I haven't found such a file. maybe still missing?

mlindauer:
btw: have you checked that the random forest can handle -1 for categorical parameters?

mfeurer:
@sfalkner could you please comment on the question from Marius? It boils down to: does the random forest code from commit [21157707735bc8a11ebf3790a8f5d8de2fcab860](https://github.com/sfalkner/random_forest_run/tree/21157707735bc8a11ebf3790a8f5d8de2fcab860) (basically version 0.2 from last April) support -1 as a value for categorical and numerical values?

sfalkner:
No it doesn't. What would be the purpose of that value anyway?

mfeurer:
An outlier value representing a missing value. I'll update this to impute the default, but still be fast.

sfalkner:
You could use 0 for undefined and use 1...num_categories for the actual values...Don't know if that helps, though. But I guess you wanted the same value for numerical and categorical variables.

mfeurer:
Now it's the default again (as it was before). When can decide when the deadlines are over.

smac/runhistory/runhistory2epm.py
```diff
@@ -296,12 +296,12 @@ def _build_matrix(self, run_dict, runhistory, instances=None, par_factor=1):
         for row, (key, run) in enumerate(run_dict.items()):
             # Scaling is automatically done in configSpace
             conf = runhistory.ids_config[key.config_id]
-            conf = impute_inactive_values(conf)
+            conf_array = impute_inactive_hyperparameters([conf])[0]
```
mlindauer: this is an vector and not an array, right?

smac/configspace/util.py
```diff
@@ -0,0 +1,23 @@
+from typing import List
+
+import numpy as np
+
+from smac.configspace import Configuration
+
+
+def impute_inactive_hyperparameters(configs: List[Configuration]) -> np.ndarray:
```
mlindauer: please add docstring

mlindauer: the name of the method is a bit misleading, since it does more than imputation; it also converts it into our internal numpy representation.

smac/runhistory/runhistory2epm.py
mfeurer: Which is the same class in numpy, but I get your point. Will change this.

smac/configspace/util.py
mfeurer: Docstring is there, how about `convert_configurations_to_array`?

mlindauer:
Thanks!

PR #180, opened 2017-02-14
Issue/#159 - consistent names for loggers

shukon:
Changed logger-names and removed most trailing white-spaces while at it.

Coverage decreased because a line duplicate is removed in runhistory2epm... not sure what to do about it.

### DIFF
--- a/smac/epm/random_epm.py
+++ b/smac/epm/random_epm.py
@@ -22,7 +22,7 @@ class RandomEPM(AbstractEPM):
         ----------
         rng : np.random.RandomState
         '''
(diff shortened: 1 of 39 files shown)


mfeurer:
You don't need to worry about minimal coverage reduction. And I agree with Marius' suggestion for class naming.

shukon:
I can do that, but logger-names will not be as nice for the eyes, e.g. `smac.scenario.scenario.Scenario` or `test_scenario.test_scenario.ScenarioTest`. Is that good or should I crop the last part of the module-name?

PR #181, opened 2017-02-17
MAINT make trajectory entries named tuple (issue/#153)

shukon:
Trajectory entries are named tuples, allowing access by name, e.g.: `tl.trajectory[n].train_perf'`


shukon:
Added `stats.ta_runs`, not sure about the formats - saved in aclib-format as "algo_runs"...

mlindauer:
Sorry, there was a misunderstanding.
With the "internal trajectory", I meant only `self.trajectory`.
The output files have a specified format which we cannot change easily.
Please remove the `Algorithm Runs` field from the output files.

shukon:
That makes sense. Done.

PR #183, opened 2017-02-18
ADD method get_tae_runner to SMAC-class (Issue/#128)

shukon:
Adding method `get_tae_runner` to `SMAC`.


shukon:
I found only the rf-example using this method? If so, done.

PR #189, opened 2017-03-01
Issue/#161 log scenario

shukon:
Adds a logging-function to the Scenario-class.
I was confused by the "name" and "dest" -fields of the `scenario._arguments`, but I have now adapted this in the `_write`-function. We are processing all options to write them to a scenario-file, as we would by hand. This way we do not need any extra reloading-function. I still feel that "name" and "dest" instead of one all-purpose name per option is unnecessarily complicated.
The `ConfigSpace` is also written to the `output_dir` and the `pcs_fn`-path is changed accordingly in the logged scenario.


shukon:
So now configspace, train-instances, test-instances and features are logged in the output-directory, regardless of whether they are specified in a file or simply handed over as a dict. The write-function did get a bit longer and if-elif-heavy, but is aiming to be systematical and easily maintainable.

mlindauer:
I fixed two little issues.
Unfortunately, there are some merge conflicts.
@shukon could you please resolve them.

Furthermore, the reading and writing is right now implemented inconsistently. For reading, we have a class (`InputReader`) and for writing, we have a method inside of `Scenario`. Actually, I prefer the former way: encapsulating the functionality into a separate class. @shukon: could you please create a new class and move the write method into this class. This also allows for moving the `extract_instance_specific` outside of the `write` function---I always try to prevent nesting of methods since it gets confusing.

mlindauer:
@shukon please also fix the unit test

shukon:
@mlindauer resolved merge-conflicts, fixed unit-tests and encapsulated scenario-logging. Added an `OutputWriter` analogous to `InputReader`. The write has no nested functions anymore - although it was the `transform_value` (now `_parse_argument`), that was nested, not `extract_instance_specific `. The latter is still nested, but in `scenario`'s `_transform_arguments`. Maybe that should be it's own issue, since it has nothing to do with writing.

mlindauer:
@shukon Thank you for the changes.
Could you please use our doc string format for all methods.

> it was the `transform_value` (now _parse_argument), that was nested, not `extract_instance_specific`

sorry, my mistake.

shukon:
Fixed docstrings (not sure about sideffects).

PR #194, opened 2017-03-17
Fix/adaptive cap

mlindauer:
FIX adaptive capping with re-occurring configurations

* ADD new StatusType.CAPPED
* ADD new exception (CappedRunException) to prevent infinite loop in intensification
* FIX CAPPED runs are added to runhistory in data but not in the remaining
fast data structures as used in intensification
* MAINT improve corresponding unit test

### DIFF
--- a/smac/facade/smac_facade.py
+++ b/smac/facade/smac_facade.py
@@ -240,7 +240,7 @@ class SMAC(object):
                     scenario=scenario, num_params=num_params,
                     success_states=[StatusType.SUCCESS, ],
                     impute_censored_data=True,
-                    impute_state=[StatusType.TIMEOUT, ],
+                    impute_state=[StatusType.TIMEOUT, StatusType.CAPPED ],
                     imputor=imputor)
 
             elif scenario.run_obj == 'quality':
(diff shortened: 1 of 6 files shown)


smac/facade/smac_facade.py
```diff
@@ -240,7 +240,7 @@ def __init__(self,
                     scenario=scenario, num_params=num_params,
                     success_states=[StatusType.SUCCESS, ],
                     impute_censored_data=True,
-                    impute_state=[StatusType.TIMEOUT, ],
+                    impute_state=[StatusType.TIMEOUT, StatusType.CAPPED ],
```
KEggensperger: I think TIMEOUT is no longer necessary in this list as we only impute TIMEOUT with runtime < cutoff. These runs are now of status CAPPED.

smac/runhistory/runhistory2epm.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, scenario, num_params,
             self.rs = np.random.RandomState()
 
         if self.impute_state is None:
-            self.impute_state = [StatusType.TIMEOUT, ]
+            self.impute_state = [StatusType.TIMEOUT, StatusType.CAPPED]
```
KEggensperger: also here

PR #197, opened 2017-03-21
ADD intensification_percentage option

shukon:
Adding option intensification_percentage (analogous to SMAC2), must be in (0,1), defines the fraction of time assigned to intensification. Uses the `time_bound`-parameter in `intensifier.intensify`.

### DIFF
--- a/smac/smbo/smbo.py
+++ b/smac/smbo/smbo.py
@@ -123,8 +123,18 @@ class SMBO(BaseSolver):
             challengers = self.choose_next(X, Y)
 
             time_spend = time.time() - start_time
-            logging.debug(
-                "Time spend to choose next configurations: %.2f sec" % (time_spend))
+
+            # Calculate time left for intensify
+            frac_intensify = self.scenario.intensification_percentage
+            if (frac_intensify <= 0 or frac_intensify >= 1):
+                raise ValueError("The value for intensification_percentage-"
+                                 "option must lie in (0,1), instead: %.2f" % (frac_intensify))
+            total_time = time_spend / (1-frac_intensify)
+            time_left = frac_intensify * total_time
+            self.logger.debug("Total time: %.4f, time spend on choosing next "
+                              "configurations: %.4f (%.2f), time left for "
+                              "intensification: %.4f (%.2f)" % (total_time,
+                    time_spend, (1-frac_intensify), time_left, frac_intensify))
 
             self.logger.debug("Intensify")
 
@@ -133,7 +143,7 @@ class SMBO(BaseSolver):
                 incumbent=self.incumbent,
                 run_history=self.runhistory,
                 aggregate_func=self.aggregate_func,
-                time_bound=max(0.01, time_spend))
+                time_bound=max(0.01, time_left))
 
             if self.scenario.shared_model:
                 pSMAC.write(run_history=self.runhistory,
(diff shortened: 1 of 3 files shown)


test/test_smbo/test_smbo.py
```diff
@@ -299,9 +299,25 @@ def target(x):
...
+        scen = Scenario({'cs': test_helpers.get_branin_config_space(),
+                         'run_obj': 'quality',
+                         'intensification_percentage' : 1})
+        smbo = SMAC(scen, tae_runner=target, rng=1).solver
+        self.assertRaises(ValueError, smbo.run)
+
     def tearDown(self):
-            for d in glob.glob('smac3-output*'):
-                shutil.rmtree(d)
+        for d in glob.glob('smac3-output*'):
```
mlindauer: you can set the output_dir to "" to avoid the output directory

test/test_smbo/test_smbo.py
```diff
@@ -299,9 +299,25 @@ def target(x):
         smbo = SMAC(scen, tae_runner=target, rng=1).solver
         self.assertRaises(FirstRunCrashedException, smbo.run)
 
+    def test_intensification_percentage_bad_value(self):
```
mlindauer: could you please also a test for a value in the valid range

test/test_smbo/test_smbo.py
shukon: Actually only changed indention - but I agree that it does not really make sense to remove all smac-output*-directories, now that I look at it.

shukon:
Adds positive test for option (via log-capture, since we kind of have to test local variables) and fix the tearDown.

mlindauer:
@mfeurer since you are back, do you also want to review this PR? or can I merge it?

test/test_smbo/test_smbo.py
```diff
@@ -294,14 +296,56 @@ def target(x):
...
+        self.assertRaises(ValueError, smbo.run)
+        smbo = get_smbo(-0.2)
+        self.assertRaises(ValueError, smbo.run)
+        # Test for >= 1
+        smbo = get_smbo(1)
+        self.assertRaises(ValueError, smbo.run)
+        smbo = get_smbo(1.2)
+        self.assertRaises(ValueError, smbo.run)
+
     def tearDown(self):
```
mfeurer: This could be deleted now.

smac/smbo/smbo.py
```diff
@@ -122,9 +122,19 @@ def run(self):
             # get all found configurations sorted according to acq
             challengers = self.choose_next(X, Y)
 
-            time_spend = time.time() - start_time
-            logging.debug(
-                "Time spend to choose next configurations: %.2f sec" % (time_spend))
+            time_spent = time.time() - start_time
```
mfeurer: I think the functionality should be moved into it's own method. This will keep the SMBO loop simple and easy to read. Also, it will allow for a simple unit test of the functionality in which you don't have to mess around with the logger.

test/test_smbo/test_smbo.py
shukon: Yes.

smac/smbo/smbo.py
shukon: Agreed. Makes everything easier.

PR #200, opened 2017-03-27
ADD convenient function to get X,y of RH

mlindauer:
@sfalkner requested an interface to get all X,y from SMAC

* ADD X,y interface
* change return values of `fmin_smac` to return also the smac object which allows to use other convenience function

### DIFF
--- a/smac/facade/smac_facade.py
+++ b/smac/facade/smac_facade.py
@@ -384,27 +384,18 @@ class SMAC(object):
 
     def get_X_y(self):
         '''
-            returns all data in runhistory
-            in X, y format, 
+            simple interface to obtain all data in runhistory
+            in X, y format 
+            
+            Uses smac.runhistory.runhistory2epm.AbstractRunHistory2EPM.get_X_y()
 
             Returns
             ------- 
-            X: matrix of all configurations (+ instance features)
-            y vector of cost values; can include censored runs
-            cen: list of bools indicating whether the y-value is censored
+            X: numpy.ndarray
+                matrix of all configurations (+ instance features)
+            y numpy.ndarray
+                vector of cost values; can include censored runs
+            cen: numpy.ndarray
+                vector of bools indicating whether the y-value is censored
         '''
-        X = []
-        y = []
-        cen = []
-        feature_dict = self.solver.scenario.feature_dict
-        params = self.solver.scenario.cs.get_hyperparameters()
-        for k, v in self.runhistory.data.items():
-            config = self.runhistory.ids_config[k.config_id]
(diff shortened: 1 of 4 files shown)


smac/facade/smac_facade.py
```diff
@@ -340,16 +340,71 @@ def optimize(self):
...
+            'train_perf', 'incumbent_id', 'incumbent',
+            'ta_runs', 'ta_time_used', 'wallclock_time'
+        '''
+
         if not hasattr(self, 'trajectory'):
             raise ValueError('SMAC was not fitted yet. Call optimize() prior '
                              'to accessing the runhistory.')
         return self.trajectory
+
+    def get_X_y(self):
```
mfeurer: This code looks a lot like a copy from `RunHistory2EPM`. Maybe that code can be reused?

mlindauer: In the beginning, I thought so, too.
However, the rh2epm methods do a lot more at the moment than we would like to do here:
- it imputes censored data -- we don't want this here, but a vector indicating whether some y's are censored
- it uses a log-transformation for runtime scenarios -- I would need to undo it again
- it uses the unit-hypercube parameter representation -- I would also need to undo it, since Stefan preferred the original values

Since, the undoing of all these things needs at least as many code lines as this function, I chose to write it directly without reusing rh2epm.

smac/facade/smac_facade.py
```diff
@@ -340,16 +340,71 @@ def optimize(self):
...
+            cen: list of bools indicating whether the y-value is censored
+        '''
+        X = []
+        y = []
+        cen = []
+        feature_dict = self.solver.scenario.feature_dict
+        params = self.solver.scenario.cs.get_hyperparameters()
+        for k, v in self.runhistory.data.items():
+            config = self.runhistory.ids_config[k.config_id]
+            x = [config[p.name] for p in params]
```
mfeurer: This isn't sorted, so you'd have to iterate over the hyperparameters of the configuration space.

smac/facade/smac_facade.py
mfeurer: Okay.

mfeurer:
A unittest would be nice, though.

mlindauer:
@mfeurer I added a unit test -- I wanted to wait for the approval of the interface before writing it.

@sfalkner : What do you think about the interface?

mlindauer:
If SMAC optimizes quality (/cost/loss) and the target algorithm also returns its running time (which is recommended), we could also return the running time. However, I don't know whether this is really interesting. As long as nobody asks for it, I would like to not support it.

PR #206, opened 2017-03-29
MAINT instances are casted to str

mlindauer:
as discussed today


mlindauer:
@mfeurer a consequence of the change is that scenarios without instances use now `"None"` as a single instance and not `None`. Somehow, I don't like it, but I don't know a real solution for it.

Issue #204, opened 2017-03-29
Always save runhistory?


mfeurer:
At the moment SMAC3 saves the runhistory into its output directory if the shared model mode is activated. In case a user wants to save the runhistory, but his tool (in this case auto-sklearn) can work in both the parallel and the sequential setting, the user has to check whether parallel model is on or not and depending on that write the necessary information to disk. As SMAC2 automatically saved a runs and results file, I wonder whether we should save the runhistory by default as well.

mlindauer:
just my 2 cents: the command line interface of SMAC3 already writes the runhistory to disk by default. The `SMAC` facade doesn't do it because the idea was if the user lives inside the python world, he/she does not want to have unnecessary large files on disk and the runhistory can be a very large file.

PR #208, opened 2017-03-30
ADD #204, always save the runhistory

mfeurer:
Adds #204, always saves the runhistory at the end of an optimization run instead of only saving it when called from command line. As a side-effect, the filename of the runhistory as saved for pSMAC no longer has the SMAC seed in it's file name. This should anyway be no longer needed since @mlindauer implemented pSMAC loading data from different directories.

### DIFF
--- a/smac/facade/smac_facade.py
+++ b/smac/facade/smac_facade.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import typing
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


smac/facade/smac_facade.py
```diff
@@ -337,6 +338,11 @@ def optimize(self):
             self.logger.info("Final Incumbent: %s" % (self.solver.incumbent))
             self.runhistory = self.solver.runhistory
             self.trajectory = self.solver.intensifier.traj_logger.trajectory
+
+            if self.solver.scenario.output_dir is not None:
```
mlindauer: why not in a `finally ` as I did in the cli?
If we raised a  TAEAbortException (or something else crashed), we maybe still want to have the data.

Issue #209, opened 2017-03-30
Add seed to output directory name


mfeurer:
In order to have an overview of what has been run one would sometimes like to know the seed of previous runs. One solution would be to store it in the output directory name, but this isn't as easy as it sounds as the scenario object does not know the seed for the SMAC object.

KEggensperger:
A remark: We don't use a seed but a rng.

Nevertheless, for the CLI there exists a seed and adding that to the folder name still makes sense.

KEggensperger:
Sorry, but I think this is not yet fully fixed:

```from smac.facade.smac_facade import SMAC
from smac.scenario.scenario import Scenario
from smac.configspace import ConfigurationSpace
from ConfigSpace.hyperparameters import UniformIntegerHyperparameter
import numpy as np

def dummy(cfg, seed):
    return 1

cs = ConfigurationSpace()
a = UniformIntegerHyperparameter("a", 0, 10, default=5)
cs.add_hyperparameter(a)

scenario = Scenario({
        "run_obj": "quality",
        "runcount-limit": 100,
        "cs": cs,
        "output_dir": "./a"})

smac = SMAC(scenario=scenario, tae_runner=dummy, rng=2)
smac = SMAC(scenario=scenario, tae_runner=dummy, rng=np.random.RandomState(2))
```

I would expect that at least one of these cmds writes to a directory "a/run2", but both write to "a_run1". Apparently, the seed/rng does not affect the name of the ```output_dir```

mlindauer:
no, you have to pass the `run_id` to `Scenario`. The `scenario` object is right now responsible for creating the output directory and in `SMAC` it is already to late to use the `rng`. The alternative would be that the optimizer (here `SMAC`) has to create the `output_directory`; however this would imply that every optimizer is responsible for the same task.

btw: why have you closed the issue again?

KEggensperger:
I clicked the wrong button ..

Yes, that's what @mfeurer already pointed out. What about adding the seed to the Scenario instead of the optimizer?

mlindauer:
Our seed handling is already a bit strange (in some classes). I would say that seeds should be passed on a function-level and not on a class-level. I don't want to recreate all objects again only because I want to run the same stuff with a different seed.
The same applies for the scenario object. To run SMAC again with a different seed, I would not like to recreate the scenario.

mfeurer:
@mlindauer @KEggensperger is this fixed by #240 or #334 ?

[...]

PR #217, opened 2017-04-07
Output directory name for scenario file

jtuyls:
Add milliseconds to output directory name for scenario to avoid crash (OSError) when starting multiple smac instances in parallel (did happen on cluster)

### DIFF
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -255,7 +255,7 @@ class Scenario(object):
                           default="smac3-output_%s" % (
                               datetime.datetime.fromtimestamp(
                                   time.time()).strftime(
-                                  '%Y-%m-%d_%H:%M:%S')))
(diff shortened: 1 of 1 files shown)


KEggensperger:
Minor remark: Fixing #209 would also help here plus we avoid having long folder names. Right?

jtuyls:
@KEggensperger  Yes, if there is always a seed that can be retrieved, that should also work most of time. Except when seeds are the same by chance, but I think that is a rather small chance.

mfeurer:
I think using seeds or milliseconds serve a different purpose. Milliseconds can only be used to distinguish the runs, while seeds can be used to actually know what runs were started. Good points that we don't always have a seed. Not sure what to do about that.

KEggensperger:
Yes, that's exactly what I added as a comment to that issue. I just wanted to mention, that a seed would be more meaningful and solve the problem as well. 

BTW: I do not vote against merging this PR as a solution for now.

PR #218, opened 2017-04-07
Wrapper around acquisition function

jtuyls:
Wrapper class around the acquisition function to improve extensibility of the acquisition function. In this class preliminary computations can be done on the configurations before they are transformed to their array representation. (For me this came in handy when I extended SMAC to compute reductions in runtime for configurations that are cached before computing the acquisition function value. Otherwise I had to add these computations to both the _sort_configs_by_acq_value method in the SMBO class and the maximize method in local search class.)  It might be interesting to add this for extensibility in the future, what do you think?


KEggensperger:
Probably I miss something: Why can't you subclass AcquisitionFunction, do transformations, and then call the supermethod?

jtuyls:
@KEggensperger Because this class works with vector representations of the configurations and not the Configuration objects.

KEggensperger:
@jtuyls Sorry, I missed that. You are right. One last question: Would it be possible to modify the acquisition function to accept configs instead of vectors?

jtuyls:
@KEggensperger Modifying the acquisition function to accept configs instead of vectors is also possible. I did not do this because I thought it was more elegant in my case to separate the two and have clear responsibilities (I did not substantially alter the acquisition function but calculated it for different configs which I constructed in the wrapper).

mfeurer:
Sorry for the long delay. We have decided that it would make more sense to make the acquisition function work directly on configurations instead of arrays with the `__call__` function of the base acquisition function class being responsible for the conversion from configurations to arrays. For your usecase, you could then simply subclass the acquisition function and use that.

PR #221, opened 2017-04-11
MAINT refactor EPM module

mlindauer:
* AbstractEPM handles PCA and marginalization over instances
* AbstractEPM can be seen as a wrapper around all other EPMs -- should make it easier to add other EPM models
* all instances of AbstractEPM have to call the inherited constructor
* remove a check in predict_marginalized_over_instances() since it is
not guarantueed that self.types is available

### DIFF
--- a/smac/epm/base_epm.py
+++ b/smac/epm/base_epm.py
@@ -194,9 +194,4 @@ class AbstractEPM(object):
             var[i,0] = var_x
             mean[i,0] = np.mean(means)
 
-        if len(mean.shape) == 1:
-            mean = mean.reshape((-1, 1))
-        if len(var.shape) == 1:
-            var = var.reshape((-1, 1))
-
         return mean, var
--- a/smac/epm/base_epm.py
+++ b/smac/epm/base_epm.py
@@ -45,6 +45,9 @@ class AbstractEPM(object):
         if self.pca_dims and self.n_feats > self.pca_dims:
             self.pca = PCA(n_components=self.pca_dims)
             self.scaler = MinMaxScaler()
+            
+        # Never use a lower variance than this
+        self.var_threshold = 10 ** -5
 
     def train(self, X, Y, **kwargs):
         '''Trains the random forest on X and y.
@@ -61,6 +64,9 @@ class AbstractEPM(object):
         -------
         self
         '''
+        
+        self.n_params = X.shape[1] - self.n_feats
+        
         # reduce dimensionality of features of larger than PCA_DIM
         if self.pca:
             X_feats = X[:,:-self.n_feats]
@@ -69,7 +75,6 @@ class AbstractEPM(object):
             X_feats = np.nan_to_num(X_feats)  # if features with max == min
             # PCA
             X_feats = self.pca.fit_transform(X_feats)
-            self.n_params = X.shape[1] - self.n_feats
             X = np.hstack((X[:, :self.n_params], X_feats ))
             if hasattr(self, "types"):
                 # for RF, adapt types list
@@ -145,8 +150,7 @@ class AbstractEPM(object):
 
         Parameters
         ----------
-        X : np.ndarray of shape = [n_samples, n_features (config + instance
-        features)]
+        X : np.ndarray of shape = [n_samples, n_features (config)]
 
         Returns
         -------
@@ -170,8 +174,13 @@ class AbstractEPM(object):
             raise ValueError(
                 'Expected 2d array, got %dd array!' % len(X.shape))
 
-        mean = np.zeros(X.shape[0])
-        var = np.zeros(X.shape[0])
+        if X.shape[1] != self.n_params:
+            raise ValueError(
+                "Rows in X should have %d entries "
+                "but have %d!" %(self.n_params, X.shape[1]))
+
+        mean = np.zeros((X.shape[0],1))
+        var = np.zeros((X.shape[0],1))
         for i, x in enumerate(X):
             X_ = np.hstack(
                 (np.tile(x, (n_instances, 1)), self.instance_features))
@@ -179,14 +188,11 @@ class AbstractEPM(object):
             # use only mean of variance and not the variance of the mean here
             # since we don't want to reason about the instance hardness distribution
             var_x = np.mean(vars) # + np.var(means)
-            if var_x < self.var_threshold:
+            if var_x < self.var_threshold or np.isnan(var_x):
                 var_x = self.var_threshold
 
-            var[i] = var_x
-            mean[i] = np.mean(means)
-
-        var[var < self.var_threshold] = self.var_threshold
-        var[np.isnan(var)] = self.var_threshold
+            var[i,0] = var_x
+            mean[i,0] = np.mean(means)
 
         if len(mean.shape) == 1:
             mean = mean.reshape((-1, 1))
--- a/smac/epm/base_epm.py
+++ b/smac/epm/base_epm.py
@@ -15,26 +15,33 @@ __version__ = "0.0.1"
 class AbstractEPM(object):
     '''Abstract implementation of the EPM API '''
 
-    def __init__(self, pca_dims:float=None, n_feats:int=0):
+    def __init__(self, 
+                 instance_features:np.ndarray=None,
+                 pca_dims:float=None):
         '''
         initialize random number generator
 
         Parameters
         ----------
+        instance_features: np.ndarray (I, K)
+            Contains the K dimensional instance features
+            of the I different instances
         pca_dims: float
             if set to a float, use PCA to reduce dimensionality of instance features
             also requires to set n_feats (> pca_dims)
-        n_feats: int
-            number of instances features -- always appended to X
+        
         '''
+        self.instance_features = instance_features
         self.pca_dims = pca_dims
-        self.n_feats = n_feats
+        if instance_features is not None:
+            self.n_feats = instance_features.shape[1]
+        else:
+            self.n_feats = 0
         
         self.n_params = None # will be updated on train()
         
         self.pca = None
         self.scaler = None
-        print(pca_dims)
         if self.pca_dims and self.n_feats > self.pca_dims:
             self.pca = PCA(n_components=self.pca_dims)
             self.scaler = MinMaxScaler()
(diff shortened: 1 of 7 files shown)


smac/epm/base_epm.py
```diff
@@ -12,64 +15,182 @@
...
+    def predict_marginalized_over_instances(self, X:np.ndarray):
+        '''Predict mean and variance marginalized over all instances.
 
         Returns the predictive mean and variance marginalised over all
         instances for a set of configurations.
 
         Parameters
         ----------
-        X : np.ndarray of shape = [n_features (config), ]
+        X : np.ndarray of shape = [n_samples, n_features (config + instance
```
KEggensperger: Is this true? 
It seems like instance features are added in line 177?

smac/epm/base_epm.py
```diff
@@ -12,64 +15,182 @@
...
+            # use only mean of variance and not the variance of the mean here
+            # since we don't want to reason about the instance hardness distribution
+            var_x = np.mean(vars) # + np.var(means)
+            if var_x < self.var_threshold:
+                var_x = self.var_threshold
+
+            var[i] = var_x
+            mean[i] = np.mean(means)
+
+        var[var < self.var_threshold] = self.var_threshold
```
KEggensperger: This is already done in line 182

smac/epm/base_epm.py
```diff
@@ -12,64 +15,182 @@
...
+        var[np.isnan(var)] = self.var_threshold
```
KEggensperger: see above

smac/epm/base_epm.py
```diff
@@ -12,64 +15,182 @@
...
+
+        if len(mean.shape) == 1:
```
KEggensperger: If you set the shape correctly when initialising with numpy.zero, you don't need this here

smac/epm/rf_with_instances.py
```diff
@@ -62,9 +58,11 @@ def __init__(self, types,
                  max_depth=20,
                  eps_purity=1e-8,
                  max_num_nodes=1000,
-                 seed=42):
+                 seed=42,
+                 **kwargs):
+
+        super().__init__(**kwargs)
```
KEggensperger: This will fail if instance_features and num_pca_dim are not given. Is this intended?

smac/epm/base_epm.py
KEggensperger: I also don't see where self.var_threshold is initialized.

smac/epm/uncorrelated_mo_rf_with_instances.py
```diff
@@ -25,13 +25,15 @@ def __init__(self, target_names, types, **kwargs):
             See RandomForestWithInstances documentation
 
         """
+        super().__init__(**kwargs)
```
KEggensperger: See comment above

smac/epm/base_epm.py
```diff
@@ -12,64 +15,182 @@
...
+            The corresponding target values. n_objectives must match the
+            number of target names specified in the constructor.
+
+        Returns
+        -------
+        self
         '''
-        raise NotImplementedError()
+        # reduce dimensionality of features of larger than PCA_DIM
+        if self.pca:
```
KEggensperger: Is it necessary to compute PCA every time train is called? I would assume that instance features stay the same between multiple calls to train? 

This also means that PCA is only computed using the instances that are in the training set, doesn't it? Is that what we want?

smac/epm/rf_with_instances.py
mlindauer: why should it fail? `AbstractEPM` has default values for both arguments.

smac/epm/base_epm.py
mlindauer: This was the reason in the first place for changing it. For warmstarting (and other external data), the instance set is not limited to our current training instances.

smac/epm/rf_with_instances.py
KEggensperger: Sure, you are right.

smac/epm/base_epm.py
mlindauer: Fixed

smac/epm/uncorrelated_mo_rf_with_instances.py
mlindauer: see above ;-)

smac/epm/base_epm.py
mlindauer: fixed.

smac/epm/base_epm.py
mlindauer: fixed.

smac/epm/base_epm.py
mlindauer: fixed

smac/epm/base_epm.py
mlindauer: fixed

mlindauer:
@KEggensperger thanks for the thorough review. I hope all your comments are addressed now. If not, please comment again.

smac/epm/base_epm.py
```diff
@@ -12,64 +15,188 @@
...
 
         Parameters
         ----------
-        rng : np.random.RandomState
+        instance_features: np.ndarray (I, K)
+            Contains the K dimensional instance features
+            of the I different instances
+        pca_dims: float
+            if set to a float, use PCA to reduce dimensionality of instance features
+            also requires to set n_feats (> pca_dims)
```
mfeurer: I think the parameter name of this parameter should more closely follow the parameter name of the scikit-learn parameter, as it is only a wrapper to the scikit-learn parameter `n_components`. Also, you can use `typing.Union[int, float]` to show that the parameter can be both int and float.

smac/epm/base_epm.py
```diff
@@ -12,64 +15,188 @@
...
+            # use only mean of variance and not the variance of the mean here
+            # since we don't want to reason about the instance hardness distribution
+            var_x = np.mean(vars) # + np.var(means)
+            if var_x < self.var_threshold or np.isnan(var_x):
+                var_x = self.var_threshold
+
+            var[i,0] = var_x
+            mean[i,0] = np.mean(means)
+
+        if len(mean.shape) == 1:
```
KEggensperger: This can now be deleted

smac/epm/base_epm.py
```diff
@@ -12,64 +15,183 @@
...
-        X: np.ndarray (N, D)
-            Input data points. The dimensionality of X is (N, D),
-            with N as the number of points and D is the number of features.
-        Y: np.ndarray (N, 1)
-            The corresponding target values.
+        X : np.ndarray [n_samples, n_features (config + instance features)]
+            Input data points.
+        Y : np.ndarray [n_samples, n_objectives]
+            The corresponding target values. n_objectives must match the
+            number of target names specified in the constructor.
```
mfeurer: There is no argument `n_objectives` in the constructor.

smac/epm/base_epm.py
```diff
@@ -12,64 +15,183 @@
...
+        -------
+        self
         '''
-        raise NotImplementedError()
+        
+        self.n_params = X.shape[1] - self.n_feats
+        
+        # reduce dimensionality of features of larger than PCA_DIM
+        if self.pca:
+            X_feats = X[:,:-self.n_feats]
```
mfeurer: In the module `runhistory2epm` I get the impression that the [configuration comes before the features](https://github.com/automl/SMAC3/blob/94e83c98c4c7c32e336ace9b64ea767a1013f194/smac/runhistory/runhistory2epm.py#L344). I think you have to do `X_feats = X[:, -self.n_feats:]`.

KEggensperger:
Now the changes look good to me. 

But as far as I understand the coverage report none of our tests uses PCA:
https://codecov.io/gh/automl/SMAC3/pull/221/changes#L46

Would it be possible to add a test using PCA for either the base_epm class or tests for rf_with_instances or both?

smac/epm/base_epm.py
```diff
@@ -12,64 +15,183 @@
...
+            # scale features
+            X_feats = self.scaler.fit_transform(X_feats)
```
mfeurer: Where is the scaler initialized?

smac/epm/base_epm.py
mlindauer: I could agree with `pca_components`, but `n_components` is very vague for our EPMs.

smac/epm/base_epm.py
mlindauer: there was never such an argument. why should there be now one?

smac/epm/base_epm.py
mlindauer: very good catch. thanks. fixed

smac/epm/base_epm.py
mfeurer: Makes sense.

smac/epm/base_epm.py
mfeurer: I don't know. But it wasn't in the base class' docstring before.

smac/epm/base_epm.py
mlindauer: in the constructor!?

smac/epm/base_epm.py
mlindauer: furthermore, since I would like to avoid to accept floats here since we want to check whether I can apply PCA (#feats > #pca_components). if we accept floats as pca_components, we would need to add an ugly type check

smac/epm/base_epm.py
mfeurer: Yup, not sure why I didn't see this.

smac/epm/base_epm.py
mlindauer: so, what exactly is your change request?

mfeurer: It references number of targets in the constructor, but there is no such argument in the constructor. I think some rewording of the docstring like `n_objectives is always one, except changed by a subclass`.

mlindauer: That's true for the `predict_marginalized_over_instances`, but for all other functions the subclass specifies the number of objectives it can handle. I could add a corresponding note in the docstrings if you wish

mfeurer: Yes, thank you.

mlindauer:
Why is travis CI failing? Anyone any clue? I don't understand it, in particular because I haven't touched the intensification code.

KEggensperger:
I reran the test and it worked fine. Probably the execution time of one run was too long and it received a TIMEOUT causing that unittest to fail

mlindauer:
Thanks for rerunning it. However, non-deterministic fails of unit test could be a problem....

PR #224, opened 2017-04-13
increase mem_limit for rf-example

shukon:
fixing issue/#223

- increase memory_limit to 3072 and runcount-limit to 200
- make example non-deterministic (since it actually is a randomized algorithm)
- use boston-housing-data from sklearn
- use cross-validation from sklearn

### DIFF
--- a/examples/rf.py
+++ b/examples/rf.py
@@ -3,8 +3,10 @@ import os
 import inspect
 
 import numpy as np
-from sklearn.model_selection import KFold
+from sklearn.metrics import make_scorer
+from sklearn.model_selection import cross_val_score
 from sklearn.ensemble import RandomForestRegressor
+from sklearn.datasets import load_boston
 
 from smac.configspace import ConfigurationSpace
 from ConfigSpace.hyperparameters import CategoricalHyperparameter, \
@@ -13,6 +15,8 @@ from ConfigSpace.hyperparameters import CategoricalHyperparameter, \
 from smac.tae.execute_func import ExecuteTAFuncDict
 from smac.scenario.scenario import Scenario
 from smac.facade.smac_facade import SMAC
+
+boston = load_boston()
 
 def rfr(cfg, seed):
     """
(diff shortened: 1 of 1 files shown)


examples/rf.py
```diff
@@ -100,11 +87,11 @@ def rfr(cfg, seed):
...
 # SMAC scenario oject
-scenario = Scenario({"run_obj": "quality",  # we optimize quality (alternative runtime)
-                     "runcount-limit": 20,  # maximum number of function evaluations
-                     "cs": cs,              # configuration space
-                     "deterministic": "true",
-                     "memory_limit": 1024,
+scenario = Scenario({"run_obj": "quality",   # we optimize quality (alternative runtime)
+                     "runcount-limit": 200,  # maximum number of function evaluations
+                     "cs": cs,               # configuration space
+                     "deterministic": "false",
```
mfeurer: If setting this to be random, you should also set a value for `maxR`. In order to not confuse ML people, we should maybe keep this to be deterministic.

examples/rf.py
```diff
@@ -89,7 +76,7 @@ def rfr(cfg, seed):
 # Or we can add multiple hyperparameters at once:
 num_trees = UniformIntegerHyperparameter("num_trees", 10, 50, default=10)
 max_depth = UniformIntegerHyperparameter("max_depth", 20, 30, default=20)
```
mfeurer: Hm, maximum depth of a random forest is normally not tuned if this isn't a problem with memory.

examples/rf.py
```diff
@@ -45,36 +49,19 @@ def rfr(cfg, seed):
...
-        rfr.fit(X_train, y_train)
-
-        y_pred = rfr.predict(X_test)
-
-        # We use root mean square error as performance measure
-        rmse = np.sqrt(np.mean((y_pred - y_test)**2))
-        rmses.append(rmse)
-    return np.mean(rmses)
+    def rmse(y, y_pred):
+        return np.sqrt(np.mean((y_pred - y)**2))
```
mfeurer: Shouldn't there be a sum around the `np.sqrt` to return the error on the dataset?

examples/rf.py
```diff
@@ -45,36 +49,19 @@ def rfr(cfg, seed):
...
+    # Creating root mean square error for sklearns crossvalidation
+    rmse_scorer = make_scorer(rmse, greater_is_better=False)
+    score = cross_val_score(rfr, boston.data, boston.target, cv=11, scoring=rmse_scorer)
```
mfeurer: Could you please rename the target function to not have the same name as the estimator object?

examples/rf.py
shukon: I don't think so, since it returns a single float right now which is interpreted by the cross_val_score. I chose this over the old version for readability, would you prefer the old (explicit) version?

examples/rf.py
shukon: Done

examples/rf.py
shukon: Done

examples/rf.py
shukon: Alright. Reset to deterministic.

examples/rf.py
mfeurer: My bad, I overlooked the mean inside the `sqrt`.

Issue #225, opened 2017-04-15
Optimization goes crazy when target algorithm returns np.NaN


mfeurer:
If the target algorithm return `np.NaN`, the following check in `intensification.py` evaluates to `False` ([here](https://github.com/automl/SMAC3/blob/master/smac/intensification/intensification.py#L432)):

    if chal_perf > inc_perf and len(chall_runs) >= self.minR:

This results in an incumbent change. If the next configuration is working fine again (i.e. not returning an `np.NaN`), it becomes the new incumbent and the optimization goes on normally. If all configurations return `np.NaN`, each new configuration will become the new incumbent.

Can be reproduced by changing the return value in the rosenbrock example to always return NaN.

shukon:
Does it make sense to treat NaN like CRASHED, so this could be caught at TAE-level? Something like
```
if self.run_obj == "runtime" and np.isnan(runtime) or np.isnan(cost):
    status = StatusType.CRASHED
```

mfeurer:
Good idea :) @KEggensperger @mlindauer what do you think?

KEggensperger:
Yes, that's what I would have proposed as well: Treating NaNs as CRASHED.
I assume this also applies to -inf and inf

shukon:
Implemented in branch https://github.com/automl/SMAC3/tree/issue/%23225_check_TA_for_nan.

mlindauer:
this issue is solved by adding a scenario option to define the quality of crashed runs. see #229

mlindauer:
Unfortunately, this issue is not completely solved.
Unexpectedly, one of my TAs returned a nan and SMAC accepted it without complaining enough. 
So, we have to look in this issue again.

mlindauer:
I couldn't reproduce it, so closed again.

PR #229, opened 2017-04-26
ADD check for NaN and inf in TAE

shukon:
Adding check as proposed in #225


smac/tae/execute_ta_run.py
```diff
@@ -165,6 +171,7 @@ def start(self, config:Configuration,
 
         if self.run_obj == "runtime":
             if status != StatusType.SUCCESS:
+                print(cutoff, self.par_factor)
```
mlindauer: please rm this print statement

smac/tae/execute_ta_run.py
```diff
@@ -149,6 +149,12 @@ def start(self, config:Configuration,
                                                           seed=seed,
                                                           instance_specific=instance_specific)
 
+        # Catch NaN or inf.
+        if (self.run_obj == 'runtime' and not np.isfinite(runtime) or
+            self.run_obj == 'quality' and not np.isfinite(cost)):
+            status = StatusType.CRASHED
+            runtime, cost = 0, 0
```
mfeurer: Adding a cost of `0` has led to some issues in Java-SMAC. Not sure if we should go this way. Also, we have observed a runtime which could be used here.

shukon: I removed the line. Crashed runs won't be considered anyway, so they might aswell have NaNs as cost/runtime.

mlindauer: @mfeurer should we show a warning in such a case? I would say that NaN returned values are an indication for unwanted behavior of the target algorithm

mfeurer: Yes, that's a good idea.

shukon: Done. Warns now that "Target Algorithm returned NaN or inf as cost or runtime. Algorithm run is treated as CRASHED."

mlindauer:
Right now, we would still return NaNs, which could be a problem.

@shukon Could you please a scenario option for a default "cost values" for crashed runs and use it in the TAE.

shukon:
Implemented cost-for-crash option, needs a review though. Now never returning NaN or inf costs, but possibly NaN or inf runtimes. Is this ok or should runtime always be finite? (not sure about intentions here)

mlindauer:
@mfeurer and @KEggensperger  please review this PR again. I'm happy with it.

smac/tae/execute_ta_run.py
```diff
@@ -149,7 +160,22 @@ def start(self, config:Configuration,
...
 
-        if self.stats.ta_runs == 0 and status == StatusType.CRASHED:
+        # update SMAC stats
+        self.stats.ta_runs += 1
+        self.stats.ta_time_used += float(runtime)
+
+        # Catch NaN or inf.
+        if (self.run_obj == 'runtime' and not np.isfinite(runtime) or
+                self.run_obj == 'quality' and not np.isfinite(cost)):
+            self.logger.warning("Target Algorithm returned NaN or inf as {} "
```
mfeurer: There's a dot missing at the end of this sentence.

test/test_tae/test_exec_tae_run.py
```diff
@@ -81,6 +83,85 @@ def test_start_crash_first_run(self, test_run):
...
+        def get_tae(obj):
+            """ Create ExecuteTARun-object for testing. """
+            scen = Scenario(scenario={'cs': ConfigurationSpace(), 'run_obj': obj,
+                'abort_on_first_run_crash': '0', 'cutoff_time': '10'}, cmd_args=None)
+            stats = Stats(scen)
+            stats.start_timing()
+            eta = ExecuteTARun(ta=lambda *args: None, stats=stats, run_obj=obj)
+            # Add first run to not trigger FirstRunCrashedException
+            test_run.return_value = StatusType.SUCCESS, 1, 1, {}
+            eta.start(config={}, instance=1)
```
mfeurer: What is this call doing here? I assume it prevents an exception if the first run crashes. If yes, `stats.ta_run += 1` would me more intuitive.

mlindauer:
@KEggensperger since this is really important PR, could you please also review it

smac/tae/execute_ta_run.py
```diff
@@ -159,17 +185,21 @@ def start(self, config:Configuration,
...
-        self.stats.ta_runs += 1
-        self.stats.ta_time_used += float(runtime)
-
         if self.run_obj == "runtime":
-            if status != StatusType.SUCCESS:
-                cost = cutoff * self.par_factor
-            else:
+            if cutoff is not None and runtime > self.par_factor * cutoff:
+                self.logger.warn("Returned running time is larger "
+                                 "than 2 times the passed cutoff time. "
```
KEggensperger: This warning seems to be incorrect. This if-statement is true when the runtime is larger that par\*cutoff and not 2\*cutoff.

smac/tae/execute_ta_run.py
```diff
@@ -159,17 +185,21 @@ def start(self, config:Configuration,
...
+                                 "Clamping to {} x cutoff.".format(self.par_factor))
+                runtime = cutoff * self.par_factor
+            if status == StatusType.SUCCESS:
                 cost = runtime
+            else:
+                cost = cutoff * self.par_factor
```
KEggensperger: This statement is also true if cutoff is None. If this can happen, we don#t want to use cutoff\*par.
Is there a case where cutoff is None and we optimize runtime? If no, the first if can also be changed.

smac/tae/execute_ta_run.py
mlindauer: Thanks. fixed

smac/tae/execute_ta_run.py
mlindauer: Good point. I removed the cutoff=none in the above if-statement and added an error log message if running is used but no cutoff was passed -- this is not intended right now; however, if the target algorithm always returned SUCCESS runs, it would not be a problem.

mlindauer:
@shukon could you please resolve the merge conflict.

PR #233, opened 2017-04-27
FIX #117

### LINKED ISSUE #117: Why do we round the cutoff?
Why do we use math.ceil on the cutoff and cast it to `int` in `execute_ta_run`?

```
        if cutoff is not None:
            cutoff = int(math.ceil(cutoff))
```

mlindauer:


KEggensperger:
This does not fix #117, but #209

mlindauer:
ups, sorry!

PR #235, opened 2017-04-28
obtain random samples for intensification from a generator

mfeurer:
This pull request is work in progress and and right now just a proof of concept. I'll add unit tests and documentation if the general ideas is approved.

Currently, the challengers are all created before calling the intensify function. As each second challenger is random, a lot of random samples are wasted creating this list of challengers. This PR replaces the ordinary python list by a custom list/iterator. Instead of holding a list of random configurations, it has all non-random configurations and a configuration sampler. In each second iteration, instead of returning the next element from the list of challengers, it samples a new random configuration.

On the RF example, this reduced the time to obtain a new list of challengers from 0.0181 to 0.0121 seconds (improvement of ~30%).

### DIFF
--- a/smac/optimizer/smbo.py
+++ b/smac/optimizer/smbo.py
@@ -379,6 +379,6 @@ class ChallengerList(object):
             return config
         else:
             self._next_is_random = True
+            config = self.challengers[self._index]
             self._index += 1
-            config = self.challengers[self._index - 1]
             return config
--- a/smac/optimizer/smbo.py
+++ b/smac/optimizer/smbo.py
@@ -355,12 +355,15 @@ class ChallengerList(object):
         return self
 
     def __next__(self):
-        if self._index == len(self.challengers):
+        if self._index == len(self.challengers) and not self._next_is_random:
             raise StopIteration
         elif self._next_is_random:
-            return self.configuration_space.sample()
+            self._next_is_random = False
+            config = self.configuration_space.sample_configuration()
+            config.origin = 'Random Search'
+            return config
         else:
+            self._next_is_random = True
             self._index += 1
             config = self.challengers[self._index - 1]
-            config.origin = 'Random Search'
             return config
--- a/smac/optimizer/smbo.py
+++ b/smac/optimizer/smbo.py
@@ -344,6 +344,21 @@ class SMBO(BaseSolver):
 
 
 class ChallengerList(object):
+    """Helper class to interleave random configurations in a list of challengers.
+
+    Provides an iterator which returns a random configuration in each second
+    iteration. Reduces time necessary to generate a list of new challengers
+    as one does not need to sample several hundreds of random configurations
+    in each iteration which are never looked at.
+
+    Parameters
+    ----------
+    challengers : list
+        List of challengers (without interleaved random configurations)
+
+    configuration_space : ConfigurationSpace
+        ConfigurationSpace from which to sample new random configurations.
+    """
 
     def __init__(self, challengers, configuration_space):
         self.challengers = challengers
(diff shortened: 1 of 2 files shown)


smac/optimizer/smbo.py
```diff
@@ -347,3 +342,25 @@ def _get_timebound_for_intensification(self, time_spent):
...
+        self._index = 0
+        self._next_is_random = False
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        if self._index == len(self.challengers):
+            raise StopIteration
+        elif self._next_is_random:
```
mlindauer: `_next_is_random` is never set to True?

I would prefer some way to control the frequency of random configs, e.g.
`elif self.index % K == 0`

smac/optimizer/smbo.py
```diff
@@ -347,3 +342,25 @@ def _get_timebound_for_intensification(self, time_spent):
...
+            return self.configuration_space.sample()
+        else:
+            self._index += 1
+            config = self.challengers[self._index - 1]
+            config.origin = 'Random Search'
```
mlindauer: I don't get why all challengers in `self.challengers` stem from "random search"?

mfeurer: That's obviously an issue, fixed.

smac/optimizer/smbo.py
mfeurer: I fixed the first comment.

Regarding the second comment, we'd have to adapt intensify for that as well, or make sure that the second configuration is still random. Anyway, I think this is something for a second PR as this one is about increasing efficiency.

mlindauer: Why do we have to change intensify for this? Maybe this was a misunderstanding: I would prefer that not every second configurations is a random configuration but we can set a parameter to change the frequency of random configurations. This would perfectly fit in this PR, since we only have to change a single line.

mfeurer: Okay, let's assume I set the parameter to draw each third configuration at random. Then, the third element in the list of challengers would be a random configuration. This would not guarantee that SMAC actually looks at a random configuration ever, because intensifiy only guarantees to look at two configurations. Therefore, I think one needs to be cautious about this.

mlindauer: I see your point. But as far as I know, we don't have any (empirical) evidence that random configurations are necessary at all. So, I don't know whether it is really important that each intensify call runs at least one random configuration. 
Of course, I don't want to change the default behavior of SMAC. But for the HPOlib benchmarks from Aaron, I would expect that we need a lower frequency of random configurations anyway.

mfeurer:
PING @KEggensperger and @mlindauer as you both like the idea, I'd appreciate a review of this PR.

mlindauer:
@mfeurer I commented 33 minutes ago and have not got a reply...

mfeurer:
I don't know an answer to that. I can implement this, but it's not one line... (or you can't use it from command line).

smac/optimizer/smbo.py
```diff
@@ -347,3 +342,43 @@ def _get_timebound_for_intensification(self, time_spent):
...
+            raise StopIteration
+        elif self._next_is_random:
+            self._next_is_random = False
+            config = self.configuration_space.sample_configuration()
+            config.origin = 'Random Search'
+            return config
+        else:
+            self._next_is_random = True
+            self._index += 1
+            config = self.challengers[self._index - 1]
```
KEggensperger: You could switch 382 and 383?

mfeurer: Good catch, thanks!

PR #246, opened 2017-05-10
MAINT output format

mlindauer:
* adapt logging format depending on logging level
* add output in the beginning of rh2epm
* running pep8 tool on smac_cli.py

### DIFF
--- a/smac/smac_cli.py
+++ b/smac/smac_cli.py
@@ -31,21 +31,33 @@ class SMACCLI(object):
         '''
             constructor
         '''
-        self.logger = logging.getLogger(self.__module__ + "." + self.__class__.__name__)
+        self.logger = logging.getLogger(
+            self.__module__ + "." + self.__class__.__name__)
 
     def main_cli(self):
         '''
             main function of SMAC for CLI interface
         '''
-        self.logger.info("SMAC call: %s" %(" ".join(sys.argv)))
+        self.logger.info("SMAC call: %s" % (" ".join(sys.argv)))
 
         cmd_reader = CMDReader()
         args_, misc_args = cmd_reader.read_cmd()
 
-        logging.basicConfig(level=args_.verbose_level)
-
         root_logger = logging.getLogger()
         root_logger.setLevel(args_.verbose_level)
+        logger_handler = logging.StreamHandler(
+                stream=sys.stdout)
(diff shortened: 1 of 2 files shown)


mlindauer:
@KEggensperger could you please explain the codecov report? I haven't touched `smac/intensification/intensification.py`, but the coverage decreased `94.54% <0%> (-1.22%)` -- why? how?

smac/smac_cli.py
```diff
@@ -31,21 +31,33 @@ def __init__(self):
...
         cmd_reader = CMDReader()
         args_, misc_args = cmd_reader.read_cmd()
 
-        logging.basicConfig(level=args_.verbose_level)
-
         root_logger = logging.getLogger()
         root_logger.setLevel(args_.verbose_level)
+        logger_handler = logging.StreamHandler(
+                stream=sys.stdout)
+        if root_logger.level >= 20:
```
mfeurer: What does `20` mean? Maybe it would be good to use the respective constant like `logging.DEBUG`.

smac/smac_cli.py
```diff
@@ -31,21 +31,33 @@ def __init__(self):
...
+            formatter = logging.Formatter(
+                "%(levelname)s:\t%(message)s")
```
mfeurer: I understand that you don't want to have the name of the logger here, but we could also add the time in H:M:S format here. I think this would be highly useful.

mlindauer: I don't think this is really interesting for a "normal" user. For example, most standard Linux programs  also don't print a time stamps.

smac/smac_cli.py
mlindauer: thanks, fixed

smac/smac_cli.py
mfeurer: Maybe you're right. I'll complain the next time I see myself as a normal user and need that output ;)

mlindauer:
please note that the changes are only effective by using the cli interface. I would assume that users using the Python interface want to adjust the logging output themselves.

KEggensperger:
I guess this is a "Unexpected Coverage Change"
https://docs.codecov.io/docs/unexpected-coverage-changes

Is there maybe a nondeterministic test? I'll rerun ci and check whether codecov changes as well and then we can merge

PR #247, opened 2017-05-11
FIX #188

### LINKED ISSUE #188: Force runs on default in intensify
We should add an option such that intensification always runs the default on a new instance-seed pair, because 

1. to prevent over-tuning on heterogeneous instance sets
2. some ideas of better scaling solution quality values across instances...

mlindauer:
* new option for racing new incumbent against default configuration to
prevent over-tuning

### DIFF
--- a/smac/facade/smac_facade.py
+++ b/smac/facade/smac_facade.py
@@ -191,6 +191,8 @@ class SMAC(object):
                                       cutoff=scenario.cutoff,
                                       deterministic=scenario.deterministic,
                                       run_obj_time=scenario.run_obj == "runtime",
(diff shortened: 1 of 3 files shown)


mlindauer:
@mfeurer for this one new if-statement? If you insist, of course I can add one.

mfeurer:
I wasn't sure if it's big enough for a unit test, that's why I asked. I briefly thought about extracting it to a different function to keep the main function clean, but that's probably not necessary either.

mlindauer:
I ran some experiments on scenarios where we had over-tuning with our current release. Using this option indeed helped to recover from the over-tuning and SMAC returned to the default configuration.

Issue #52, opened 2016-09-16
Make run_obj a mandatory argument


mfeurer:
I think we should make arguments where users can really screw up the setting mandatory. One example is run_obj, which can make the algorithm behave totally wrong for a scenario. By making a few important choices required, we lower the chance of accidental misconfiguration.

mfeurer:
And, if the run objective is runtime, cutoff should be a mandatory argument as well (if not given, SMAC crashes).

mfeurer:
@mlindauer what's your opinion on this?

mlindauer:
I agree that cutoff should be mandatory if it is a running time scenario.

I'm undecided whether run_obj should be mandatory. I see your point.
However in my experience, it is very crucial for new users to have a very low initial hurdle for the first use. With more required arguments, it gets a lot more complex for new users which could prevent  them to try SMAC.

mlindauer:
Ok, we voted for making "run_obj" as a mandatory field.

@shukon could you please create a PR to resolve this issue and change it the docu accordingly

PR #253, opened 2017-05-22
Issue #52 make run_obj mandatory

shukon:
Fixing issue #52, making run_obj mandatory and cutoff_time mandatory if run_obj is runtime.

### DIFF
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -244,7 +244,7 @@ class Scenario(object):
         self.add_argument(name='paramfile', help=None, dest='pcs_fn',
                           mutually_exclusive_group='cs')
         self.add_argument(name='run_obj', help=None, default='runtime',
-                          required=True)
+                          required=True, choice=['runtime', 'quality'])
         self.add_argument(name='overall_obj', help=None, default='par10')
         self.add_argument(name='cost_for_crash', default=float(MAXINT),
                           help="Defines the cost-value for crashed runs "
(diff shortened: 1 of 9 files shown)


smac/scenario/scenario.py
```diff
@@ -77,6 +77,10 @@ def __init__(self, scenario, cmd_args=None, run_id=1):
         self._groups = defaultdict(set)
         self._add_arguments()
 
+        # Make cutoff mandatory if run_obj is runtime
+        if 'run_obj' in scenario and scenario['run_obj'] == 'runtime':
```
KEggensperger: Can 'run_obj' be not in scenario? If the answer to this question is yes, we should move this check to another place.

smac/scenario/scenario.py
```diff
@@ -239,7 +243,8 @@ def _add_arguments(self):
                           help=None, callback=float)
         self.add_argument(name='paramfile', help=None, dest='pcs_fn',
                           mutually_exclusive_group='cs')
-        self.add_argument(name='run_obj', help=None, default='runtime')
+        self.add_argument(name='run_obj', help=None, default='runtime',
```
KEggensperger: This should be a choice argument with the options "cost" and "runtime"

smac/scenario/scenario.py
shukon: Not sure where that would be? To not have it in scenario, a lot of code would need changes I think.

smac/scenario/scenario.py
shukon: Agree!

KEggensperger: My bad. It is 'quality'

smac/scenario/scenario.py
KEggensperger: I assume that we always have 'run_obj' (as it is a required argument) and there is no need to check whether 'run_obj' exists.

But, as you check for it I wondered whether there is a case where we don't have 'run_obj'. If such a case would exist, we would need to handle it in this PR.  I assume there is no such case, so you can safely remove that check?

shukon: Ah, now I get it. Yes, can do.

PR #254, opened 2017-05-24
Issue/#199 simplify rh2epm

shukon:
Build rh-data-dicts through dict-comprehension, remove runhistory.__select, remove RunType.

### DIFF
--- a/smac/runhistory/runhistory2epm.py
+++ b/smac/runhistory/runhistory2epm.py
@@ -20,21 +20,11 @@ __email__ = "eggenspk@cs.uni-freiburg.de"
 __version__ = "0.0.1"
 
 
-class RunType(object):
-    '''
-       class to define numbers for status types.
-       Makes life easier in select_runs
-    '''
-    SUCCESS = 1
-    TIMEOUT = 2
-    CENSORED = 3
-
-
 class AbstractRunHistory2EPM(object):
     __metaclass__ = abc.ABCMeta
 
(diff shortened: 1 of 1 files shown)


smac/runhistory/runhistory2epm.py
```diff
@@ -153,29 +143,31 @@ def transform(self, runhistory):
...
-        self.logger.debug("Transform Rh into X,y format")
-        
+        self.logger.debug("Transform runhistory into X,y format")
+
         assert isinstance(runhistory, RunHistory)
 
         # consider only successfully finished runs
-        s_run_dict = self.__select_runs(rh_data=copy.deepcopy(runhistory.data),
-                                        select=RunType.SUCCESS)
+        s_run_dict = {run: runhistory.data[run] for run in runhistory.data.keys()
```
mfeurer: There's a deepcopy happening in the old version. Do you have any idea on whether that has an effect?

shukon: I figured it was because we were iterating over it in https://github.com/automl/SMAC3/blob/development/smac/runhistory/runhistory2epm.py#L252-L254, however that too was only reading. Maybe for caution, if someone decided to modify the __select? Since we are not changing anything in the comprehension, I don't think it's necessary.

mfeurer: As we don't change anything later, I guess this is fine

PR #257, opened 2017-05-29
Fix forest options

mfeurer:
Apparently, a lot of features which were so far passed to the random forest options object need to be passed to a tree options object. This PR does the necessary changes. Furthermore, as having only 1000 nodes per tree only allows to model ~3000 distinct values, I increased the number to ~1M nodes per tree. Please keep in mind that there was no limit at all so far.

### DIFF
--- a/smac/epm/rf_with_instances.py
+++ b/smac/epm/rf_with_instances.py
@@ -62,7 +62,7 @@ class RandomForestWithInstances(AbstractEPM):
                  min_samples_leaf=3,
                  max_depth=20,
                  eps_purity=1e-8,
-                 max_num_nodes=1000,
+                 max_num_nodes=2**20,
                  seed=42,
                  **kwargs):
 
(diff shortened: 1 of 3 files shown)


AndreBiedenkapp:
@mfeurer, @sfalkner I thought that the forest_opts object of pyrfr was directly responsible for propagating the necessary information down to the individual trees.

I just looked into which tests fail and why.
First things first. Testing the test case [test_predict_with_actual_values](https://github.com/automl/SMAC3/blob/3adefabaad99eebf799ed243b339ff880b3b8086/test/test_epm/test_rfr_with_instances.py#L128-L161) fails because the predictions are slightly off.

The input to the tree is as follows in this case:

| Feat_1 | Feat_2 | Feat_3 | y |
|     :---:      |   :---:      |    :---:      |        ---: |
|  0 | 0 | 0 | 0.1 |
|  0 | 0 | 1 | 0.2 |
|  0 | 1 | 0 | 9.0 |
|  0 | 1 | 1 | 9.2 |
|  1 | 0 | 0 | 100.0 |
|  1 | 0 | 1 | 100.2 |
|  1 | 1 | 0 | 109.0 |
|  1 | 1 | 1 | 109.2 |

And just to make sure the tree can completely overfit the X matrix and y vector are repeated 8 times.
Before the introduced changes the tree perfectly was able to predict ŷ for [0, 0, 0] to be 0.1

I played around a bit and it seems like forest_opts.tree_opts is the culprit here. If only max_num_nodes is set to 2**20. The tree is still able to perfectly predict the example values.
However if tree_opts is used the following results is produced:

| Feat_1 | Feat_2 | Feat_3 | y_true |  y_pred |
|     :---:      |   :---:      |    :---:      |        ---: |      ----: |
|  0 | 0 | 0 | 0.1 | 12.45 |
|  0 | 0 | 1 | 0.2 |   5.97 |
|  0 | 1 | 0 | 9.0 | 14.56 |
|  0 | 1 | 1 | 9.2 |   8.18 |
|  1 | 0 | 0 | 100.0 | 92.69 |
|  1 | 0 | 1 | 100.2 | 95.96 |
|  1 | 1 | 0 | 109.0 | 103.98 |
|  1 | 1 | 1 | 109.2 | 108.15 |

As you can see the predictions are now quite off.



The failing test in [runhistory2epm in line 201](https://github.com/automl/SMAC3/blob/forest_options/test/test_runhistory/test_runhistory2epm.py#L201) simply is caused by this the forests predictions. Now the imputed value for the capped run is 16.393 not 16.422

mfeurer:
I think the culprit is the `ratio_features`. Please try setting this to 1.0 (don't forget to fix the calculations in line 79 accordingly), and the unit test will pass. I will fix the unit tests in a bit.

mfeurer:
@AndreBiedenkapp thanks for the comments. My latests commit fixes the unit tests (at least on my local machine).

AndreBiedenkapp:
Then everything seems perfectly fine.

Issue #256, opened 2017-05-29
warm starting SMAC crashes due to an empty runhistory


aaronkl:
SMAC crashes with the error message below if I warm start it by calling the choose_next() function of the SMBO class.
SMAC usually starts the local search from the previously best seen configuration. However, since there are no previous runs the run history is empty and thus SMAC tries to index an empty list.

  File "/home/kleinaa/virtualenv/local/lib/python3.5/site-packages/smac-0.5.0-py3.5.egg/smac/optimizer/smbo.py", line 210, in choose_next
  File "/home/kleinaa/virtualenv/local/lib/python3.5/site-packages/smac-0.5.0-py3.5.egg/smac/optimizer/smbo.py", line 312, in _sort_configs_by_acq_value
  File "/home/kleinaa/virtualenv/local/lib/python3.5/site-packages/smac-0.5.0-py3.5.egg/smac/configspace/util.py", line 25, in convert_configurations_to_array
IndexError: list index out of range

mlindauer:
Do you need a fix soon?

aaronkl:
After the first evaluation I know get the error:

ValueError: runhistory is empty and the cost value of the incumbent is unknown.

mlindauer:
for documentation: if the runhistory is empty, you have to provide the cost value of the incumbent.

since labwatch is now runnning with SMAC ( using #259 ), closed

Issue #212, opened 2017-04-03
# TODO try to calculate an incumbent from the runhistory!


mlindauer:
https://github.com/automl/SMAC3/blob/master/smac/smbo/smbo.py#L184

This ToDo should be fixed.

mfeurer:
You mean that this TODO should be replaced by code that uses the incumbent calculation from the runhistory?

mlindauer:
Yes, it should probably be replaced by an implementation.
However, I'm not sure under which condition we do not have an incumbent---so, maybe we could delete this simply.

shukon:
From what I'm seeing, the only possibilities in the current scheme to have `self.incumbent = None` are:
- SMBO-object is initialized and `choose_next` is called before `run` (might be done to get challengers from outside the object, since it is not a private function? Not sure if this is realistic - maybe with a combination of customized modules and resumed optimization that reloads old runhistories/warmstart?)
-  `self.initial_design.run` returns `None` in which case we would run into trouble at latest when calling `intensify` with `incumbent=None`

mfeurer:
Regarding your first option, this is exactly what Auto-sklearn does. Auto-sklearn implements the run function itself, thus never calling the run function in the smac package (although this might change at some point).

PR #259, opened 2017-05-31
Fix/#256

mlindauer:
FIX for #256 and #212 

* FIX smbo.choose_next() for empty runhistory
* ADD unit test


mlindauer:
Aaron approved this by testing it in labwatch.

PR #260, opened 2017-05-31
FIX #258, ADD small testcase

### LINKED ISSUE #258: smac.configspace.pcs.read does not consider OrdinalHyperparameters
I got the following error, when I use the OrdinalHyperparameter `x2 ordinal {0,15} [0]' in the parameter.pcs of the branin example.

`python ../../scripts/smac --scenario branin_scenario.txt
INFO:smac.smac_cli.SMACCLI:SMAC call: ../../scripts/smac --scenario branin_scenario.txt
INFO:smac.scenario.scenario.Scenario:Reading scenario file: branin_scenario.txt
  File "/home/mrudolph/SMAC3/smac/scenario/scenario.py", line 352, in _transform_arguments
    self.cs = pcs.read(pcs_str)
  File "/home/mrudolph/anaconda3/lib/python3.6/site-packages/ConfigSpace/io/pcs.py", line 232, in read
    raise NotImplementedError("Could not parse: %s" % line)
NotImplementedError: Could not parse: x2 ordinal {0,15} [0]

During handling of the above exception, another exception occurred:

  File "../../scripts/smac", line 20, in <module>
    smac.main_cli()
  File "/home/mrudolph/SMAC3/smac/smac_cli.py", line 51, in main_cli
    run_id=args_.seed)
  File "/home/mrudolph/SMAC3/smac/scenario/scenario.py", line 102, in __init__
    self._transform_arguments()
  File "/home/mrudolph/SMAC3/smac/scenario/scenario.py", line 355, in _transform_arguments
    self.cs = pcs_new.read(pcs_str)
  File "/home/mrudolph/anaconda3/lib/python3.6/site-packages/ConfigSpace/io/pcs_new.py", line 309, in read
    raise NotImplementedError("Could not parse: %s" % line)
NotImplementedError: Could not parse: x1 [-5,10] [0]
`

AndreBiedenkapp:
I fixed the Ordinal Parameter Issue and added a small test case to make sure that the bounds are correctly set.


smac/utils/util_funcs.py
```diff
@@ -16,6 +16,11 @@ def get_types(config_space, instance_features=None):
             types[i] = n_cats
             bounds[i] = (int(n_cats), np.nan)
 
+        elif isinstance(param, (OrdinalHyperparameter)):
+            n_cats = len(param.sequence)
```
mlindauer: ordinals are treated in the same way as categoricals?
This would be wrong because the ordering is indeed important---that's the whole point of ordinals compared to categoricals.

AndreBiedenkapp: @mlindauer getting the bounds should be the same, should it not?
If not, what would be the correct way of doing it?

I'll take another look into how pyrfrs data container actually processes this information and if there's a special way of dealing with ordinals.

mlindauer: I don't know the details of the new rf API. But it is important that ordinals are not handled in the same way as categorical, because they have an ordering and categorical have not. So, the split on such feature can be different.

Without looking into the API docu, I would expect something along these lines:
`bounds[i] = (0.0, len(param.choices)-1)`
`types[i]`  should still be set to 0 since it is not a categorical feature.

AndreBiedenkapp:
@mlindauer You are correct. I had some faulty reasoning when I arrived at the first FIX. Thanks for the clarification.
Now Ordinals are perfectly distinguishable from categoricals or float/integer parameters by the forest.

PR #269, opened 2017-06-23
ADD script to plot predicted performance of all incumbents

mlindauer:


KEggensperger:
Wouldn't it be useful to plot the performance of all instances separately as light lines in the background?

mlindauer:
> Wouldn't it be useful to plot the performance of all instances separately as light lines in the background?

For a representative set, maybe yes. But I wonder whether we can see anything if we would use all n  instances  (and sometimes, n is 2000).  Do you think this would work? If yes, I'm happy to accept changes to the PR.

mlindauer:
> I don't understand why you want to use the SMAC src and not the installed version,

because I'm too lazy to install all the time the current SMAC version... but you are right. I will change it.

Issue #267, opened 2017-06-15
validation


shukon:
SMAC validation missing. A few questions regarding implementation:

- should new runs be added to existing runhistory?
- how to control "budget" for non-deterministic runs? Via Stats with wallclock/cpu-limits or rather setting a certain number of runs (i.e. at least 20 runs per instance)?
- should there be a scenario-option like "validate = True/False" to activate automatic validation after a smac-run? 
- validation on train-instances as an option?
- should one seed be used for all instances or rather completely independent instance-seed-pairs?
- anything else you think is important?

mlindauer:
> should new runs be added to existing runhistory?

smac-validate should have the option to read in already existing runhistories and use them to avoid rerunning already validated runs. however, it would be better if smac-validate would write its on runhistorie in the end consisting only of the validation runs and nothing more.

btw: the only required input arguments should be the scenario and a trajectory file.

> how to control "budget" for non-deterministic runs? Via Stats with wallclock/cpu-limits or rather setting a certain number of runs (i.e. at least 20 runs per instance)?

wallclock/cpu-limit should not matter for smac-validate. 
Repeated runs for non-deterministic algorithms is our default. 
However, the default should be set to 1.

> should there be a scenario-option like "validate = True/False" to activate automatic validation after a smac-run?

For the first version, "validate" has not be part of a smac run. Therefore, no.

> validation on train-instances as an option?

Yes.
In detail, we would need the following:
* for the instance set: TRAIN or TEST or TRAIN+TEST
* for the configuration to be validated: DEF or INC or DEF+INC or TIME
where "TIME" means that the incumbent over time will be validated,
e.g., performance of the incumbent at time steps 2^0, 2^1, 2^2, 2^3, ...

> should one seed be used for all instances or rather completely independent instance-seed-pairs?

for deterministic algorithms, the seed can be fixed to 0.
for non-deterministic algorithms, you need n instance-seed pairs depending on the option discussed in question 2. These instance-seed pairs have to be the same for all validated configurations.

> anything else you think is important?

highly parallelized validation on a cluster would be nice, i.e., 1 master that knows all the validation runs, and 1000 workers on a cluster that do the actual target algorithm runs and report the results to the master. for this, please also see my e-mail regarding `joblib`.

mfeurer:
I think an option to validate the whole trajectory as in SMAC2 would be great, something like an ALL option.

PR #271, opened 2017-06-26
Issue/#267 validation

shukon:
Implements validation as described in issue #267 .
New class validator in utils/validate.py
Processes given runhistories to impute as many runs as possible while guaranteeing that inst/seed-pairs over configs remain the same.
Uses joblib to parallelize validation-runs.


scripts/smac-validate.py
```diff
@@ -0,0 +1,68 @@
...
+                          help="path to SMAC scenario")
+    req_opts.add_argument("--trajectory", required=True,
+                          help="path to SMAC trajectory")
+    req_opts.add_argument("--output", required=True,
+                          help="path to save runhistory to")
+
+    req_opts = parser.add_argument_group("Optional Options")
+    req_opts.add_argument("--configs", default="def+inc", type=str,
+                          choices=["def", "inc", "def+inc", "time", "all"],
+                          help="what configs to evaluate")
```
mlindauer: what is "all"? please clarify all chocies in the help string

smac/utils/validate.py
```diff
@@ -0,0 +1,371 @@
...
+            self.rng = rng
+        elif isinstance(rng, int):
+            self.rng = np.random.RandomState(seed=rng)
+        else:
+            num_run = np.random.randint(MAXINT)
+            self.rng = np.random.RandomState(seed=num_run)
+
+        self.rh = RunHistory(average_cost)  # update this rh with validation-runs
+
+    def validate(self, config_mode='def', instance_mode='test', repetitions=1,
```
mlindauer: please use `typing` to annotate the types

smac/utils/validate.py
```diff
@@ -0,0 +1,371 @@
...
+        runs = self.get_runs(configs, instances, repetitions=repetitions,
+                             runhistory=runhistory)
+
+        # Create new Stats without limits
+        inf_scen = Scenario({'run_obj':'quality'})
+        inf_stats = Stats(inf_scen)
+        inf_stats.start_timing()
+
+        # Create TAE
+        tae = ExecuteTARunOld(ta=self.scen.ta,
```
mlindauer: could you please an option to alterantively use ExecuteTARunAClib

smac/utils/validate.py
```diff
@@ -0,0 +1,371 @@
...
+                             %mode)
+        if mode == "def" or mode == "def+inc":
+            configs.append(self.scen.cs.get_default_configuration())
+        if mode == "inc" or mode == "def+inc":
+            configs.append(self.traj[-1]["incumbent"])
+        if mode == "time":
+            # add configs at evaluations 2^1, 2^2, 2^3, ...
+            counter = 1
+            for entry in self.traj[:-1]:
+                if (entry["evaluations"] >= counter and
```
mlindauer: please use `wallclock_time` here instead of `evaluations`.
and please start with the first entry in the traj
and then continue with 2^-4, 2^-3, 2^-2, ...

smac/utils/validate.py
```diff
@@ -0,0 +1,371 @@
...
+        -------
+        instances: list<strings>
+            instances to be used
+        """
+        instance_mode = mode.lower()
+        if mode not in ['train', 'test', 'train+test']:
+            raise ValueError("%s not a valid option for instance_mode in validation."
+                             %mode)
+
+        instances = dict()
```
mlindauer: why is this a `dict`? The docstring also states that this should be a list.

shukon: because i understood instances to be a `dict` with `instance :instance_specifics`.. i'd correct the docstrings rather than the code. or am i on the wrong track?

smac/utils/validate.py
shukon: Only for validate or for all functions?

smac/utils/validate.py
shukon: Would it be better to add an option to pass any instance of ExecuteTARun or do we specifically want to provide Old and AClib?

smac/utils/validate.py
mlindauer: for all please

smac/utils/validate.py
mlindauer: yes, an argument to pass an tae object would be great. however, on the command line it should still be a string indicating which class to use.

smac/utils/validate.py
mlindauer: `scen.train_insts` should be a list. So I don't see why you want to convert it to a dict

shukon: my bad. i misinterpreted a line there.

shukon:
@mlindauer I think it's all covered...

smac/utils/validate.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+    def _get_configs(self, mode):
+        """
+        Return desired configs
+
+        Parameters
+        ----------
+        mode : string
+            from [def, inc, def+inc, time, all], time means eval at 2^0, 2^1, 2^2, ...
```
mlindauer: please update sequence; starting from 2**-4 .. also in docstrings above

mlindauer:
@mfeurer @KEggensperger do you also want to review this PR or should I merge it? 
Since this is an external script on top of SMAC, I think, a further review is not necessary.

smac/utils/validate.py
```diff
@@ -0,0 +1,383 @@
...
+        insts: list<strings>
+            instances to be validated
+        repetitions: int
+            number of seeds per instance/config to be evaluated
+        runhistory: RunHistory or None
+            if given, try to reuse these results and save some runs
+
+        Returns
+        -------
+        runs: list<dict<string,string,string,string>>
```
mfeurer: See comment above.

smac/utils/validate.py
```diff
@@ -0,0 +1,383 @@
...
+                    specs = self.scen.instance_specific[i] if i else ""
+                    runs.append({'config':config,
+                                 'inst':i,
+                                 'seed':seed,
+                                 'inst_specs': specs})
+
+        self.logger.info("Collected %d runs from %d configurations on %d instances "
+                         "with %d repetitions.", len(runs), len(configs), len(insts),
+                         repetitions)
+        self.logger.info("Imputed %d runs from given runhistory.", imputed)
```
mfeurer: I think the word "imputed" could be confusing for scikit-learn users, because imputed has a very different meaning there. Actually, I can't figure out myself what it means here.

shukon: I misused the word, I'd rather say "taken from runhistory", so they don't need to be recalculated.

shukon:
@mfeurer by adding a validate-function to the optimizer, did you mean whats implemented now? I added a function validate to the `SMAC`-object, that calls a function validate in the `SMBO`-object, which creates a Validator and executes the specified validation...

examples/svm.py
```diff
@@ -110,3 +110,10 @@ def svm_from_cfg(cfg):
 inc_value = svm_from_cfg(incumbent)
 
 print("Optimized Value: %.2f" % (inc_value))
+
+
+# We can also validate our results (though this makes a lot more sense with instances)
```
mfeurer: I think it still makes a lot of sense for machine learning as a user normally has a separate test set. @mlindauer do you think it would make the example too complicated to add a separate test set?

mlindauer: @mfeurer do you want to simply change the dataset in `svm_from_cfg() `,
or do you want to use training and test instances in the scenario?
The first should be trivial but a bit hacky;
the latter would be the intended way from SMAC's perspective, but harder to understand (/more complicated)

mfeurer: I thought of the second one. We could have the RF example be the rather simple use case without categorical hyperparameters and instances and the SVM be the complicated one showcasing what SMAC can do as well.

shukon: So, to see if I get this right:
Let's say we have instances 1-50 for training and 60-100 for test.
To evaluate one training instance (for example 3), I train the model on the remaining training-instances (that would be 1, 2, 4 - 50)? And, in turn, if I evaluate a test-instance, I train on remaining test-instances? This seems very computationally expensive to me, training a SVM for every instance. What do you think about (hardcoding) packages, so we have instances `1-to-10`, `11-to-20`, ... ? This would make all learning a bit more general.

mfeurer: Hm, excellent question. First of all, the ML term instance (a data point) isn't the kind of instances to use in SMAC, instead one would like a cross-validation fold to be a SMAC instance. However, I'm not sure what validation on the training set should be for machine learning. Ideally, the number of CV folds is so low that all folds were evaluated during the optimization. Regarding the test set, all data points in the test set should be one SMAC instance, namely THE test instance. So, something like this:

* data points 1-70 are training data points and split into 3 SMAC instances
* data points 71-100 are one SMAC instance used for validation.

mlindauer:
I also tried to run this PR; again on `examples/spear_qcp`

```
$ python ../../scripts/smac-validate.py --scenario scenario.txt --trajectory smac3-output_2017-08-25_09\:42\:05_\(034034\)_run1/traj_aclib2.json --output validate_results/
INFO:smac.scenario.scenario.Scenario:Reading scenario file: scenario.txt
INFO:smac.scenario.scenario.Scenario:Output to smac3-output_2017-08-25_10:07:14_(827986)
DEBUG:smac.scenario.scenario.Scenario:Output directory does not exist! Will be created.
DEBUG:smac.scenario.scenario.Scenario:Writing scenario-file to smac3-output_2017-08-25_10:07:14_(827986)_run1/scenario.txt.
DEBUG:smac.scenario.scenario.Scenario:Scenario Options:
DEBUG:smac.scenario.scenario.Scenario:abort_on_first_run_crash = True
DEBUG:smac.scenario.scenario.Scenario:always_race_default = False
DEBUG:smac.scenario.scenario.Scenario:execdir = .
DEBUG:smac.scenario.scenario.Scenario:deterministic = False
DEBUG:smac.scenario.scenario.Scenario:intensification_percentage = 0.5
DEBUG:smac.scenario.scenario.Scenario:pcs_fn = ./target_algorithm/spear-python/spear-params-mixed.pcs
DEBUG:smac.scenario.scenario.Scenario:run_obj = runtime
DEBUG:smac.scenario.scenario.Scenario:overall_obj = PAR10
DEBUG:smac.scenario.scenario.Scenario:cost_for_crash = 2147483647.0
DEBUG:smac.scenario.scenario.Scenario:cutoff = 5.0
DEBUG:smac.scenario.scenario.Scenario:algo_runs_timelimit = inf
DEBUG:smac.scenario.scenario.Scenario:wallclock_limit = 30.0
DEBUG:smac.scenario.scenario.Scenario:ta_run_limit = inf
DEBUG:smac.scenario.scenario.Scenario:minR = 1
DEBUG:smac.scenario.scenario.Scenario:maxR = 2000
DEBUG:smac.scenario.scenario.Scenario:train_inst_fn = instances.txt
DEBUG:smac.scenario.scenario.Scenario:feature_fn = features.txt
DEBUG:smac.scenario.scenario.Scenario:output_dir = smac3-output_2017-08-25_10:07:14_(827986)
DEBUG:smac.scenario.scenario.Scenario:shared_model = False
DEBUG:smac.scenario.scenario.Scenario:initial_incumbent = DEFAULT
INFO:smac.utils.io.traj_logging.TrajLogger:No output directory for trajectory logging specified -- trajectory will not be logged.
DEBUG:smac.utils.validate.Validator:Validating configs 'def+inc' on instances 'test', repeating 1 times with 1 parallel runs on backend 'None'.
DEBUG:smac.utils.validate.Validator:Gathered 2 configurations for mode def+inc.
  File "../../scripts/smac-validate.py", line 85, in <module>
    args_.n_jobs, args_.runhistory, tae)
  File "/home/lindauer/git/SMAC3_play/smac/utils/validate.py", line 120, in validate
    runhistory=runhistory)
  File "/home/lindauer/git/SMAC3_play/smac/utils/validate.py", line 223, in get_runs
    inst_seed_config = self._process_runhistory(configs, insts, runhistory)
  File "/home/lindauer/git/SMAC3_play/smac/utils/validate.py", line 295, in _process_runhistory
    for key in runhistory.data:
AttributeError: 'ExecuteTARunOld' object has no attribute 'data'
```

Please fix.

shukon:
Fixed. Sorry, I have to remember to always check the CLIs. I tried quite a few things now, but I'm quite confident that the build-fail has nothing to do with this PR. It's the `adding some rundata to RunHistory2EPM4Cost and impute censored data`, which also fails on dev and master.

shukon:
@mlindauer may I merge this? unit tests will be refined (for codecov/patch) in a follow-up that I've almost finished, where also EPMs will be introduced. I think it would be good to merge this first.

mlindauer:
Yes, I will merge it. Thanks for providing the PR.

btw: I will open a new issue to fix a little thing that smac2-validate handles differently/better.

PR #272, opened 2017-06-27
Issue/#268 Switch to new pcs format

KEggensperger:
With this PR we would switch to the latest pcs format. I added ordinal hyperparameter values, but other than that I could not find anything that does not work using some quick checks and writing simple pcs files.

This PR can be merged as soon as the ConfigSpace is updated on pypi and we agree that it works for our optimization problems.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ scipy>=0.18.1
 six
 psutil
 pynisher>=0.4.1
-ConfigSpace>=0.3.2
+ConfigSpace>=0.3.8
 scikit-learn>=0.18.0
 typing
 pyrfr>=0.4.0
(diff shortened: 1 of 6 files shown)


mlindauer:
@mfeurer I hope there will be no issues. In theory, the new pcs should be a superset of the old pcs regarding the expressiveness.

mlindauer:
As discussed in the ConfigSpace repo, I changed all iteration over configuration to use the get() function instead of []. 
I hope we can merge this soon.

mfeurer:
I just updated the ConfigSpace on pypi to allow reading/writing of the new pcs format with Constants. Your latest issue regarding inconsistencies is not solved yet, though.

mlindauer:
@mfeurer I still think it would be great if the inconsistency could be solved at some point. but for now, let's merge this.

PR #276, opened 2017-07-04
ADD EPILS

mlindauer:
ADD a ParamILS version with sorting neighbors by acquisition function

* side effect for SMAC: incumbent won't change if a challenger has the same performance (no walk on plateaus anymore)

I would appreciate a fast review of this PR since I need it for pSMAC

### DIFF
--- a/smac/facade/epils_facade.py
+++ b/smac/facade/epils_facade.py
@@ -1,0 +1,418 @@
+import logging
+import os
+import typing
+
+import numpy as np
+
+from smac.tae.execute_ta_run import ExecuteTARun
+from smac.tae.execute_ta_run_old import ExecuteTARunOld
+from smac.tae.execute_func import ExecuteTAFuncDict
+from smac.tae.execute_ta_run import StatusType
+from smac.stats.stats import Stats
+from smac.scenario.scenario import Scenario
+from smac.runhistory.runhistory import RunHistory
+from smac.runhistory.runhistory2epm import AbstractRunHistory2EPM, \
+    RunHistory2EPM4LogCost, RunHistory2EPM4Cost
+from smac.initial_design.initial_design import InitialDesign
+from smac.initial_design.default_configuration_design import \
+    DefaultConfiguration
+from smac.initial_design.random_configuration_design import RandomConfiguration
+from smac.initial_design.multi_config_initial_design import \
+    MultiConfigInitialDesign
+from smac.intensification.intensification import Intensifier
+from smac.optimizer.epils import EPILS_Solver
+from smac.optimizer.objective import average_cost
+from smac.optimizer.acquisition import EI, LogEI, AbstractAcquisitionFunction
+from smac.optimizer.local_search import LocalSearch
+from smac.epm.rf_with_instances import RandomForestWithInstances
+from smac.epm.rfr_imputator import RFRImputator
+from smac.epm.base_epm import AbstractEPM
+from smac.utils.util_funcs import get_types
+from smac.utils.io.traj_logging import TrajLogger
+from smac.utils.constants import MAXINT
+from smac.configspace import Configuration
+
+
+__author__ = "Marius Lindauer"
+__copyright__ = "Copyright 2016, ML4AAD"
+__license__ = "3-clause BSD"
+
+
+class EPILS(object):
+
+    def __init__(self,
+                 scenario: Scenario,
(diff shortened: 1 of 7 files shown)


smac/facade/epils_facade.py
```diff
@@ -0,0 +1,418 @@
+import logging
```
mfeurer: I assume a lot of code here is copied from the SMAC-facade. Maybe we should have a file utils.py in the facade directory which contains helper functions to for example retrieve the TAE in order to avoid code duplications.

smac/intensification/intensification.py
```diff
@@ -460,10 +472,16 @@ def _compare_configs(self, incumbent: Configuration,
...
         # Line 16
         if not set(inc_runs) - set(chall_runs):
+
+            # no plateau walks
+            if chal_perf >= inc_perf:
+                self.logger.debug("Incumbent (%.4f) is better than challenger (%.4f) on %d runs." % (
+                                    inc_perf, chal_perf, len(chall_runs)))
+                return incumbent 
+            
             # Challenger is as good as incumbent
```
mfeurer: Shouldn't this now be `Challenger is better than incumbent`?

smac/facade/epils_facade.py
mlindauer: Yes, a lot of code was copied. I don't know whether we should go for a utils.py here or a parent class with some general functions. My intuition would be the latter.

However, I would prefer to do this in another PR and merge this one for now. Ok?

smac/intensification/intensification.py
mlindauer: Yes, thank you for finding this.

smac/optimizer/epils.py
```diff
@@ -0,0 +1,268 @@
...
+            X, Y = self.rh2EPM.transform(self.runhistory)
+            self.model.train(X, Y)            
+            self.acquisition_func.update(model=self.model, eta=self.runhistory.get_cost(self.incumbent))
+
+
+            if iteration == 1:
+                start_point = self.incumbent
+            else:
+                # Restart?
+                if random.random() < self.restart_prob:
```
KEggensperger: This is not seeded, right? Is that on purpose?

smac/optimizer/epils.py
```diff
@@ -0,0 +1,268 @@
...
+            The best found configuration
+        """
+        
+        self.intensifier.minR = self.fast_race_minR # be aggressive here!
+        self.intensifier.Adaptive_Capping_Slackfactor = self.fast_race_adaptive_capping_factor
+        
+        incumbent = start_point
+        # Compute the acquisition value of the incumbent
+        incumbent_array = convert_configurations_to_array([incumbent])
+        acq_val_incumbent = self.acquisition_func(incumbent_array)
```
KEggensperger: This is not used anywhere?

test/test_smbo/test_epils.py
```diff
@@ -0,0 +1,128 @@
...
+        self.assertIsInstance(epils.rng, np.random.RandomState)
+        self.assertIsInstance(epils.num_run, int)
+        epils = EPILS(self.scenario, rng=1).solver
+        rng = np.random.RandomState(1)
+        self.assertEqual(epils.num_run, 1)
+        self.assertIsInstance(epils.rng, np.random.RandomState)
+        epils = EPILS(self.scenario, rng=rng).solver
+        self.assertIsInstance(epils.num_run, int)
+        self.assertIs(epils.rng, rng)
+        # ML: I don't understand the following line and it throws an error
```
KEggensperger: If this comment is still true the unittests should fail. As they don't I assume this issue is solved and the comment can be removed.

smac/optimizer/epils.py
mlindauer: thanks. fixed.

smac/optimizer/epils.py
mlindauer: correct. fixed.

test/test_smbo/test_epils.py
mlindauer: comment removed.

smac/facade/epils_facade.py
mfeurer: Sure.

mlindauer:
@mfeurer do you think we can merge this PR, too?

mfeurer:
I think so.

PR #278, opened 2017-07-11
MAINT init now shown alongside moduledocstring, fixed some aquisition…

shukon:
Docs now showing init as well as module docstring (issue/#265), also fixed some docstrings in aquisition.

```
cd doc
make html
```
shoud suffice.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -199,6 +199,7 @@ buildapi:
 	do                   \
 	  sed -i '/Submodules/d' $$f; \
 	  sed -i '/----------/d' $$f; \
+	  sed -i 's/    :show-inheritance:/    :show-inheritance:\n    :inherited-members:/g' $$f; \
 	done;
 	@echo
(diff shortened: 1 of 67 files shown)


shukon:
@mfeurer aaah, I misunderstood the problem. Inherited members are now shown. Can you confirm this on your local build?

mfeurer:
Great, I think it now looks as expected. I will go over the docstrings to make sure they're fine (wasn't part of your task I think) as soon as I find the time to do so.

mfeurer:
I did a few changes to the docstrings in alphabetical manner until `initial_design`. Then I rebased it on master and will merge now to avoid merge conflicts in the upcoming Hackathon.

mfeurer:
@KEggensperger @mlindauer not sure if there is time to do so during the Hackathon, but it would be great if someone could rework the modules from intensification onwards (mostly fixing grammar, making sure the signatures are correct and adding links).

PR #279, opened 2017-07-12
Fix/rh overwrite

mlindauer:
required for my auto_net implementation

### DIFF
--- a/smac/runhistory/runhistory.py
+++ b/smac/runhistory/runhistory.py
@@ -51,7 +51,10 @@ class RunHistory(object):
         function to aggregate perf across instances
     '''
 
-    def __init__(self, aggregate_func):
+    def __init__(self, 
+                 aggregate_func: typing.Callable,
+                 overwrite_existings: bool=False
+                 ):
 
         # By having the data in a deterministic order we can do useful tests
         # when we serialize the data and can assume it's still in the same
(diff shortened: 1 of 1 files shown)


mfeurer:
Could you please give an intuition on why this is necessary? To re-run failed runs?

mlindauer:
I'm calling the target algorithm (i.e., param_net) multiple time -- each time it continues the optimization of the NN. Since the TAE classes adds each run to the runhistory, the current implementation saves only the first call to the TAE and silently ignores all future calls. This PR fixes this problem.

For example:
1. Call TAE with config_1 and train the NN for 5 epochs; returns loss of 0.5
2. Call TAE with config_1 and train the NN for another 5 epochs; returns loss of 0.3
3. Call TAE with config_1 and train the NN for another 5 epochs; returns loss of 0.2

The runhistory will save only the first results (0.5) and ignores the improved performance values.
However after the 3 calls, the performance of config_1 in the runhistory should be 0.2.

smac/runhistory/runhistory.py
```diff
@@ -49,9 +49,16 @@ class RunHistory(object):
...
+    overwrite_existing: bool
+        allows to overwrites old results if pairs of 
+        algorithm-instance-seed were measured  
+        multiple times
     '''
 
-    def __init__(self, aggregate_func):
+    def __init__(self, 
+                 aggregate_func: typing.Callable,
+                 overwrite_existings: bool=False
```
KEggensperger: This should be "overwrite_existing" without the 's' as in the docstring. "overwrite_existing_runs" would be even better

mlindauer: Thanks. fixed

mlindauer:
@KEggensperger I see your point. However, the rh is already a memory bottleneck and I'm not convinced that we should store learning curves / all runs in it.

Issue #220, opened 2017-04-09
Do not save external data in the runhistory?


mfeurer:
The runhistory maintains a flag for external data. By not saving external data in the runhistory, we could speed up pSMAC.

mfeurer:
Was solved via #280.

PR #280, opened 2017-07-13
FIX issue #220, do not store external data in runhistory

mfeurer:


smac/optimizer/pSMAC.py
```diff
@@ -62,7 +62,7 @@ def read(run_history: RunHistory,
                  'runs.' % difference)
 
 
-def write(run_history:RunHistory, output_directory:str, num_run:int):
+def write(run_history:RunHistory, output_directory:str):
```
mlindauer: please fix the docstring

smac/runhistory/runhistory.py
```diff
@@ -248,23 +253,27 @@ def empty(self):
...
-        configs = {id_: conf.get_dictionary()
-                   for id_, conf in self.ids_config.items()}
-
         data = [([int(k.config_id),
                   str(k.instance_id) if k.instance_id is not None else None,
                   int(k.seed)], list(v))
-                for k, v in self.data.items()]
+                for k, v in self.data.items()
+                if save_external or not self.external[k]]
+        configs_to_serialize = set([entry[0][0] for entry in data])
```
mlindauer: please rename variable to `configids_to_serialize`

smac/runhistory/runhistory.py
```diff
@@ -314,7 +323,7 @@ def update_from_json(self, fn, cs):
         """
         new_runhistory = RunHistory(self.aggregate_func)
         new_runhistory.load_json(fn, cs)
-        self.update(runhistory=new_runhistory)
+        self.update(runhistory=new_runhistory, external_data=True)
```
mlindauer: I think it should be an argument of the function whether the new data is treated as external_data

mlindauer: Thinking more about it, I even believe that treating loaded runhistories as external in the context of pSMAC is wrong. external_data is not visible in the intensification procedure but we want to have it visible there.

mfeurer: I checked the code, and the way it is treated right now is that the data used in a different pSMAC run is used in both the EPM building and intensify. My "fix" would make the data be used in the EPM building only, right? Didn't SMAC2 treat pSMAC data like this? I'm currently wondering about the implications. If intensify sees external data, it will take more time for a new configuration to become an incumbent, because the number of evaluated instance seed pairs for the current incumbent grows #workers times faster than usually, possibly making SMAC spend more time in intensify than it would usually do. Is this intended? Maybe we should run both versions against each other to see the effect of both?

mlindauer: >  My "fix" would make the data be used in the EPM building only, right? Didn't SMAC2 treat pSMAC data like this?

Yes, but this was not due to principal considerations but due to problems in the SMAC2 code base (as far as I know). 

> If intensify sees external data, it will take more time for a new configuration to become an incumbent, because the number of evaluated instance seed pairs for the current incumbent grows #workers times faster than usually, possibly making SMAC spend more time in intensify than it would usually do. Is this intended? 

Yes, on the one hand SMAC could need more time to change its incumbent. But on the other hand, the decision will be based on more data. 
Related to this: is it efficient if all pSMAC runs start with the same incumbent --- I'm trying to figure this out right now.
(I can hopefully send you some more details about pSMAC on Wednesday).

> Maybe we should run both versions against each other to see the effect of both?

Maybe yes, but this is rather low on my todo stack right now.

mfeurer: Okay, as we do not want external data to be written to disk, how about making the bool an integer (or an enum) and giving it the following values:

* 0: internal data, written to disk and used by the model and intensify
* 1: external data, not written to disk but used by the model and intensify
* 2: external data, not written to disk, only used by the model

This shouldn't take more space than the current implementation and keep the code relatively simple.

mlindauer: Agreed.

mfeurer:
I incorporated your change requests and rebased on the latest version of the development branch.

smac/runhistory/runhistory.py
```diff
@@ -77,14 +100,18 @@ def __init__(self,
         # runs_per_config is necessary for computing the moving average
         self.runs_per_config = {}  # config_id -> number of runs
 
+        # Store whether a datapoint is "external", which means it was read from
+        # a JSON file. Can be chosen to not be written to disk
+        self.external = {}  # RunKey -> bool
```
mlindauer: From Line 169, I would expect that the values are `DataOrigin` and not `bool`

smac/runhistory/runhistory.py
```diff
@@ -127,21 +151,23 @@ def add(self, config, cost, time,
...
             # overwrite capped runs with uncapped runs
-            self._add(k, v, status, external_data)
+            self._add(k, v, status, origin)
         elif status == StatusType.CAPPED and self.data[k].status == StatusType.CAPPED and cost > self.data[k].cost:
             # overwrite if censored with a larger cutoff
-            self._add(k, v, status, external_data)
+            self._add(k, v, status, origin)
 
-    def _add(self, k, v, status, external_data):
+    def _add(self, k, v, status, origin):
```
mlindauer: please add type hints here.

smac/runhistory/runhistory.py
```diff
@@ -326,18 +356,19 @@ def update_from_json(self, fn, cs):
         """
         new_runhistory = RunHistory(self.aggregate_func)
         new_runhistory.load_json(fn, cs)
-        self.update(runhistory=new_runhistory)
+        self.update(runhistory=new_runhistory, origin=DataOrigin.EXTERNAL_SAME_INSTANCES)
```
mlindauer: the origin should maybe also here an argument of the method. I don't have a specific use case, but the code would be more flexible.

smac/runhistory/runhistory.py
```diff
@@ -39,6 +40,28 @@ def default(self, obj):
...
+
+    """
+    Definition of how data in the runhistory is used.
+    
+    * ``INTERNAL``: internal data which was gathered during the current 
+      optimization run. It will be saved to disk, used for building EPMs and 
+      during intensify.
+    * ``EXTERNAL_SAME_INSTANCES``: external data, which was gathered by running
+       another program on the same instances as the current optimization run 
+       runs on (for example pSMAC). Itwill not be saved to disk, but used both
```
mlindauer: Itwill -> it will

mfeurer:
@mlindauer I just rebased the commits and included your change requests.

mlindauer:
Thanks for the fixes and the PR.

btw: I still think that rebase on shared branches is a bad idea. See e.g. https://git-scm.com/docs/git-rebase#_recovering_from_upstream_rebase

mfeurer:
I do agree that rebase is not necessarily good and should be used with caution and only when nobody else is working on the same code (which I assumed is the case here). However, the rebase outweighs the complex merge conflicts in my opinion that would otherwise have shown up (and in case this branch would not be ready to be merged could have worsened).

PR #282, opened 2017-07-25
Fix/min intensify

mlindauer:
Parameterize Intensification to control its behavior (in particular for cheap artificial functions)


mlindauer:
I discussed with @KEggensperger that we don't want to show these very specific arguments to users. So, we will hide them for now until we have a better solution than having all SMAC options in the scenario class

PR #288, opened 2017-07-27
Issue/#265 finish api doc

KEggensperger:
This PR moves existing docstring to the place they belong, replaces single quotes by double quotes and to completes type annotation where possible.

For methods without docstring I added "TODO" to remind us to write a docstring at some point in the future.


KEggensperger:
As a reference, we are using this docstring-style:

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

mlindauer:
Great. Thank you.

The rendering of the attributes list seems still to be broken. Is it working on your machine?

KEggensperger:
It is not really broken, it the only way to do it.
Do we maybe have access to the style of our documentation, so we could decrease the fontsize?

PR #291, opened 2017-07-31
SklearnWrapper for SMAC 

janvanrijn:


mfeurer:
Didn't know you're already working on this. You might be interested in this one, too: https://github.com/scikit-optimize/scikit-optimize/pull/405

mlindauer:
Please commit PRs only to "development" and never to "master"

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,324 @@
...
+        self.scorer_ = check_scoring(self.estimator, scoring=self.scoring)
+
+        X, y, groups = indexable(X, y, groups)
+        n_splits = cv.get_n_splits(X, y, groups)
+
+        base_estimator = clone(self.estimator)
+
+        cv_iter = list(cv.split(X, y, groups))
+
+        #################################### START DIFFERENCE WITH BASESEARCH_CV ####################################
```
mlindauer: the other code is original code from sklearn?
If yes, can we somehow avoid to copy code from sklearn?

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,324 @@
...
+        self.n_iter = n_iter
+
+        # sets important hyperparameters of search module
+        super(ModelBasedOptimization, self).__init__(
+              estimator=estimator, scoring=scoring, fit_params=fit_params,
+              n_jobs=1, iid=iid, refit=refit, cv=cv, verbose=verbose,
+              pre_dispatch=False, error_score=error_score,
+              return_train_score=return_train_score)
+
+    def fit(self, X, y=None, groups=None):
```
mlindauer: please add type hints for all public functions

test/test_utils/test_sklearn_wrapper.py
```diff
@@ -0,0 +1,113 @@
...
+        does not contain the right functionality yet.
+        '''
+        param_grid = {}
+
+        for hyperparameter in config_space.get_hyperparameters():
+            if isinstance(hyperparameter, CategoricalHyperparameter):
+                param_grid[hyperparameter.name] = hyperparameter.choices
+            elif isinstance(hyperparameter, UniformIntegerHyperparameter):
+                if hyperparameter.log:
+                    raise ValueError('Parameter logscale unimplemented: %s' + str(hyperparameter.__class__.__name__))
```
mlindauer: in most cases, a unittest should use "asserts" and not raised errors

test/test_utils/test_sklearn_wrapper.py
```diff
@@ -0,0 +1,113 @@
...
+        scenario = Scenario({"run_obj": "quality",  "runcount-limit": n_iter, "cs": config_space, "deterministic": "true", "memory_limit": 3072})
+
+        # To optimize, we pass the function to the SMAC-object
+        smac = SMAC(scenario=scenario, rng=random_seed,
+                    tae_runner=obj_function)
+        smac_incumbent = smac.optimize()
+        smac_inc_value = np.mean(np.array(smac.get_tae_runner().run(smac_incumbent, 1)[3]['test_scores']))
+
+        # note that multiple configurations could have led to the same
+        # SMAC always takes the most recently found # TODO: check assumption
```
mlindauer: please fix all TODOs before we merge the PR

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,324 @@
+'''
+Created on July 26, 2017
+
+@author: Jan N. van Rijn
```
mlindauer: Please drop this docstring
and add some module attributes such as
__author__ = "Jan N. van Rijn"
__license__ = "3-clause BSD"

janvanrijn: Agreed.

smac/utils/sklearn_wrapper.py
janvanrijn: Can you elaborate on this? Isn't this what the docstring is doing? 
e.g.
```
X : array-like, shape = [n_samples, n_features], etc
y : array-like, shape = [n_samples] or [n_samples, n_output], optional
groups : array-like, with shape (n_samples,), optional
```

smac/utils/sklearn_wrapper.py
janvanrijn: If I could, I obviously would have done it. It's a long sklearn function, and we need mostly the same functionality, except for the few marked lines. 

Do you have any suggestion how to avoid this?

test/test_utils/test_sklearn_wrapper.py
janvanrijn: Agreed, but this function is not a testcase. This is more a utility function, that would not have been misplaced in the main codebase. However, due to the absence of logscale support of this function, I do not find it mature enough for that. 

If something goes wrong in this function, this does not indicate that the code is broken, rather it indicates that the test case is set up wrongly. 

Your decision. Raise or assert?

smac/utils/sklearn_wrapper.py
mlindauer: in our smac repo, we are using type hints as introduced in python 3.5, see
https://www.python.org/dev/peps/pep-0484/

test/test_utils/test_sklearn_wrapper.py
janvanrijn: Basically, there are two TODO's. The other one is extending the function`_param_distributions_to_config_space` to support scipy ranges, which I consider future work. 

I removed them from the comments and raised appropriate errors.

janvanrijn:
Basically, this PR provides a scikitlearn wrapper for SMAC. people that are using the scikitlearn interfaces for GridSearch, RandomSearch etc, can now also use the same interface for using SMAC. The main advantage of this is that there is an enormous amount of toolboxes that work on the sklearn interfaces. For example (but definitely not limited to) openml-python. By having this interface we would be able to compare SMAC to the various Scikit-learn search strategies, and once Hyperband is in sklearn, also against Hyperband.

mlindauer:
@mfeurer could you please review this PR. I have not enough insides in sklearn to review this in-depth.

smac/utils/sklearn_wrapper.py
janvanrijn: I need to amend this statement, I might have found something that can avoid code duplications. Please wait with reviewing, if it works it will be quite a huge update.

janvanrijn: My intended re-factorization required a PR in sklearn, which according to @amueller has little chance of success. He recommended to copy the code, as is done in this PR. I propose we keep the PR like this and proceed with the review

mfeurer:
The tests are failing. Is this ready for review (given that is wasn't requested by @janvanrijn)?

janvanrijn:
Yes,

would be great. The pull tests fail indeed (I branched from master) this
should be something minor.

2017-08-04 5:12 GMT+02:00 Matthias Feurer <notifications@github.com>:

smac/utils/sklearn_wrapper.py
amueller: I think this is the best approach indeed.

test/test_utils/test_sklearn_wrapper.py
```diff
@@ -0,0 +1,112 @@
...
+                    tae_runner=obj_function)
+        smac_incumbent = smac.optimize()
+        smac_inc_value = np.mean(np.array(smac.get_tae_runner().run(smac_incumbent, 1)[3]['test_scores']))
+
+        # note that multiple configurations could have led to the same
+        # SMAC always takes the most recently found (in case of 1 instance problems)
+        # so we find the last idx in mbo results
+        mbo_idx = n_iter - 1 - np.argmax(mbo_wrapper.cv_results_['mean_test_score'][::-1])
+
+        self.assertAlmostEqual(mbo_wrapper.best_score_, smac_inc_value, 2) # apparently, scikit-learn search returns only 2 digits
```
amueller: what? No!

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,329 @@
...
+from smac.configspace import ConfigurationSpace
+from smac.facade.smac_facade import SMAC
+from smac.scenario.scenario import Scenario
+
+
+class ModelBasedOptimization(BaseSearchCV):
+    """
+    Scikit-Learn wrapper for SMAC
+    """
+    def __init__(self, estimator: BaseEstimator, param_distributions: Dict[str, List], n_iter: int = 10,
```
mfeurer: I would consider using a larger number as a default, such as 200.

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,329 @@
...
+                 scoring: Union[str, Callable, None] = None, fit_params: Union[Dict, None] = None, iid: bool = True,
+                 refit: bool = True, cv: Union[BaseCrossValidator, int] = None, verbose: int = 0,
+                 random_state: Union[int, np.random.RandomState] = None, error_score: Union[str, int] = 'raise',
+                 return_train_score: bool = True):
+        """Scikit-learn wrapper for Sequential Model Based Optimization (SMAC). Is useful as
+        this allows to use the various Scikit-learn interfaces, for example to OpenML.org
```
mfeurer: I know what you want to say, but this sentence makes little sense to me, maybe `This allows to use SMAC together with various tools relying on the scikit-learn API, such as OpenML.org.`

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,329 @@
...
+
+        Parameters
+        ---------
+        estimator : estimator object.
+            A object of that type is instantiated for each grid point.
+            This is assumed to implement the scikit-learn estimator interface.
+            Either estimator needs to provide a ``score`` function,
+            or ``scoring`` must be passed.
+
+        param_distributions : dict
```
mfeurer: You should consider updating the docstring here.

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,329 @@
...
+        fit_params : dict, optional
+            Undocumented function of Scikit-learn
+
+        iid : boolean, default=True
+            If True, the data is assumed to be identically distributed across
+            the folds, and the loss minimized is the total loss per sample,
+            and not the mean loss across the folds.
+
+        refit : bool
+            Undocumented function of Scikit-learn
```
mfeurer: I think you wanted to write `undocumented feature of scikit-learn`.

test/test_utils/test_sklearn_wrapper.py
```diff
@@ -0,0 +1,112 @@
...
+        classifier = DecisionTreeClassifier()
+        cs = ConfigurationSpace()
+        cs.add_hyperparameter(UniformIntegerHyperparameter('max_leaf_nodes', 4, 32, default=4))
+        cs.add_hyperparameter(UniformIntegerHyperparameter('min_samples_leaf', 1, 128, default=1))
+        cs.add_hyperparameter(UniformIntegerHyperparameter('max_depth', 1, 32, default=1))
+        cs.add_hyperparameter(CategoricalHyperparameter('criterion', ['gini', 'entropy'], default='gini'))
+
+        self._compare_with_smac(X, y, classifier, cs, random_seed=42, n_iter=5)
+
+
```
mfeurer: It would be great to have some unit tests, too; for example checking the construction of the SMAC object, the target function, the extraction of results into the CV object etc.

smac/utils/sklearn_wrapper.py
janvanrijn: Can you elaborate?

test/test_utils/test_sklearn_wrapper.py
janvanrijn: Either this is the case, or I am overlooking something here.
```
print(mbo_wrapper.best_score_)
```
Although not consequently, this gives me sometimes a 2-digit score as return, whereas it requires more. For example when you run the unit test `test_mbo_wrapper_dummy` 
Sklearn version seems to be: `scikit-learn>=0.18.0`

smac/utils/sklearn_wrapper.py
janvanrijn: Probably. If there is a SMAC default I would be happy to comply to that one? For now 200 it is.

smac/utils/sklearn_wrapper.py
janvanrijn: Perfect

smac/utils/sklearn_wrapper.py
janvanrijn: I elaborated on the sklearn classifier.

smac/utils/sklearn_wrapper.py
janvanrijn: You are right. In fact, it is Deprecated

test/test_utils/test_sklearn_wrapper.py
janvanrijn: I added unit tests for the param_dist to configspace function and _obj_function, which are the other other stand alone functions. Furthermore I added a check that requires the trajectories of SMAC and the wrapper to be completely the same (implicitly checking the extraction of trajectories). 

Looks OK?

janvanrijn:
Strange, the unit test sometimes fails because of different traces between the wrapper and original instantiation of SMAC. Given a Random Seed, is SMAC deterministic?

janvanrijn:
Moreover, this problem is hard to reproduce on my local environment, even if I also install MiniConda.

mfeurer:
No, SMAC is not deterministic by default, you can check Line 18 of Procedure 2 in the [SMAC paper](http://ml.informatik.uni-freiburg.de/papers/11-LION5-SMAC.pdf). However, there is an option to make it deterministic: [intensification_percentage](https://automl.github.io/SMAC3/stable/scenario_options.html?highlight=fraction).

janvanrijn:
What I understand from this is that it trades off the runtime of building a new forest and evaluating a few more configurations. So theoretically setting this to 1.0 would instruct SMAC to always rebuild the forest. Is this correct?

However, when I set this value to 1.0, it throws an error, stating that it should be between 0 and 1. Is this a bug? One would expect that this feature can be enforced? (And likewise, in order to instantiate ROAR, one would need to instruct SMAC to never build a Random Forest).

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+
+        fit_params : dict, optional
+            Deprecated feature of Scikit-learn
+
+        iid : boolean, default=True
+            If True, the data is assumed to be identically distributed across
+            the folds, and the loss minimized is the total loss per sample,
+            and not the mean loss across the folds.
+
+        refit : bool
```
amueller: That's not deprecated in scikit-learn? deprecated here?

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+        scoring : string, callable or None, default=None
+            A string (see model evaluation documentation) or
+            a scorer callable object / function with signature
+            ``scorer(estimator, X, y)``.
+            If ``None``, the ``score`` method of the estimator is used.
+
+        fit_params : dict, optional
+            Deprecated feature of Scikit-learn
+
+        iid : boolean, default=True
```
amueller: This one will actually be deprecated ;) But soon we'll also have https://github.com/scikit-learn/scikit-learn/pull/9599 which will make everything easier!

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+              - An iterable yielding train, test splits.
+
+            For integer/None inputs, if the estimator is a classifier and ``y`` is
+            either binary or multiclass, :class:`StratifiedKFold` is used. In all
+            other cases, :class:`KFold` is used.
+
+            Refer :ref:`User Guide <cross_validation>` for the various
+            cross-validation strategies that can be used here.
+
+        refit : boolean, default=True
```
amueller: refit is documented twice?

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+            SMAC parameter, the fraction of time to be used on intensification
+            (versus choice of next Configurations). In case of None, the SMAC
+            default will be used.
+            """
+
+        # Current implementation does not like distributions yet.
+        for param, values in param_distributions.items():
+            if not isinstance(values, list):
+                raise ValueError('Not implemented (yet): Wrapper does not work with distributions yet. Please use lists. ')
+        # For now, this is not supported
```
amueller: What does it use, then? Accuracy? one of the definitions of average accuracy?

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+        self.intensification_percentage = intensification_percentage
+
+        # sets important hyperparameters of search module
+        super(ModelBasedOptimization, self).__init__(
+              estimator=estimator, scoring=scoring, fit_params=fit_params,
+              n_jobs=1, iid=iid, refit=refit, cv=cv, verbose=verbose,
+              pre_dispatch=False, error_score=error_score,
+              return_train_score=return_train_score)
+
+    def fit(self, X: Union[List[List], np.array], y: Union[List, np.array, None] = None,
```
amueller: In scikit-learn, X could be many more things. Like a list of strings or a dataframe.

mlindauer:
@janvanrijn any progress on this PR?

mlindauer:
@mfeurer @janvanrijn could you two please decide what we will do with this PR before @janvanrijn leaves our group.

janvanrijn:
My opinion is MERGE.

2018-01-31 13:32 GMT+01:00 Marius Lindauer <notifications@github.com>:

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+    Scikit-Learn wrapper for SMAC
+    """
+
+    def __init__(self, estimator: BaseEstimator, param_distributions: Dict[str, List], n_iter: int = 200,
+                 # scoring: Union[str, Callable, None] = None, NOT SUPPORTED FOR NOW
+                 fit_params: Union[Dict, None] = None, iid: bool = True,
+                 refit: bool = True, cv: Union[BaseCrossValidator, int] = None, verbose: int = 0,
+                 random_state: Union[int, np.random.RandomState] = None, error_score: Union[str, int] = 'raise',
+                 return_train_score: bool = True, intensification_percentage: Union[float, None] = None):
+        """Scikit-learn wrapper for Sequential Model Based Optimization (SMAC). This allows to use SMAC together
```
mfeurer: `SMAC` actually stands for `SMAC: Sequential Model-based Algorithm Configuration`.

smac/utils/sklearn_wrapper.py
mfeurer: I agree with @amueller that this isn't deprecated in scikit-learn.

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+              - None, to use the default 3-fold cross validation,
+              - integer, to specify the number of folds in a `(Stratified)KFold`,
+              - An object to be used as a cross-validation generator.
+              - An iterable yielding train, test splits.
+
+            For integer/None inputs, if the estimator is a classifier and ``y`` is
+            either binary or multiclass, :class:`StratifiedKFold` is used. In all
+            other cases, :class:`KFold` is used.
+
+            Refer :ref:`User Guide <cross_validation>` for the various
```
mfeurer: Does this render correctly in our SMAC docs?

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+            Value to assign to the score if an error occurs in estimator fitting.
+            If set to 'raise', the error is raised. If a numeric value is given,
+            FitFailedWarning is raised. This parameter does not affect the refit
+            step, which will always raise the error.
+
+        return_train_score : boolean, default=True
+            If ``'False'``, the ``cv_results_`` attribute will not include training
+            scores.
+
+        intensification_percentage : float, default=None
```
mfeurer: Hm, I do suggest a different way to pass arguments to SMAC, because the user might be interested in passing other ones, too. How about a dict mapping from argument to value?

smac/utils/sklearn_wrapper.py
mfeurer: And it should actually be `np.ndarray`, `np.array` is a function.

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+        #################################### START DIFFERENCE WITH BASESEARCH_CV ####################################
+        cv_iter = list(cv.split(X, y, groups))
+        self.config_space = self._param_distributions_to_config_space(self.param_distributions)
+        self.scorers = scorers
+
+        scenario_params = {"run_obj": "quality",
+                           "runcount-limit": self.n_iter,
+                           "cs": self.config_space,
+                           "deterministic": "true",
+                           "memory_limit": 3072}
```
mfeurer: It should be possible to set these from the outside.

smac/utils/sklearn_wrapper.py
```diff
@@ -0,0 +1,392 @@
...
+    def _param_distributions_to_config_space(param_distributions: Dict[str, List]):
+        cs = ConfigurationSpace()
+
+        for param, distribution in param_distributions.items():
+            if not isinstance(distribution, list):
+                raise ValueError('Currently, only param_distributions of type list are allowed. ')
+            if all(isinstance(x, int) for x in distribution):
+                minimum = min(distribution)
+                maximum = max(distribution)
+                hyperparameter = UniformIntegerHyperparameter(param, minimum, maximum, minimum)
```
mfeurer: What happens if the list consists of three values?

mlindauer:
Since there is no progress on this PR for a long time and I reminded you two several times, I'm closing this PR now.

janvanrijn:
You can't be serious. It took more than 6 months and several reminders from my side before the first serious review came in. I haven't got time to address the requested changes, but it sounds like a waste of everyones invested time to now close it.

janvanrijn:
It's hard to define soon given my current duties, but the plan is to finish this, yes.

PR #311, opened 2017-08-26
Fix/abort on first crash

mlindauer:
FIX abort_on_first_crash option

I looked into it because of
https://github.com/automl/auto-sklearn/issues/345

However, I was not able to reproduce the issue exactly.
Nevertheless, the old solution was broken anyway because it would never return an incumbent if the frist run was a crash. I fixed this for now (maybe in a suboptimal way but I also need a solution for one of my project quickly).


mlindauer:
hmm... something is still broken... I was encountering the same issue as reported in the autosklearn issue tracker; however, so far I'm not able to reproduce it. So, please don't merge it for now.

mlindauer:
> I was encountering the same issue as reported in the autosklearn issue tracker; however, so far I'm not able to reproduce it. 

I take back what I said -- I had a typo in my scenario file. So, if @KEggensperger approve this PR, we could merge it.

Issue #310, opened 2017-08-26
Scenario options docs not rendered properly


mfeurer:
As can be seen here: https://automl.github.io/SMAC3/stable/apidoc/smac.scenario.scenario.html

Also, it would be great if the scenario object actually had a link to the scenario object: https://automl.github.io/SMAC3/stable/apidoc/smac.scenario.scenario.html

shukon:
Not sure if this is what you mean, rendering as fields now (see screenshot) and adding a link to the scenario-api in branch https://github.com/automl/SMAC3/tree/issue/%23310_scenario_options/doc

![smac3_scen_opts](https://user-images.githubusercontent.com/9417378/30004512-01376558-90d1-11e7-8001-b764719c49cb.png)

mfeurer:
Yes, that's exactly what I meant. It looks great! And I just realized that one of my links was incorrect, please excuse this.

shukon:
#313 fixed this

PR #313, opened 2017-09-04
fix Issue/#310 rendering of scenario options

shukon:
Fixing issue #310 and a typo in the installation-docs.

### DIFF
--- a/doc/options.rst
+++ b/doc/options.rst
@@ -20,7 +20,7 @@ Optional:
 
 Scenario
 ~~~~~~~~
-The `scenario-object <apidoc/smac.scenario.scenario.html>`_ is used to configure *SMAC* and can be constructed either by providing an actual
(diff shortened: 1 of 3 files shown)


doc/options.rst
```diff
@@ -20,7 +20,7 @@ Optional:
 
 Scenario
 ~~~~~~~~
-The scenario-object is used to configure *SMAC* and can be constructed either by providing an actual
+The `scenario-object <apidoc/smac.scenario.scenario.html>`_ is used to configure *SMAC* and can be constructed either by providing an actual
```
mfeurer: I think it would be better to reference the class if possible, something like:

    :class:`smac.scenario.scenario.Scenario`

as this is the entity present in SMAC (while the html is only rendered).

shukon: I was not aware of that possibility. Very neat, thank you for pointing that out.

PR #320, opened 2017-09-14
FIX #318

### LINKED ISSUE #318: `ExecuteTARunOld` does not always return a `StatusType`
If the return string of the TA includes a "wrong" status (e.g., because of a typo),
`ExecuteTARunOld` does not check this and returns a string as status type and not an instance of `StatusType`. Afterwards SMAC performs a pure random search without complaining.

TODO: the parsing+checking in `ExecuteTARunOld` has to be stricter.

mlindauer:
Proper parsing of status returned by TA.

+ some restructure for better unit tests

### DIFF
--- a/smac/tae/execute_ta_run_old.py
+++ b/smac/tae/execute_ta_run_old.py
@@ -1,6 +1,7 @@
 import sys
 from subprocess import Popen, PIPE
 
+from smac.configspace import Configuration
 from smac.tae.execute_ta_run import StatusType, ExecuteTARun
 
 __author__ = "Marius Lindauer"
@@ -18,10 +19,11 @@ class ExecuteTARunOld(ExecuteTARun):
     (SMAC < v2.10)
     """
 
-    def run(self, config, instance=None,
-            cutoff=None,
-            seed=12345,
-            instance_specific="0"
+    def run(self, config:Configuration, 
+            instance:str=None,
+            cutoff:float=None,
+            seed:int=12345,
+            instance_specific:str="0"
             ):
         """Runs target algorithm <self.ta> with configuration <config> on
         instance <instance> with instance specifics <specifics> for at most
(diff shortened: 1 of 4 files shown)


mlindauer:
@AndreBiedenkapp could you please review this PR.

smac/tae/execute_ta_run_old.py
```diff
@@ -94,16 +84,18 @@ def run(self, config, instance=None,
...
+        elif status.upper() in ["CRASHED"]:
             status = StatusType.CRASHED
-        elif status in ["ABORT"]:
+        elif status.upper() in ["ABORT"]:
             status = StatusType.ABORT
-        elif status in ["MEMOUT"]:
+        elif status.upper() in ["MEMOUT"]:
             status = StatusType.MEMOUT
+        else:
+            status = StatusType.CRASHED
```
AndreBiedenkapp: I'm not sure if we want the status to be crashed if the output of the target algorithm is unintelligible.

smac/tae/execute_ta_run_old.py
```diff
@@ -94,16 +84,18 @@ def run(self, config, instance=None,
...
 
         if status in [StatusType.CRASHED, StatusType.ABORT]:
             self.logger.warn(
```
AndreBiedenkapp: Especially if we display the  warning like that.
As a user of SMAC I'd then expect that my algorithm crashed and would not start checking for typos or something like that.

SMAC should see the status as crashed but the user needs some more information about what went wrong. Maybe we could add a logging warning for unintelligible output without introducing a new StatusType?

smac/tae/execute_ta_run_old.py
mlindauer: Thanks for pointing this out; added warn logging message.

PR #325, opened 2017-10-16
Feat names

shukon:
This PR strips whitespaces and \n-ewlines from the feature-names in the input-reader.

Does anyone object to saving feature names in scenario-object as `scenario.feature_names`?

### DIFF
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -396,8 +396,9 @@ class Scenario(object):
         # read feature file
         if self.feature_fn:
             if os.path.isfile(self.feature_fn):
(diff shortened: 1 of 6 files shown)


shukon:
From my point of view, the only possible complications arise, when modules work with members of the scenario-object (getting them using `dir(scenario)` or similar), instead of the `scenario._arguments`. For example when saving the scenario to a file (in smac, this won't happen, but I'm not sure about all the side-projects using smac).
However, there are a number of scenario-members that would create problems in such a case, so it's unlikely that it will create any new problems.
On the other hand it analysis a lot easier to be able to retrieve feature names without the feature file.

Issue #321, opened 2017-10-03
resumable optimization


zkurtz:
It would be useful to me to be able to run SMAC for say 30 minutes, consider the results, and then run it again for another 30 minutes, building on the previous results without starting over. If this is already a feature, a demo would be useful, as I've not been able to figure it out. This thread shows analogous functionality in hyperopt:  https://github.com/hyperopt/hyperopt/issues/267

AndreBiedenkapp:
This is being worked on in pull request #263 and a duplicate of issue #193.

mlindauer:
Dear Zach,

Thank your for proposing this feature.
As Andre pointed out, we have implemented this feature and it is part of the development branch of SMAC now. So, it will be part of the next SMAC release.
(Please check the documentation for an example)

However, I wonder whether this is really the feature you are looking for. 
SMAC already provides the current best configuration (so-called incumbent) at every point in time (both in `stdout` and in `smac3-output*/traj_aclib2.json`) and also shows you some statistics (e.g., the estimated cost of the incumbent). If you are interested in more information, we would be interested to learn more about it.

Best,
Marius

zkurtz:
Marius, thank you for the prompt response!  I'll take a look at what is getting pushed.  The feature I'm looking for is not simply awareness of the performance of the current smac run.  After considering the results of a smac run, I want to be able to start a new run with different hyper-hyperparameters (i.e. smac parameters, not the hyperparameters that smac is optimizing over), but I want the new run to build on all previous evaluations rather than starting from scratch.  Maybe the PR you reference above enables this.  I'll take a look.

zkurtz:
A python demo of the new functionality resulting from pull request #263 would be useful.

shukon:
@zkurtz Hi! I added an example of how to restore from within the python code in #326.
However, this feature is not really intended for changing other SMAC-parameters than *output_dir*, *wallclock_limit*, *runcount_limit* or *tuner-timeout*. Depending on what you intend to change, it might or might not break something. I can give you a more detailed answer, if you decide to share the parameters you want to adapt. However, using this option does continue the SMAC-run with the new specified scenario (not the scenario in the folder to be restored from, but the one passed via the `--scenario SCEN_PATH` commandline argument), so you might be able to use it. It's just not tested (or intended) for any other than the above mentioned parameters.
To use it via the cmd-line, change into the directory from where you previously executed SMAC and try
[...]

PR #326, opened 2017-10-18
Issue/#321 python demo state restoration

shukon:
add python demo of state restoration to branin example
clean up docs and branin example a little bit


examples/branin/restore_state.py
```diff
@@ -0,0 +1,64 @@
...
+    incumbent = trajectory[-1]["incumbent"]
+    # Because we changed the output_dir, we might want to copy the old
+    # trajectory-file (runhistory and stats will be complete)
+    new_traj_path = os.path.join(new_scenario.output_dir, "traj_aclib2.json")
+    shutil.copy(traj_path, new_traj_path)
+
+    # Now we can initialize SMAC with the recovered objects and restore the
+    # state where we left off. By providing stats and a restore_incumbent, SMAC
+    # automatically detects the intention of restoring a state.
+    smac = SMAC(scenario=new_scenario,
```
mlindauer: Does is it have to be so complicated? I expected (without actually trying it) that you can simply modify the already existing scenario object and then call the exisiting SMAC object again, e.g.

```
smac.solver.scenario.runcount_limit = 100 # overwrite limit
smac.optimize()
```

shukon: Yes, that works perfectly fine. I added it to the file, but commented it out, because I understood the request in #321 as an example to recover the SMAC-output and start validation both in the python-code. I chose the "extended" version, so that there is some example-code on how to recover all the needed objects.

mlindauer:
I would guess I should review this again?

shukon:
Oh. Yes. I added a simpler version in the comments, but since I expect users to check examples so they can generalize from them, I kept the "running" version as it was. Is that good for you?

examples/branin/cmdline_wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+crucial. The format of the results must be:
+    Result for SMAC: <STATUS>, <runtime>, <runlength>, <quality>, seed>, <instance-specifics>
+"""
+
+if __name__ == '__main__':
+    # Unused in this example:
+    # instance, instance_specific, cutoff, runlength = sys.argv[1:5]
+    seed = sys.argv[5]
+    # sys.argv[6] and sys.argv[8] are the names of the target algorithm 
+    # parameters (here: "-x1", "-x2")
```
mlindauer: Is it actually guaranteed in SMAC that first x1 and then x2 is passed to the wrapper?

doc/quickstart.rst
```diff
@@ -41,48 +42,44 @@ And run the f_min-function:
...
-use the whole of *SMACs* capability, a scenario_ object should be used. There are a
-wrapper `branin_cmdline.py`, a scenario-file `branin_scenario.txt` and a
-PCS-file `branin_pcs.pcs` in the branin-folder.
+use the whole of *SMACs* capability, a scenario_ object should be used.
+
+Commandline
+~~~~~~~~~~~
+There are a wrapper `cmdline_wrapper.py`, a scenario-file `scenario.txt` and a
+PCS-file `param_config_space.pcs` in the branin-folder.
 To run the example scenario, change into the root-directory of *SMAC* and type the following commands:
```
mfeurer: root directory

doc/quickstart.rst
```diff
@@ -41,48 +42,44 @@ And run the f_min-function:
    :lineno-match:
 
 This way, you can optimize a blackbox-function with minimal effort. However, to
-use the whole of *SMACs* capability, a scenario_ object should be used. There are a
-wrapper `branin_cmdline.py`, a scenario-file `branin_scenario.txt` and a
-PCS-file `branin_pcs.pcs` in the branin-folder.
+use the whole of *SMACs* capability, a scenario_ object should be used.
```
mfeurer: use all of *SMAC's* capabilities

doc/quickstart.rst
```diff
@@ -41,48 +42,44 @@ And run the f_min-function:
...
+
+Commandline
```
mfeurer: Command line

doc/quickstart.rst
```diff
@@ -41,48 +42,44 @@ And run the f_min-function:
...
+~~~~~~~~~~~
+There are a wrapper `cmdline_wrapper.py`, a scenario-file `scenario.txt` and a
+PCS-file `param_config_space.pcs` in the branin-folder.
```
mfeurer: I think these need to be double backticks.

doc/quickstart.rst
```diff
@@ -41,48 +42,44 @@ And run the f_min-function:
...
 The **paramfile** parameter tells *SMAC* which Parameter Configuration Space (PCS_)-file to use. This file contains a list of the algorithm's parameters, their domains and default values:
 
-    .. literalinclude:: ../examples/branin/branin_pcs.pcs
+    .. literalinclude:: ../examples/branin/param_config_space.pcs
 
     x1 and x2 are both continuous parameters. x1 can take any real value in the range [-5, 10], x2 in the range [0, 15] and both have the default value 0.
 
-The **run_obj** parameter specifies what *SMAC* is supposed to **minimize**. For this example we are optimizing the solution quality.
-For *SMAC* to be able to interpret the results of an algorithm run, a wrapper should return the results of the algorithm run as follows:
+The **run_obj** parameter specifies what *SMAC* is supposed to **minimize**. Here we optimize solution quality.
```
mfeurer: SMAC minimizes, but we optimizes quality? I think we don't yet have a solution for this weird contradiction.

doc/quickstart.rst
```diff
@@ -41,48 +42,44 @@ And run the f_min-function:
...
+
+The **runcount_limit** specifies the maximum number of algorithm calls.
+
+*SMAC* reads the results from the commandline output. The wrapper returns the
```
mfeurer: command line

doc/quickstart.rst
mfeurer: Is this really a wrapper here?

doc/tae.rst
```diff
@@ -22,6 +22,10 @@ provide slightly different interfaces. You can see `SVM-example
       <algo> <instance> <instance specific> <cutoff time> <runlength> <seed> <algorithm parameters>
       python algo.py 123 0 10 25 12345 -param1 value1 -param2 value2 [...]
 
+  The first two parameters after the ``algo.py`` hold the instance name on which the algorithm is evaluated and  extra information about the instance (rarely used).
```
mfeurer: how about `are the instance name, on which the target algorithm is evaluated, and extra information about the instance (rarely used).

doc/tae.rst
```diff
@@ -22,6 +22,10 @@ provide slightly different interfaces. You can see `SVM-example
...
+  The third parameter gives the cutoff time (maximal running time) of the target algorithm is allowed to run and the fourth the runlength (maximal number of steps).
```
mfeurer: The third parameter is the cutoff time, which is the maximal time the target algorithm is allowed to run. The fourth parameter is the runlength, which is the maximal number of steps an algorithm is allowed to run.

examples/branin/cmdline_wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+commandline-arguments.
+The target algorithm parameters (here: x1, x2) are also passed as
+commandline-arguments.
+A full call by SMAC looks like this:
+    <algo>           <instance> <instance specific> <cutoff time>  <runlength> <seed> <algorithm parameters>
+    python branin.py 0          0                   99999999999999 0           12345  -x1 0 -x2 0
+
+SMAC processes results from the commandline, therefore the print-statement is
+crucial. The format of the results must be:
+    Result for SMAC: <STATUS>, <runtime>, <runlength>, <quality>, seed>, <instance-specifics>
```
mfeurer: There is a < missing in front of *seed*.

examples/branin/restore_state.py
```diff
@@ -0,0 +1,74 @@
...
+
+from smac.facade.smac_facade import SMAC
+from smac.scenario.scenario import Scenario
+from smac.stats.stats import Stats
+from smac.runhistory.runhistory import RunHistory
+from smac.utils.io.traj_logging import TrajLogger
+
+"""
+This file runs SMAC and then restores the run with an extended computation
+budget. This will also work for runs that have crashed and are continued.
```
mfeurer: runs is a bit ambiguous as it could mean SMAC optimization run or target algorithm run. How about "*SMAC* runs "?

examples/branin/cmdline_wrapper.py
shukon: I think so, my argument:
In `execute_ta_run_old.py`, the parameters are added while iterating over the configuration.
In the ConfigSpace-package, the `__iter__`-method of `Configuration` loops over the hyperparameters of `ConfigSpace` which are stored in an `OrderedDict`, which in turn is created when reading in the PCS-file. Both, the `pcs` as well as `pcs_new` loop over the PCS-file and add the parameters in the same order.

doc/quickstart.rst
shukon: I'll go with optimize! What else would you call it, if not wrapper?

examples/branin/cmdline_wrapper.py
mfeurer: This is more or less correct. Hyperparameters keep their order when there are no conditions.

PR #332, opened 2017-10-27
Increase number of random configs for EI optimization

mfeurer:
As the sampling for random configuration is about 10 times faster (at least for Auto-sklearn) with the latest ConfigSpace, increasing this number should be a safe thing to do.


mfeurer:
The original implementation had 10000, but I'm not sure if the ConfigSpace is always fast enough to support this, so it is more or less tradeoff between the current state and what we would like to have.

AndreBiedenkapp:
That sounds reasonable to me.

Issue #240, opened 2017-05-09
SMAC alters output directory name given by the user


mfeurer:
The following code changes the output directory name given by the user:

    if self.output_dir:
        self.output_dir += "_run%d" %(run_id)

If the user provides an output directory, I think its name should not be changed.

mlindauer:
What should happen if a user runs SMAC multiple times? In worst case, parallel runs.
The output directories would be overwritten.

mfeurer:
Hm, but it's harder to retrieve in a program if the directory name is changed by SMAC... I'll think about this a little bit more

KEggensperger:
**As discussed offline:**

We introduce a flat folder hierarchy. If the user does not specifie otherwise a SMAC experiment will have the following structure:

`smac_<date>/run_<run_id>/*.json`

The user can specify a output directory, e.g. `./myExperiment` or `./myExperiment/` which results in:

`./myExperiment/run_<run_id>/*.json`

If smac is about to write to an already existing directory, smac will move the folder (without checking whether other processes still use that folder):

`./myExperiment/run_<run_id>/*.json -> ./myExperiment/run_<run_id>.OLD/*.json`

if `run_<run_id>.OLD` already exists, then another `.OLD` is appended (`run_<run_id>.OLD.OLD`)

mfeurer:
@mlindauer @KEggensperger did #334 fix this?

mlindauer:
> @mlindauer @KEggensperger did #334 fix this?

Yes

PR #334, opened 2017-10-29
new output structure

shukon:
Finally fixing issue #240.
Quoting from issue:

> We introduce a flat folder hierarchy. If the user does not specifie otherwise a SMAC experiment will have the following structure:
> 
> smac_<date>/run_<run_id>/*.json
> 
> The user can specify a output directory, e.g. ./myExperiment or ./myExperiment/ which results in:
> 
> ./myExperiment/run_<run_id>/*.json
> 
> If smac is about to write to an already existing directory, smac will move the folder (without checking whether other processes still use that folder):
> 
> ./myExperiment/run_<run_id>/*.json -> ./myExperiment/run_<run_id>.OLD/*.json
> 
> if run_<run_id>.OLD already exists, then another .OLD is appended (run_<run_id>.OLD.OLD)

Minor changes:
- stats now only saved if output-dir available
- restore-state now two functions (because of dependencies on scenario-creation)
- smac-cli slightly rearranged due to restore-state

### DIFF
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -122,6 +122,9 @@ class Scenario(object):
                 while (os.path.exists(move_to)):
                     move_to += ".OLD"
                 shutil.move(self.output_dir, move_to)
+                self.logger.warning("Output directory \"%s\" already exists! "
+                                    "Moving old folder to \"%s\".",
+                                    self.output_dir, move_to)
 
         self.write()
 
--- a/smac/scenario/scenario.py
+++ b/smac/scenario/scenario.py
@@ -8,6 +8,7 @@ import time
 import datetime
 import copy
 import typing
+import shutil
 
 from smac.utils.io.input_reader import InputReader
 from smac.utils.io.output_writer import OutputWriter
@@ -44,7 +45,11 @@ class Scenario(object):
     """
 
     def __init__(self, scenario, cmd_args: dict=None, run_id: int=1):
-        """Constructor
+        """ Creates a scenario-object. The output_dir will be
+        "output_dir/run_id/" and if that exists already, the old folder and its
+        content will be moved (without any checks on whether it's still used by
+        another process) to "output_dir/run_id.OLD". If that exists, ".OLD"s
+        will be appended until possible.
 
         Parameters
         ----------
@@ -112,6 +117,13 @@ class Scenario(object):
 
         if self.output_dir:
             self.output_dir = os.path.join(self.output_dir, "run_%d"%(run_id))
+            if os.path.exists(self.output_dir):
+                move_to = self.output_dir + ".OLD"
+                while (os.path.exists(move_to)):
+                    move_to += ".OLD"
+                shutil.move(self.output_dir, move_to)
+
+        self.write()
 
         self.logger.debug("Scenario Options:")
         for arg_name, arg_value in parsed_arguments.items():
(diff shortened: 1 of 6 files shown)


smac/smac_cli.py
```diff
@@ -113,29 +122,50 @@ def main_cli(self):
...
+            traj_fn.writelines(traj_list_aclib)
+        with open(traj_path_old, 'w') as traj_fn:
+            traj_fn.writelines(traj_list_old)
+        # read trajectory to retrieve incumbent
+        trajectory = TrajLogger.read_traj_aclib_format(fn=traj_path_aclib, cs=scen.cs)
         incumbent = trajectory[-1]["incumbent"]
-        root_logger.debug("Restored incumbent %s from %s", incumbent, traj_path)
-        # Copy traj if output_dir of specified scenario-file is different than
-        # the output_dir of the scenario-file in the folder from which to restore.
-        if scen.output_dir != InputReader().read_scenario_file(scen_path)['output_dir']:
```
mfeurer: Okay, so this behavior is removed (because it probably was unnecessary)?

smac/scenario/scenario.py
```diff
@@ -111,9 +116,17 @@ def __init__(self, scenario, cmd_args: dict=None, run_id: int=1):
         self._transform_arguments()
 
         if self.output_dir:
-            self.output_dir += "_run%d" %(run_id)
-
-        self.out_writer.write_scenario_file(self)
+            self.output_dir = os.path.join(self.output_dir, "run_%d"%(run_id))
```
mfeurer: Whitespace around operator, please.

test/test_scenario/test_scenario.py
```diff
@@ -318,5 +319,32 @@ def test_features(self):
...
 
+    def test_output_structure(self):
+        """Test whether output-dir is moved correctly."""
+        scen1 = Scenario(self.test_scenario_dict)
+        self.assertEqual(scen1.output_dir, os.path.join(
+                         self.test_scenario_dict['output_dir'], 'run_1'))
+        self.assertTrue(os.path.isdir(scen1.output_dir))
+
+        scen2 = Scenario(self.test_scenario_dict)
+        self.assertTrue(os.path.isdir(scen2.output_dir+'.OLD'))
```
mfeurer: Whitespace around operator, please.

test/test_scenario/test_scenario.py
```diff
@@ -318,5 +319,32 @@ def test_features(self):
...
+
+        scen3 = Scenario(self.test_scenario_dict)
+        self.assertTrue(os.path.isdir(scen3.output_dir+'.OLD.OLD'))
```
mfeurer: Same here.

test/test_scenario/test_scenario.py
```diff
@@ -318,5 +319,32 @@ def test_features(self):
...
+
+        scen4 = Scenario(self.test_scenario_dict, run_id=2)
+        self.assertEqual(scen4.output_dir, os.path.join(
+                         self.test_scenario_dict['output_dir'], 'run_2'))
+        self.assertTrue(os.path.isdir(scen4.output_dir))
+        self.assertFalse(os.path.isdir(scen4.output_dir+'.OLD.OLD.OLD'))
```
mfeurer: And here.

smac/smac_cli.py
shukon: because of the refactoring of the smac-cli-dependencies. now the traj is read in `restore_state_before_scen` and written in `restore_state_after_scen`. this is to make sure, that restoring a run in the same outputfolder does not cause any problems (because the outputfolders are created upon scenario-construction and old folders are moved to folder.OLD, so the paths are not valid anymore. this way, restoring a state is robust against any file- or folder-relocation).

PR #337, opened 2017-11-09
Bound number of local search runs

mfeurer:
to be at most as many runs as there are points to start
from in the runhistory.


mlindauer:
I will approve this PR after the unit tests approve it :-)

mfeurer:
Actually, instead of changing something in SMAC, I decided to fix the behavior of the ConfigSpace. In a recent change, it is now compiled which introduced this bug. Now, the ConfigSpace behaves as before, namely returning an empty list when trying to sample zero configurations and the test passes.

The reason why we want to allow sampling zero configurations is to allow the usage in a rather stateless manner.

mfeurer:
@mlindauer can I merge?

PR #339, opened 2017-11-16
Generalize acquisition function

mfeurer:
Instead of accepting a numpy array, the acquisition function is now
called with a list of configurations, giving the user more flexibility
what to do with that in the acquisition function calculation.

### DIFF
--- a/smac/optimizer/acquisition.py
+++ b/smac/optimizer/acquisition.py
@@ -1,9 +1,13 @@
 # encoding=utf8
 import abc
 import logging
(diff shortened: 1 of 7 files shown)


mfeurer:
I do agree, but I currently don't have the time to work on a solution for the full generalization. This here makes the whole thing a tiny bit better, and we can then think about the other fixes if we need them.

Apparently, I merged #337 into the wrong branch. Thus, this PR now contains the two commits form #337 as well :(

PR #346, opened 2017-11-23
FIX broken link in readme

shukon:
fixing broken link. also, @mfeurer do we drop 3.4? (i remember reading something like that, we might as well update the docs(?))


mfeurer:
Kind off. We do not run the unit tests for 3.4 any more. So yes, we should drop that part.

Issue #336, opened 2017-11-08
SMAC validation, saving runhistory


shukon:
There should be an argument as to whether we are only interested in the specified configurations that are validated (i.e. only default and incumbent/ time/ etc.) or if we want the given runhistory to be extended.

shukon:
Also, the trained model should be accessible afterwards.
The validation should be more efficient in memory and computation by being more easily reuseable.

shukon:
fixed by #348

PR #348, opened 2017-11-24
Update validation

shukon:
Fix issue #336

Validator saves epm.
Output-path now passed to validation method, validation saves only requested validated runs.
Runhistory not global in validator, but instead passed to and retrieved from validate-methods.
Tests updated to reflect the changes.

This is mainly so that we can reuse the validator and save computation when validating many runs on epm. Also, the trained epm can be retrieved from the validator.

### DIFF
--- a/smac/utils/validate.py
+++ b/smac/utils/validate.py
@@ -126,7 +126,9 @@ class Validator(object):
         Parameters
         ----------
         output_fn: str
-            path to runhistory to be saved
+            path to runhistory to be saved. if the suffix is not '.json', will
+            be interpreted as directory and filename will be
+            'validated_runhistory.json'
         config_mode: str or list<Configuration>
             string or directly a list of Configuration
             str from [def, inc, def+inc, wallclock_time, cpu_time, all]
@@ -242,7 +244,9 @@ class Validator(object):
         Parameters
         ----------
         output_fn: str
-            path to runhistory to be saved
+            path to runhistory to be saved. if the suffix is not '.json', will
+            be interpreted as directory and filename will be
+            'validated_runhistory_EPM.json'
         config_mode: str or list<Configuration>
             string or directly a list of Configuration
             str from [def, inc, def+inc, wallclock_time, cpu_time, all]
--- a/smac/utils/validate.py
+++ b/smac/utils/validate.py
@@ -9,12 +9,13 @@ import numpy as np
 
 from smac.configspace import Configuration, convert_configurations_to_array
 from smac.epm.rf_with_instances import RandomForestWithInstances
+from smac.epm.rfr_imputator import RFRImputator
 from smac.optimizer.objective import average_cost
 from smac.runhistory.runhistory import RunHistory, RunKey, StatusType
 from smac.runhistory.runhistory2epm import RunHistory2EPM4Cost
 from smac.scenario.scenario import Scenario
 from smac.stats.stats import Stats
-from smac.tae.execute_ta_run import ExecuteTARun
+from smac.tae.execute_ta_run import ExecuteTARun, StatusType
 from smac.tae.execute_ta_run_old import ExecuteTARunOld
 from smac.utils.constants import MAXINT
 from smac.utils.util_funcs import get_types
@@ -81,34 +82,50 @@ class Validator(object):
             num_run = np.random.randint(MAXINT)
             self.rng = np.random.RandomState(seed=num_run)
 
-    def _save_results(self, rh:RunHistory, output):
-        """ Helper to save results to file """
-        if output == "":
+    def _save_results(self, rh: RunHistory, output_fn, backup_fn=None):
+        """ Helper to save results to file
+
+        Parameters
+        ----------
+        rh: RunHistory
+            runhistory to save
+        output_fn: str
+            filename to save history to
+        backup_fn: str
+            if output_fn is not a file, treating as dir and append this filename
+        """
+        if output_fn == "":
             self.logger.info("No output specified, validated runhistory not saved.")
-            return rh
-        base = os.path.split(output)[0]
+            return
+        # Check if a folder or a file is specified as output
+        if not output_fn.endswith('.json'):
+            output_dir = output_fn
+            output_fn = os.path.join(output_dir, backup_fn)
+            self.logger.debug("Output is \"%s\", changing to \"%s\"!",
+                              output_dir, output_fn)
+        base = os.path.split(output_fn)[0]
         if not base == "" and not os.path.exists(base):
             self.logger.debug("Folder (\"%s\") doesn't exist, creating.", base)
             os.makedirs(base)
-        rh.save_json(output)
-        self.logger.info("Saving validation-results in %s", output)
-        return rh
+        rh.save_json(output_fn)
+        self.logger.info("Saving validation-results in %s", output_fn)
 
     def validate(self,
                  config_mode:Union[str, typing.List[Configuration]]='def',
                  instance_mode:Union[str, typing.List[str]]='test',
                  repetitions:int=1, n_jobs:int=1, backend:str='threading',
                  runhistory:RunHistory=None, tae:ExecuteTARun=None,
-                 output:str="") -> RunHistory:
+                 output_fn:str="",
+                 ) -> RunHistory:
         """
         Validate configs on instances and save result in runhistory.
 
-        Sideeffect: if output is specified, saves runhistory to specified
-        output-directory.
+        side effect: if output is specified, saves runhistory to specified
+        output directory.
 
         Parameters
         ----------
-        output: string
+        output_fn: str
             path to runhistory to be saved
         config_mode: str or list<Configuration>
             string or directly a list of Configuration
@@ -175,14 +192,8 @@ class Validator(object):
                              additional_info=result[3])
             idx += 1
 
-        if output:
-            # Check if a folder or a file is specified as output
-            if not output.endswith('.json'):
-                old = output
-                output = os.path.join(output, "validated_runhistory.json")
-                self.logger.debug("Output is \"%s\", changing to \"%s\"!", old,
-                                  output)
-            self._save_results(validated_rh, output)
+        if output_fn:
+            self._save_results(validated_rh, output_fn, backup_fn="validated_runhistory.json")
         return validated_rh
 
     def _validate_parallel(self, tae: ExecuteTARun, runs: typing.List[Run],
@@ -199,7 +210,7 @@ class Validator(object):
             [Run(config=CONFIG1,inst=INSTANCE1,seed=SEED1,inst_specs=INST_SPECIFICS1), ...]
         n_jobs: int
             number of cpus to use for validation (-1 to use all)
-        backend: string
+        backend: str
             what backend to use for parallelization
 
         Returns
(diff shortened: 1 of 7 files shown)


mlindauer:
> Validator saves epm.

What do you mean by this?
Looking at the code, I would say that the EPM is not saved to disk.

shukon:
Oh, maybe saving was not the right expression, what I meant was: keeping it as an attribute internally, so it does not need to be retrained when evaluating multiple runhistories.
Also, can be used for other things (such as importance-calculations, cost-over-time etc.) that do not involve a validator, but need the same EPM.

smac/utils/validate.py
```diff
@@ -84,35 +81,35 @@ def __init__(self, scenario: Scenario, trajectory: list, output: str,
             num_run = np.random.randint(MAXINT)
             self.rng = np.random.RandomState(seed=num_run)
 
-        self.rh = RunHistory(average_cost)  # update this rh with validation-runs
-
-    def _save_results(self, rh:RunHistory, filename:str):
+    def _save_results(self, rh:RunHistory, output):
         """ Helper to save results to file """
-        if self.output == "":
+        if output == "":
```
mfeurer: Hm, I'm not sure if output is a good name for this variable - to me this reads like that this is either the output, or whether to output something, but not where to output something.

smac/utils/validate.py
```diff
@@ -84,35 +81,35 @@ def __init__(self, scenario: Scenario, trajectory: list, output: str,
...
         return rh
 
-    def validate(self, config_mode:Union[str, typing.List[Configuration]]='def',
+    def validate(self,
+                 config_mode:Union[str, typing.List[Configuration]]='def',
                  instance_mode:Union[str, typing.List[str]]='test',
                  repetitions:int=1, n_jobs:int=1, backend:str='threading',
-                 runhistory:RunHistory=None, tae:ExecuteTARun=None):
+                 runhistory:RunHistory=None, tae:ExecuteTARun=None,
+                 output:str="") -> RunHistory:
```
mfeurer: For the future, I would suggest to format function arguments differently to make it easier to see the changes:
```python
def validate(
    self,
    arg1,
    arg2,
) -> something:
```
Adding a new argument will only change a single line then.

smac/utils/validate.py
```diff
@@ -84,35 +81,35 @@ def __init__(self, scenario: Scenario, trajectory: list, output: str,
...
         """
         Validate configs on instances and save result in runhistory.
 
+        Sideeffect: if output is specified, saves runhistory to specified
```
mfeurer: sideeffect -> side effect

smac/utils/validate.py
```diff
@@ -84,35 +81,35 @@ def __init__(self, scenario: Scenario, trajectory: list, output: str,
...
+        output-directory.
```
mfeurer: output-directory -> output directory

smac/utils/validate.py
```diff
@@ -84,35 +81,35 @@ def __init__(self, scenario: Scenario, trajectory: list, output: str,
...
+
         Parameters
         ----------
+        output: string
```
mfeurer: `output : str`

smac/utils/validate.py
```diff
@@ -84,35 +81,35 @@ def __init__(self, scenario: Scenario, trajectory: list, output: str,
...
+            path to runhistory to be saved
```
mfeurer: Is this the path to the output directory (as implied by the docstring for the argument above) or the path to the json file where the output should be written to?

smac/utils/validate.py
```diff
@@ -216,14 +216,21 @@ def _validate_parallel(self, tae: ExecuteTARun, runs: typing.List[Run],
...
+                     output="", reuse_epm=True) -> RunHistory:
         """
         Use EPM to predict costs/runtimes for unknown config/inst-pairs.
 
+        Sideeffect: if output is specified, saves runhistory to specified
+        output-directory.
+
         Parameters
         ----------
+        output: string
```
mfeurer: `output : str`

smac/utils/validate.py
```diff
@@ -216,14 +216,21 @@ def _validate_parallel(self, tae: ExecuteTARun, runs: typing.List[Run],
...
+    def validate_epm(self,
+                     config_mode:Union[str, typing.List[Configuration]]='def',
                      instance_mode:Union[str, typing.List[str]]='test',
-                     repetitions:int=1, runhistory:RunHistory=None) -> RunHistory:
+                     repetitions:int=1, runhistory:RunHistory=None,
+                     output="", reuse_epm=True) -> RunHistory:
         """
         Use EPM to predict costs/runtimes for unknown config/inst-pairs.
 
+        Sideeffect: if output is specified, saves runhistory to specified
```
mfeurer: side effect

smac/utils/validate.py
```diff
@@ -216,14 +216,21 @@ def _validate_parallel(self, tae: ExecuteTARun, runs: typing.List[Run],
...
+        output-directory.
```
mfeurer: output directory

smac/utils/validate.py
```diff
@@ -237,31 +244,35 @@ def validate_epm(self, config_mode:Union[str, typing.List[Configuration]]='def',
             number of repetitions in nondeterministic algorithms
         runhistory: RunHistory
             optional, RunHistory-object to reuse runs
+        reuse_epm: bool
+            if true (and if self.epm), reuse epm to validate runs
```
mfeurer: It would be great if you could add some backticks here to format it python-like.

smac/utils/validate.py
```diff
@@ -237,31 +244,35 @@ def validate_epm(self, config_mode:Union[str, typing.List[Configuration]]='def',
...
 
         Returns
         -------
         runhistory: RunHistory
             runhistory with predicted runs
         """
-        if not isinstance(runhistory, RunHistory):
+        if not isinstance(runhistory, RunHistory) and (self.epm == None or
```
mfeurer: Please use `self.epm is None`.

smac/utils/validate.py
```diff
@@ -237,31 +244,35 @@ def validate_epm(self, config_mode:Union[str, typing.List[Configuration]]='def',
...
+                                                       reuse_epm == False):
```
mfeurer: I also prefer `reuse_epm is False`, but that does not really matter.

smac/utils/validate.py
```diff
@@ -237,31 +244,35 @@ def validate_epm(self, config_mode:Union[str, typing.List[Configuration]]='def',
...
-                          str(X.shape), str(y.shape))
-
-        types, bounds = get_types(self.scen.cs, self.scen.feature_array)
-        model = RandomForestWithInstances(types=types,
-                                          bounds=bounds,
-                                          instance_features=self.scen.feature_array,
-                                          seed=self.rng.randint(MAXINT),
-                                          ratio_features=1.0)
-        model.train(X, y)
+        elif reuse_epm == False or self.epm == None:
```
mfeurer: same here

smac/utils/validate.py
```diff
@@ -291,12 +300,19 @@ def validate_epm(self, config_mode:Union[str, typing.List[Configuration]]='def',
                             additional_info={"additional_info":
                                 "ESTIMATED USING EPM!"})
 
-        self._save_results(self.rh_epm, 'validated_runhistory_EPM.json')
-        return self.rh_epm
+        if output:
```
mfeurer: Could this be made a method to avoid code duplication?

smac/utils/validate.py
```diff
@@ -291,12 +300,19 @@ def validate_epm(self, config_mode:Union[str, typing.List[Configuration]]='def',
...
+                output = os.path.join(output, "validated_runhistory_EPM.json")
+                self.logger.debug("Output is \"%s\", changing to \"%s\"!", old,
+                                  output)
+            self._save_results(rh_epm, output)
+        return rh_epm
 
     def get_runs(self, configs: Union[str, typing.List[Configuration]],
                  insts: Union[str, typing.List[str]], repetitions: int=1,
-                 runhistory: RunHistory=None) -> typing.List[Run]:
+                 runhistory: RunHistory=None) -> (typing.List[Run], RunHistory):
```
mfeurer: I think this should be `typing.Tuple[typing.List[Run], RunHistory]`.

test/test_utils/test_validate.py
```diff
@@ -40,40 +40,50 @@ def setUp(self):
...
-                              self.output_rh, 42)
-        validator = Validator(scen, self.trajectory,
-                              self.output_rh)
+        validator = Validator(scen, self.trajectory, 42)
+        validator = Validator(scen, self.trajectory)
 
     def test_no_output(self):
         scen = Scenario(self.scen_fn, cmd_args={'run_obj':'quality'})
-        validator = Validator(scen, self.trajectory, "")
+        validator = Validator(scen, self.trajectory)
```
mfeurer: Shouldn't there be an assert?

test/test_utils/test_validate.py
```diff
@@ -40,40 +40,50 @@ def setUp(self):
...
 
     def test_nonexisting_output(self):
         scen = Scenario(self.scen_fn, cmd_args={'run_obj':'quality'})
-        validator = Validator(scen, self.trajectory,
-                "test/test_files/validation/test/nonexisting/output")
+        validator = Validator(scen, self.trajectory)
+        validator.validate(output="test/test_files/validation/test/nonexisting/output")
```
mfeurer: Same here?

test/test_utils/test_validate.py
```diff
@@ -40,40 +40,50 @@ def setUp(self):
...
+        """ if no runhistory is passed to epm, but there was a model trained
+        before, that model should be reused! (if reuse_epm flag is set) """
+        scen = Scenario(self.scen_fn, cmd_args={'run_obj':'quality'})
+        scen.feature_array = None
+        validator = Validator(scen, self.trajectory)
+        old_rh = RunHistory(average_cost)
+        for config in [e["incumbent"] for e in self.trajectory]:
+            old_rh.add(config, 1, 1, StatusType.SUCCESS, instance_id='0',
+                       seed=127)
+        validator.validate_epm(runhistory=old_rh)
```
mfeurer: Maybe add an assert here to make it clear what you're testing for?

mfeurer:
Please ping if this is ready for another review.

smac/utils/validate.py
```diff
@@ -237,31 +244,35 @@ def validate_epm(self, config_mode:Union[str, typing.List[Configuration]]='def',
...
-        model = RandomForestWithInstances(types=types,
-                                          bounds=bounds,
-                                          instance_features=self.scen.feature_array,
-                                          seed=self.rng.randint(MAXINT),
-                                          ratio_features=1.0)
-        model.train(X, y)
+        elif reuse_epm == False or self.epm == None:
+            # Train random forest and transform training data (from given rh)
+            rh2epm = RunHistory2EPM4Cost(num_params=len(self.scen.cs.get_hyperparameters()),
+                                         scenario=self.scen, rng=self.rng)
```
mlindauer: If we have runtime data and adaptive capping was used, you should set 
`impute_censored_data=True` and pass a `imputor` instance
to handle the censored data.

shukon:
@mfeurer I think all your comments are fixed. Ready for review. Thanks a lot!!
@mlindauer can you review correct use of the imputator?

smac/utils/validate.py
mfeurer: This question is not yet addressed.

shukon: Clarified the docstrings. It's the filename. However, if the suffix is not '.json', it is interpreted as directory and an appropriate filename gets appended.

smac/utils/validate.py
```diff
@@ -81,34 +82,50 @@ def __init__(self, scenario: Scenario, trajectory: list,
...
+        """ Helper to save results to file
+
+        Parameters
+        ----------
+        rh: RunHistory
+            runhistory to save
+        output_fn: str
+            filename to save history to
+        backup_fn: str
+            if output_fn is not a file, treating as dir and append this filename
```
mfeurer: Sorry for extreme nitpicking, but this is still not clear. Also, it contradicts the code below because it does not need to be a file to not create a new directory, but a string ending on `.json`.

shukon: Don't be sorry, I should be more precise from the start. How does it look now?

PR #350, opened 2017-12-12
Fix RS

mfeurer:
The fact that we previously used the full EI optimization machinery for ROAR it could happen that a configuration from local search gets a high EI value while being not strictly random. Using only random search as an EI optimization algorithm solves this issue.

### DIFF
--- a/smac/facade/roar_facade.py
+++ b/smac/facade/roar_facade.py
@@ -5,7 +5,6 @@ import numpy as np
 
 from smac.optimizer.objective import average_cost
 from smac.optimizer.ei_optimization import RandomSearch
-from smac.optimizer.acquisition import EI
 from smac.tae.execute_ta_run import StatusType, ExecuteTARun
 from smac.stats.stats import Stats
 from smac.scenario.scenario import Scenario
@@ -13,7 +12,6 @@ from smac.runhistory.runhistory import RunHistory
 from smac.runhistory.runhistory2epm import RunHistory2EPM4Cost
 from smac.initial_design.initial_design import InitialDesign
 from smac.intensification.intensification import Intensifier
-from smac.epm.random_epm import RandomEPM
 from smac.facade.smac_facade import SMAC
 from smac.configspace import Configuration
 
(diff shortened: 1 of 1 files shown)


smac/facade/roar_facade.py
```diff
@@ -89,6 +91,20 @@ def __init__(self,
...
+        # initialize empty runhistory
+        if runhistory is None:
+            runhistory = RunHistory(aggregate_func=aggregate_func)
+        # inject aggr_func if necessary
+        if runhistory.aggregate_func is None:
+            runhistory.aggregate_func = aggregate_func
+
+        self.stats = Stats(scenario)
+        rs = RandomSearch(
+            acquisition_function=EI(model=model),
```
mlindauer: Sorry, I don't understand why we have to pass EI into ROAR. ROAR should not have an model and hence, it should not use EI at all.

mfeurer: The only real change here is that EI is now passed explicitly, before that it was passed implicitly in the super class (`SMACFacade`) and actually used during acquisition function optimization. Also, ROAR **currently does have a model**. However, I could simply change the PR to pass `None` as an acquisition function, then we could also remove the `RandomEPM` from ROAR and the whole SMAC3 code base. Sounds good?

mlindauer: sounds very good.

smac/facade/roar_facade.py
```diff
@@ -102,4 +118,5 @@ def __init__(self,
             stats=stats,
             rng=rng,
             run_id=run_id,
+            acquisition_function_optimizer=rs,
```
KEggensperger: More a conceptual remark: It is a bit counterintuitive to have an optimizer called random search, but I guess there is nothing we can change here.

mfeurer: I could rename it ;) and I will do so because it's actually confusing to pass random search to random search

smac/facade/roar_facade.py
mfeurer: I changed the ROAR-part of the code. I left the `RandomEPM` in the code base as @KEggensperger reminded me that this is good for debugging.

smac/facade/roar_facade.py
mfeurer: I tried to rename it but did not find any better name that is not extremely long or does duplicate information given in the package name.

Issue #275, opened 2017-06-30
SMAC generates unnecessary warning


mfeurer:
in case the acquisition function value becomes `NaN` for some reason:

```
/usr/local/lib/python3.5/dist-packages/smac/optimizer/acquisition.py:164: RuntimeWarning: invalid value encountered in less
  if (f < 0).any():
```
Simply replacing `NaN` by zero would alleviate this problem. 

Also, we should maybe catch these kind of warnings and only show them if the logging level is set to debug:

```
/usr/local/lib/python3.5/dist-packages/scipy/stats/_distn_infrastructure.py:876: RuntimeWarning: invalid value encountered in less_equal
  return (self.a <= x) & (x <= self.b)
```

These two seem to confuse people unfamiliar with Bayesian optimization.

mlindauer:
> in case the acquisition function value becomes NaN for some reason:

shouldn't we figure out the reason and try to fix this?
I worry a bit that there could be real issue hidden somewhere.

In principle, I agree that we should hide warnings that we cannot fix and that are acceptable for us.
In fact, we are doing this already for two warnings in the censored data imputation.

mfeurer:
This should be fixed by #275.

PR #352, opened 2017-12-14
Issue #275 nan values (SMAC generates unnecessary warning)

aaronki:
Fixing issue #275: SMAC generates unnecessary warning
Refactoring: avoid dividing by zero


mfeurer:
Please direct pull requests into the development branch. If your branch is ready for review, please choose two people and ask them to review your code (on the right hand side next to your comment).

smac/optimizer/acquisition.py
```diff
@@ -157,14 +157,20 @@ def _compute(self, X: np.ndarray, **kwargs):
...
+            z = (self.eta - m - self.par) / s
+            return (self.eta - m - self.par) * norm.cdf(z) + s * norm.pdf(z)
+
         if np.any(s == 0.0):
             # if std is zero, we have observed x on all instances
             # using a RF, std should be never exactly 0.0
-            self.logger.warn("Predicted std is 0.0 for at least one sample.")
-            f[s == 0.0] = 0.0
-
+            self.logger.warning("Predicted std is 0.0 for at least one sample.")
```
mfeurer: While this code would certainly work it looks overly complicated. How about suppressing the warning and replacing the returned value by an acquisition value of 0. This should make the code easier to understand.

aaronki: The warning was a ZeroDivision Error so this is a major issue that should not be suppressed in my opinion. The code now has the semantics that -- I guess -- were initially intended, but without the warning. Also, how a zero division error is handled by the CPU/FPU, OS and python runtime environment is specific to implementation or configuration so I want to avoid this Exception to not add potential overhead by Exception handling or other routines running. From my engineering point of view I consider it bad practice to program with Exceptions when they can be avoided. I added a comment telling what the if-branch is doing.

mfeurer: It should be specific to numpy and its configuration, right? Anyway, it looks fine to me.

Issue #292, opened 2017-08-01
pSMAC crashed at file reading because of race condition


mlindauer:
When one pSMAC run writes its runhistory json file and another run reads the same file at the same time, the reading run may fail because the json file is not valid at this moment.
(So far, I believed that this is more a theoretical problem, but I encountered that problem in practice.)

We would need a way to lock files for writing them before we can read them--- parallel reading shouldn't be a problem. 
A solution could be the following package:
https://pypi.python.org/pypi/filelock/

mfeurer:
In auto-sklearn we use the deprecated [lockfile](https://pypi.python.org/pypi/lockfile/). Maybe the packages they suggest as a replacement are worth looking into, too.

PR #353, opened 2017-12-14
Issue #292 locks for runhistory files (pSMAC crashed at file reading because of race condition)

aaronki:
Fixing issue #292: pSMAC crashed at file reading because of race condition
avoid race condition with locks


mfeurer:
Could you please describe your reasoning for adding a lock which works for a single process only and describe how this fixes the issue? In my understanding, the write operations happen in different processes, so there still could be race conditions with the current implementation.

aaronki:
As no code that showed how to reproduce the problem was provided I assumed it to be a classical threading problem (I also could not find a loop containing calls opening multiple subprocesses).
To reduce this kind of misunderstandings in the future I hope we can agree on issues being submitted with a test case or a script that shows how to reproduce the problem.

smac/runhistory/runhistory.py
```diff
@@ -83,6 +86,9 @@ class RunHistory(object):
     overwrite_existing_runs
     """
 
+    """Locks for files being written to and read from"""
+    rw_locks = {}
```
mfeurer: I'm not sure why it is necessary to keep track of the lock files.

smac/runhistory/runhistory.py
```diff
@@ -328,9 +333,20 @@ def save_json(self, fn: str="runhistory.json", save_external: bool=False):
...
 
-        with open(fn, "w") as fp:
-            json.dump({"data": data,
-                       "configs": configs}, fp, cls=EnumEncoder)
+        # add lock object for `fn` if it does not exist
+        try:
+            if fn not in RunHistory.rw_locks:
+                RunHistory.rw_locks[fn] = filelock.FileLock(fn + ".lock")
+        except Exception as e:
+            logging.error(e)
```
mfeurer: I don't think it is a good idea to catch an exception here if it is not handled. The program should rather crash if there is an issue instead of continuing in an undefined state.

smac/runhistory/runhistory.py
```diff
@@ -328,9 +333,20 @@ def save_json(self, fn: str="runhistory.json", save_external: bool=False):
...
+        # lock `fn` (writing)
+        with RunHistory.rw_locks[fn]:
+            try:
+                with open(fn, "w") as fp:
+                    json.dump({"data": data,
+                               "configs": configs}, fp, cls=EnumEncoder)
+            except Exception as e2:
+                logging.error(e2)
```
mfeurer: Same here.

aaronki:
Using file locks now for process sychronization instead of thread synchronization. Also took your comments into account, @mfeurer.

mfeurer:
@mlindauer ping

Issue #284, opened 2017-07-26
No such file or directory: 'python' when installing python 3.5.2 after uninstalling python 2.7


loshchil:
I uninstalled python 2.7 on my fresh Ubuntu 16 and then I installed python 3.5.2. 
When calling the branin example, the code fails suggesting that 'python' does not exist because it actually the case and only 'python3' exists / is callable. 

`python3 ../../scripts/smac --scenario branin_scenario.txt
INFO:smac.smac_cli.SMACCLI:SMAC call: ../../scripts/smac --scenario branin_scenario.txt
INFO:smac.scenario.scenario.Scenario:Reading scenario file: branin_scenario.txt
INFO:smac.scenario.scenario.Scenario:Output to smac3-output_2017-07-26_14:25:26_(659782)
INFO:smac.stats.stats.Stats:##########################################################
INFO:smac.stats.stats.Stats:Statistics:
INFO:smac.stats.stats.Stats:#Incumbent changed: -1
INFO:smac.stats.stats.Stats:#Target algorithm runs: 0 / 500.0
INFO:smac.stats.stats.Stats:Used wallclock time: 0.00 / inf sec
INFO:smac.stats.stats.Stats:Used target algorithm runtime: 0.00 / inf sec
INFO:smac.stats.stats.Stats:##########################################################
INFO:smac.facade.smac_facade.SMAC:Final Incumbent: None
Traceback (most recent call last):
File "../../scripts/smac", line 20, in 
smac.main_cli()
File "/home/ilya/work/SMAC3/smac/smac_cli.py", line 86, in main_cli
optimizer.optimize()
File "/home/ilya/work/SMAC3/smac/facade/smac_facade.py", line 339, in optimize
incumbent = self.solver.run()
File "/home/ilya/work/SMAC3/smac/optimizer/smbo.py", line 104, in run
self.incumbent = self.initial_design.run()
File "/home/ilya/work/SMAC3/smac/initial_design/single_config_initial_design.py", line 80, in run
instance_specific=self.scenario.instance_specific.get(rand_inst, "0"))
File "/home/ilya/work/SMAC3/smac/tae/execute_ta_run.py", line 147, in start
instance_specific=instance_specific)
File "/home/ilya/work/SMAC3/smac/tae/execute_ta_run_old.py", line 80, in run
p = Popen(cmd, shell=False, stdout=PIPE, stderr=PIPE, universal_newlines=True)
[...]

PR #355, opened 2017-12-14
Issue #284 python 2 3 problems (No such file or directory: 'python' when installing python 3.5.2 after uninstalling python 2.7)

aaronki:
Fixing issue #284: No such file or directory: 'python' when installing python 3.5.2 after uninstalling python 2.7
by explicitly calling specific python version


examples/spear_qcp/scenario.txt
```diff
@@ -1,4 +1,4 @@
-algo = python -u ./target_algorithm/scripts/SATCSSCWrapper.py --mem-limit 1024 --script ./target_algorithm/spear-python/spearCSSCWrapper.py 
+algo = python2 -u ./target_algorithm/scripts/SATCSSCWrapper.py --mem-limit 1024 --script ./target_algorithm/spear-python/spearCSSCWrapper.py
```
mfeurer: Is there a reason to specifically use `python2` here?

aaronki: This prevents the python2 script SATCSSCWrapper.py from being run with python3 when python3 is the default, i.e. on Linux /usr/bin/python is a symlink to python3. As no features that are incompatible between python2 and python3 are used in the script, this original code magically works but in the general case it does not. This is the reason I explicitly call python2.

aaronki:
@KEggensperger see my answer to mfeurer above

examples/spear_qcp/scenario.txt
mfeurer: If the script works in python3 as well, I'd rather change the script to `python3` to make us more forward-compatible.

aaronki: This means I also have to change the shebang line in SATCSCWrapper.py. I'm doing this without explicitly reviewing the code. I consider this bad practice (the reason I changed it to explicitly use python2 and not python3) but for this case I'm doing it anyway.

mfeurer: @mlindauer any issues here?

mlindauer: There is no shebang in SATCSSCWrapper.py. So, I don't understand it.

in general, we don't support Python 2 and thus, all our examples should be based on Python 3

Issue #308, opened 2017-08-21
Cooling down fraction of random configurations over time


mlindauer:
idea from @sfalkner 
instead of every second configuration being a random configuration,
cooling down the fraction of random configurations over time.

mlindauer:
The first step was done in #356, but we need some more stuff to do it right... so, not closed for now

aaronki:
@mlindauer Could you please explain what should be done so I can continue to work on the issue?

mlindauer:
not now, let's talk about that another time.

mfeurer:
@mlindauer can we close this?

mlindauer:
yes, thanks for reminding me.

PR #356, opened 2017-12-14
Issue #308 growing distance of random configurations (Cooling down fraction of random configurations over time)

aaronki:
Fixing issue 308: Cooling down fraction of random configurations over time
Added classes ChooserNoCoolDown and ChooserLinearCoolDown to cool down fraction of random configurations over time. More such classes can be added easily. Also added a command line flag for configuring which class should be used with which parameters.


smac/optimizer/ei_optimization.py
```diff
@@ -443,25 +471,35 @@ class ChallengerList(object):
...
+
+    def __next__(self):
+
+        if self._index == len(self.challengers):
+            if self._last_was_random:
+                raise StopIteration
+            else:
+                # finish with random configuration
+                return self._sample(rand=True)
+        elif not self._last_was_random and self.random_configuration_chooser.check(self._index):
```
KEggensperger: We could also directly ask the chooser whether self._index should be a random configuration or not, e.g. by setting modulus to 2? Or am I missing something?

This would make this code much cleaner.

smac/optimizer/ei_optimization.py
```diff
@@ -423,9 +425,35 @@ def _maximize(
             num_points: int
     ) -> Iterable[Tuple[float, Configuration]]:
         raise NotImplementedError()
-
-        
         
+
+class ChooserNoCoolDown(object):
```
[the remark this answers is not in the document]
aaronki: I moved the chooser classes out of ei_optimization and added an abstract base class.
This class is indeed equivalent to ChooserLinearCooldown with modulus_increment = 0. Thank you for pointing this out.

smac/optimizer/ei_optimization.py
aaronki: I wanted to stay as close as possible to the original implementation so I did not change this detail. Note the change you propose does not preserve the last configuration to be a random one.
Despite of that I refactored the code according to your suggestion.

smac/facade/roar_facade.py
```diff
@@ -70,6 +73,9 @@ def __init__(self,
             Random number generator
         run_id: int, (default: 1)
             Run ID will be used as subfolder for output_dir.
+        random_configuration_chooser
```
KEggensperger: What does the random_configuration_chooser do for ROAR? It should not be needed at all?

smac/facade/roar_facade.py
```diff
@@ -8,7 +8,8 @@
 from smac.tae.execute_ta_run import StatusType, ExecuteTARun
 from smac.stats.stats import Stats
 from smac.scenario.scenario import Scenario
-from smac.optimizer.ei_optimization import ChooserNoCoolDown, ChooserLinearCoolDown
+from smac.optimizer.random_configuration_chooser import ChooserNoCoolDown, \
```
KEggensperger: Sorry, I can't comment at the correct location, but what does the random_configuration_chooser do for ROAR? Is it really needed.

smac/optimizer/ei_optimization.py
```diff
@@ -21,24 +22,24 @@
...
+
     In order to use this class it has to be subclassed and the method
     ``_maximize`` must be implemented.
-    
+
     Parameters
     ----------
     acquisition_function : ~smac.optimizer.acquisition.AbstractAcquisitionFunction
-        
+
```
KEggensperger: I do not have a strong opinion on that, but what is the reason for these empty lines?

smac/optimizer/random_configuration_chooser.py
```diff
@@ -0,0 +1,47 @@
...
+        """Check if iteration should be followed by a random configuration"""
+        pass
+
+
+class ChooserNoCoolDown(RandomConfigurationChooser):
+
+    def __init__(self, modulus=2):
+        self.modulus = modulus
+
+    def check(self, iteration):
```
KEggensperger: Just for completeness: Can you set argument types for these methods?

smac/optimizer/smbo.py
```diff
@@ -14,7 +14,9 @@
 from smac.optimizer import pSMAC
 from smac.optimizer.acquisition import AbstractAcquisitionFunction
 from smac.optimizer.ei_optimization import InterleavedLocalAndRandomSearch, \
-    AcquisitionFunctionMaximizer, RandomSearch, ChooserNoCoolDown, ChooserLinearCoolDown
+    AcquisitionFunctionMaximizer, RandomSearch
+from smac.optimizer.random_configuration_chooser import ChooserNoCoolDown, \
+    ChooserLinearCoolDown
```
KEggensperger: Is there a reason to always implement ChooserLinearCooldown as well? It is not used by default, right?

smac/utils/io/cmd_reader.py
```diff
@@ -3,7 +3,8 @@
 import re
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, SUPPRESS
 
-from smac.optimizer.ei_optimization import ChooserNoCoolDown, ChooserLinearCoolDown
+from smac.optimizer.random_configuration_chooser import ChooserNoCoolDown, \
```
KEggensperger: I can't add a comment below, but now the default value for RandomChooser needs to be changed

smac/optimizer/random_configuration_chooser.py
KEggensperger: If we set modulus < 1 we do random search. We should check for that and indicate that ROAR is the better choice for this.

smac/optimizer/ei_optimization.py
KEggensperger: @mlindauer: See ^Is this behaviour crucial?

smac/utils/io/cmd_reader.py
aaronki: The default values for the choosers were changed in all places (modulus = 1.0 changed to modulus = 2.0).

smac/optimizer/ei_optimization.py
aaronki: This way PyCharm, fixing that

smac/optimizer/smbo.py
aaronki: It is not used by default, but can be used via a command line switch so it has to be implemented as well.

smac/optimizer/random_configuration_chooser.py
aaronki: Done setting argument types and adding a warning on modulus <= 1

smac/facade/roar_facade.py
aaronki: Removed it from ROAR

mfeurer:
Would this make the ConfigurationList obsolete? If yes, could you please remove it?

smac/utils/io/cmd_reader.py
KEggensperger: Sorry, I meant in the docstring :-)

aaronki:
@mfeurer in my opinion this does not make ChallengerList obsolete as it is used in InterleavedLocalAndRandomSearch.maximize which in turn is used by SMBO.choose_next

mfeurer:
I see, didn't realize that you implemented it like this.

mlindauer:
@aaronki What's the status of this PR? There is no progress since more than 2 weeks and the number of potential merge conflicts is increasing.

aaronki:
@mlindauer I am waiting for your suggestion on how the number of random configurations should be reduced in every SMBO iteration in the default case (linear decay, exponential, etc. ?)
As we considered this to be an important decision we agreed that I don't implement this without your recommendation on this.

mlindauer:
I thought we discussed this PR already in our last meeting and requested some further extensions.

Furthermore, the default behavior of SMAC should not changed for the moment. This is a feature that should enable us to carry out further research.

PR #359, opened 2017-12-17
FIX #357 no output-dir no directory

### LINKED ISSUE #357: SMAC always generates an output dir
SMAC generates an output directory even though I set "output_dir": ""  in the Scenario object as it is described in the code.

shukon:
fix `create_output_dir` to not create a directory if output-dir is set to "", add test and follow-up problems in trajectory logging.

### DIFF
--- a/smac/utils/io/output_directory.py
+++ b/smac/utils/io/output_directory.py
@@ -28,10 +28,7 @@ def create_output_directory(
             "run_%d" % (run_id),
         )
     else:
-        output_dir = os.path.join(
-            os.path.abspath('.'),
-            "run_%d" % (run_id),
-        )
+        return ""
     if os.path.exists(output_dir):
         move_to = output_dir + ".OLD"
         while (os.path.exists(move_to)):
--- a/smac/utils/io/traj_logging.py
+++ b/smac/utils/io/traj_logging.py
@@ -46,7 +46,8 @@ class TrajLogger(object):
         self.logger = logging.getLogger(self.__module__ + "." + self.__class__.__name__)
 
         self.output_dir = output_dir
-        if output_dir is None:
+        if output_dir is None or output_dir == "":
+            self.output_dir = None
             self.logger.info("No output directory for trajectory logging "
                              "specified -- trajectory will not be logged.")
 
--- a/test/test_facade/test_smac_facade.py
+++ b/test/test_facade/test_smac_facade.py
@@ -192,3 +192,15 @@ class TestSMACFacade(unittest.TestCase):
         shutil.rmtree(smac.output_dir + '.OLD')
         shutil.rmtree(smac.output_dir)
         shutil.rmtree(smac4.output_dir)
+
+    def test_no_output(self):
+        """ Test whether a scenario with "" as output really does not create an
+        output. """
+        test_scenario_dict = {
+            'output_dir': '',
+            'run_obj': 'quality',
+            'cs': ConfigurationSpace()
+        }
+        scen1 = Scenario(test_scenario_dict)
+        smac = SMAC(scenario=scen1, run_id=1)
+        self.assertFalse(os.path.isdir(smac.output_dir))


shukon:
Getting the following error on travis only on Python 3.5, but not on local testing. I don't understand what it has to do with my change whatsever...
The problem seems to be that there is no local search configuration in the interleaved local and random search, but why only on Python 3.5 and only in this specific test?
@mfeurer @mlindauer do you have any idea? Are there maybe correlations I don't see or do the tests simply depend on the random initialization (and should I change that)?

Patch of test-code:
```
    def test_choose_next_2(self):
        def side_effect(X):
            return np.mean(X, axis=1).reshape((-1, 1))

        smbo = SMAC(self.scenario, rng=1).solver
        smbo.incumbent = self.scenario.cs.sample_configuration()
        smbo.runhistory = RunHistory(aggregate_func=average_cost)
        smbo.runhistory.add(smbo.incumbent, 10, 10, 1)
        smbo.model = mock.Mock(spec=RandomForestWithInstances)
        smbo.acquisition_func._compute = mock.Mock(spec=RandomForestWithInstances)
        smbo.acquisition_func._compute.side_effect = side_effect

        X = smbo.rng.rand(10, 2)
        Y = smbo.rng.rand(10, 1)

        challengers = smbo.choose_next(X, Y)
        x = [c for c in challengers]

        self.assertEqual(smbo.model.train.call_count, 1)

        self.assertEqual(len(x), 10000)
        num_random_search = 0
        num_local_search = 0
        for i in range(0, 2002, 2):
            # print(x[i].origin)
            self.assertIsInstance(x[i], Configuration)
            if 'Random Search (sorted)' in x[i].origin:
                num_random_search += 1
            elif 'Local Search' in x[i].origin:
                num_local_search += 1
        # number of local search configs has to be least 10
        # since x can have duplicates
        # which can be associated with the local search
        self.assertGreaterEqual(num_local_search, 1)
        for i in range(1, 2002, 2):
            self.assertIsInstance(x[i], Configuration)
self.assertEqual(x[i].origin, 'Random Search')
```

which fails only on Python 3.5 with the following error output:

```
======================================================================
FAIL: test_choose_next_2 (test.test_smbo.test_smbo.TestSMBO)
----------------------------------------------------------------------
  File "/home/travis/build/automl/SMAC3/test/test_smbo/test_smbo.py", line 191, in test_choose_next_2
    self.assertGreaterEqual(num_local_search, 1)
AssertionError: 0 not greater than or equal to 1
-------------------- >> begin captured logging << --------------------
smac.scenario.scenario.Scenario: DEBUG: Deactivate output directory.
(...)
smac.scenario.scenario.Scenario: DEBUG: output_dir = 
smac.scenario.scenario.Scenario: INFO: No output directory for scenario logging specified -- scenario will not be logged.
smac.utils.io.traj_logging.TrajLogger: INFO: No output directory for trajectory logging specified -- trajectory will not be logged.
smac.optimizer.ei_optimization.LocalSearch: DEBUG: Switch to one of the neighbors
smac.optimizer.ei_optimization.LocalSearch: DEBUG: Switch to one of the neighbors
smac.optimizer.ei_optimization.LocalSearch: DEBUG: Switch to one of the neighbors
smac.optimizer.ei_optimization.LocalSearch: DEBUG: Local search took 4 steps and looked at 12 configurations. Computing the acquisition value for one configuration took 0.000133 seconds on average.
smac.optimizer.ei_optimization.InterleavedLocalAndRandomSearch: DEBUG: First 10 acq func (origin) values of selected configurations: [[0.99070197382287051, 'Random Search (sorted)'], [0.97762042318840825, 'Random Search (sorted)'], [0.97654304298476968, 'Random Search (sorted)'], [0.97285452615853796, 'Random Search (sorted)'], [0.97097138461054966, 'Random Search (sorted)'], [0.97096205965558235, 'Random Search (sorted)'], [0.9696651549610048, 'Random Search (sorted)'], [0.96858145542013663, 'Random Search (sorted)'], [0.96847764184001162, 'Random Search (sorted)'], [0.96819148881149641, 'Random Search (sorted)']]
```

mfeurer:
This seems to be unrelated and I observed the same issue on a PR from @aaronki . You can simply rerun your test to make this error go away. However, could you please delve a bit into this and check why this happens? (python3.5 failing hints at a dictionary which is not sorted correctly (or at all), while the order in python3.6 is deterministic).

shukon:
Ok, thanks. I opened an issue. This PR is now ready for review.

Issue #317, opened 2017-09-14
Non-Deterministic although random seed fixed


mlindauer:
SMAC is non-deterministic although it optimizes solution quality, the budget is limited by target algorithm runs and a random seed was given. In this setting, SMAC should have a deterministic behavior.

aaronki:
Created a script to check deterministic behaviour: [https://gist.github.com/aaronki/a7c06f442448e0d819696e71d8e4e4eb](url)

mfeurer:
This should be fixed now. Could you please check and close if applicable?

mfeurer:
Actually, this is not fixed - at least the unit test `test_deterministic` fails for me if I do not add `intensification_percentage` to the scenario - I assume this test fails as I'm adding plateau walking and now the optimization of the acquisition function takes longer.

PR #362, opened 2017-12-19
fixing issue #317: non-deterministic although random seed fixed

aaronki:
* passing deterministic random number generator where possible
* log if SMAC falls back to non-deterministic rngs
* remove usage of numpy.random.RandomState.seed where ~.randint should be used

Run this script in the git root of SMAC3 to check for (non-)deterministic behaviour:
[https://gist.github.com/aaronki/a7c06f442448e0d819696e71d8e4e4eb](url)


smac/facade/epils_facade.py
```diff
@@ -339,6 +338,7 @@ def _get_rng(rng):
 
         # initialize random number generator
         if rng is None:
+            self.logger.info('no rng given, falling back to non-deterministic behaviour')
```
KEggensperger: This shouldn't be a info, but rather a debugging message

smac/facade/epils_facade.py
```diff
@@ -339,6 +338,7 @@ def _get_rng(rng):
...
             num_run = np.random.randint(1234567980)
```
KEggensperger: Any reason why we use 1234567890 here and MAXINT below?
Also, isn't this function exactly the same as in smac_facade? If yes, please create a issue so we can move this in another PR.

smac/runhistory/runhistory2epm.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, scenario: Scenario, num_params: int,
 
         # Fill with some default values
         if rng is None:
-            self.rng = np.random.RandomState()
+            self.rng = np.random.RandomState(seed=1)
```
KEggensperger: Why is this a fixed seed in contrast to the other changed lines?

.travis.yml
```diff
@@ -47,6 +47,7 @@ install:
 
 # command to run tests, e.g. python setup.py test
 script:
+  - echo testing git revision $(git rev-parse HEAD) ...  # tell commit hash
```
KEggensperger: This seems to be irrelevant for the PR, right?

smac/runhistory/runhistory2epm.py
KEggensperger: @aaronki: The link to the script you provided doesn't work for me. Could you have a look at that?

KEggensperger:
@aaronki: The link to the script you provided doesn't work for me. Could you have a look at that?

.travis.yml
aaronki: Yes, it is. A similar line can be found in another PR.

smac/facade/epils_facade.py
mlindauer: I would consider this a warning that should not occur in the best case.

However, I don't get why we fall back to non-deterministic behavior. I would expect that the initialize our own rng in this case

smac/facade/epils_facade.py
mlindauer: I agree that it should be MAXINT

smac/facade/smac_facade.py
```diff
@@ -173,7 +173,7 @@ def __init__(self,
         # initialize optimizer on acquisition function
         if acquisition_function_optimizer is None:
             acquisition_function_optimizer = InterleavedLocalAndRandomSearch(
-                acquisition_function, scenario.cs
+                acquisition_function, scenario.cs, np.random.RandomState(seed=rng.randint(MAXINT))
```
mlindauer: why do we use MAXINT as seed here? I would expect a user-given seed or rng here.

smac/facade/smac_facade.py
```diff
@@ -374,7 +373,8 @@ def _get_rng(rng):
         """
         # initialize random number generator
         if rng is None:
-            num_run = np.random.randint(1234567980)
+            self.logger.info('no rng given: falling back to non-deterministic behaviour')
```
mlindauer: same comment as above. The fallback should not be non-deterministic behavior

smac/optimizer/ei_optimization.py
```diff
@@ -40,17 +41,18 @@ def __init__(
...
     ):
+        self.logger = logging.getLogger(
+            self.__module__ + "." + self.__class__.__name__
+        )
         self.acquisition_function = acquisition_function
         self.config_space = config_space
 
         if rng is None:
-            self.rng = np.random.RandomState(seed=np.random.randint(10000))
+            self.logger.info('no rng given, falling back to non-deterministic behaviour')
```
mlindauer: again, the same as above

mlindauer:
Please also address #369 with this PR.

smac/facade/smac_facade.py
aaronki: The rng is initialized with a user-given seed and so the seed generated by it is also user-dependent.

aaronki: 1234567890 was used previously, didn't change it during the refactoring. Now using MAXINT.

smac/runhistory/runhistory2epm.py
aaronki: A fixed seed is now the default behaviour when no seed was given (see comment from @mlindauer  above: »... I would expect that the initialize our own rng in this case ...«

smac/facade/epils_facade.py
aaronki: Now using MAXINT

smac/facade/epils_facade.py
aaronki: Using a default seed of 1 now.

mlindauer:
@KEggensperger Could you please also review this PR again.

smac/optimizer/ei_optimization.py
```diff
@@ -281,7 +283,9 @@ def _one_iter(
             # Get one exchange neighborhood returns an iterator (in contrast of
             # the previously returned list).
             all_neighbors = get_one_exchange_neighbourhood(
-                incumbent, seed=self.rng.seed())
+                incumbent, seed=self.rng.randint(MAXINT))
+
+            logging.warning('__test_rng__' + str(self.rng.randint(MAXINT)))
```
KEggensperger: It that output here on purpose? If yes, could you make it clear what the warning is for?

KEggensperger:
The PR looks good to me and can be merged soon when the logging output is fixed.

Also, in the meantime I wrote a script myself to verify that SMAC runs deterministically (see below) which yields the results I would expect.

```
import logging
import numpy as np

# Import ConfigSpace and different types of parameters
from smac.configspace import ConfigurationSpace
from ConfigSpace.hyperparameters import UniformFloatHyperparameter

# Import SMAC-utilities
from smac.facade.smac_facade import SMAC
from smac.facade.func_facade import fmin_smac

def rosenbrock_2d(x):
    x1 = x['x1']
    x2 = x['x2']

    val = 100. * (x2 - x1 ** 2.) ** 2. + (1 - x1) ** 2.
    return val

def rosenbrock_2d_array(x):
    return rosenbrock_2d({'x1': x[0], 'x2': x[1]})

for i in range(20):
    print()
    np.random.seed(3)
    cs = ConfigurationSpace()

    cs.add_hyperparameter(UniformFloatHyperparameter("x1", -5, 5, default_value=-3))
    cs.add_hyperparameter(UniformFloatHyperparameter("x2", -5, 5, default_value=-4))

    scenario = Scenario({"run_obj": "quality",   # we optimize quality (alternatively runtime)
                         "runcount-limit": 10,  # maximum function evaluations
                         "cs": cs,               # configuration space
                         "deterministic": "true",
                         "initial_incumbent": "DEFAULT",
                         "intensification_percentage": 0.000001
                         })

    smac = SMAC(scenario=scenario, rng=np.random.RandomState(42),
                tae_runner=rosenbrock_2d)
    incumbent = smac.optimize()

    print(rosenbrock_2d(incumbent))

    np.random.seed(3)
    x, cost, _ = fmin_smac(func=rosenbrock_2d_array,
                       x0=[-3, -4],
                       bounds=[(-5, 5), (-5, 5)],
                       maxfun=10,
                       rng=5)

    print(cost)
```

mfeurer:
Just wondering, why did `fmin_smac` become deterministic? Also, should we maybe add the `intensification_percentage` into the scenario of `fmin_smac`?

mlindauer:
> Just wondering, why did fmin_smac become deterministic?

I'm surprised about that, too.

> Also, should we maybe add the intensification_percentage into the scenario of fmin_smac?

Only if you want to ensure deterministic behavior and we don't care that intensification will only run 2 configurations in each SMBO iteration. Actually, I don't know whether we really want the latter if SMAC can effort many function evaluations.

mfeurer:
I think we don't care about the latter one - `fmin_smac` is only for optimizing deterministic functions, so we should try doing our best there.

mlindauer:
However, `fmin_smac` has a `maxtime` argument. Therefore, I would assume that we care about wallclock performance which we will neglect by setting `"intensification_percentage": 0.000001`

mfeurer:
Now I'm wondering why `maxtime` is an argument there...

I hereby propose to remove it and replace it by `intensification_percentage`. Reasoning: The function is supposed to be an easy interface to optimize simple, deterministic black-box functions, similar to `scipy.optimize`. Therefore, it would be great if the interface was as similar as possible, and also, if the optimization algorithm was deterministic.

mlindauer:
> I hereby propose to ...

agreed. but maybe in another PR!?

mfeurer:
Agreed, I'll open a new issue if @KEggensperger also agrees.

smac/optimizer/ei_optimization.py
KEggensperger: I guess this comment got lost :-)

PR #371, opened 2018-01-16
Fix log preds

mlindauer:
FIX of predictions for runtime scenario... see changelog comments


smac/facade/smac_facade.py
```diff
@@ -156,16 +156,21 @@ def __init__(self,
...
-                                              pca_components=scenario.PCA_DIM)
+                                              pca_components=scenario.PCA_DIM,
+                                              unlog_y=scenario.run_obj == "runtime")
         # initial acquisition function
         if acquisition_function is None:
-            if scenario.run_obj == "runtime":
-                acquisition_function = LogEI(model=model)
-            else:
-                acquisition_function = EI(model=model)
+            #===================================================================
```
mfeurer: Do we need to keep these comments?

mlindauer: Nope. removed :-)

KEggensperger:
Shouldn't that be merged to development first?

PR #378, opened 2018-01-19
FIX validation failure without instance features

### LINKED ISSUE #377: Validation fails if instances without features are present
As described [here](https://github.com/automl/CAVE/issues/114).

shukon:
fixes #377, adds test

### DIFF
--- a/smac/utils/validate.py
+++ b/smac/utils/validate.py
@@ -308,13 +308,14 @@ class Validator(object):
 
         # Predict desired runs
         runs, rh_epm = self.get_runs(config_mode, instance_mode, repetitions, runhistory)
-        try:
+        if hasattr(self.scen, "feature_array") and self.scen.feature_array is not None:
             feature_array_size = len(self.scen.cs.get_hyperparameters()) + self.scen.feature_array.shape[1]
-        except AttributeError:
+        else:
             feature_array_size = len(self.scen.cs.get_hyperparameters())
         X_pred = np.empty((len(runs), feature_array_size))
         for idx, run in enumerate(runs):
-            if hasattr(self.scen, "feature_dict") and run.inst != None:
+            if (hasattr(self.scen, "feature_array") and
+                    self.scen.feature_array is not None and run.inst != None):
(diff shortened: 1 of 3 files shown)


smac/utils/validate.py
```diff
@@ -308,13 +308,14 @@ def validate_epm(self,
...
-        try:
+        if hasattr(self.scen, "feature_array") and self.scen.feature_array is not None:
             feature_array_size = len(self.scen.cs.get_hyperparameters()) + self.scen.feature_array.shape[1]
-        except AttributeError:
+        else:
             feature_array_size = len(self.scen.cs.get_hyperparameters())
         X_pred = np.empty((len(runs), feature_array_size))
         for idx, run in enumerate(runs):
-            if hasattr(self.scen, "feature_dict") and run.inst != None:
+            if (hasattr(self.scen, "feature_array") and
```
mfeurer: This looks quite error-prone. Would it be possible to always assign a `feature_array` to the scenario and only check if it is `None` here?

test/test_utils/test_validate.py
```diff
@@ -335,3 +335,22 @@ def test_objective_runtime(self):
...
             old_rh.add(config, 1, 1, StatusType.SUCCESS, instance_id='0')
         validator.validate_epm('all', 'train', 1, old_rh)
+
+    def test_inst_no_feat(self):
+        ''' test if scenarios are treated correctly if no features are
+        specified.'''
+        scen = Scenario(self.scen_fn,
+                        cmd_args={'run_obj':'quality',
+                                  'instances' : self.train_insts,
+                                  'test_instances': self.test_insts})
```
mfeurer: Could you please add some assert that no features are present?

smac/utils/validate.py
shukon: Absolutely. Actually `Scenario` is already doing that and I do not remember any other reason for using `hasattr`, so I must have missed it when writing the original code.

test/test_utils/test_validate.py
shukon: Done

PR #379, opened 2018-01-22
FIX wrong argument name in script

AndreBiedenkapp:
Fixing wrong argument name


mfeurer:
Will be fixed in #379

AndreBiedenkapp:
I overlooked that, thanks

PR #381, opened 2018-01-25
Save config origin

shukon:
save config origin in runhistory.json and pretty-print json (with indents)


AndreBiedenkapp:
@mfeurer I'm not sure what you mean as the origin attribute of a configuration is saved in the runhistory,
i.e. in save_json:
https://github.com/automl/SMAC3/blob/c80a5ef8d857ad6ddb6ffebd786e6f59886fb1a9/smac/runhistory/runhistory.py#L332-L335
So it just maps from an id to the origin which then can be used to further study the behavior of SMAC in CAVE.

shukon:
@mfeurer My description was a bit sparse I think. As André said, the `origin` actually does mean saving the `Configuration.origin` in the runhistory. As it is an additional entry in the dictionary that is saved to the json-file, it does not interfere with the rest of the data. If no `origin` is set for a `Configuration`, there is not entry for it. When loading a runhistory, the `config_origin`-key is optional in the dictionary created from the json-file, so there is no problem with backwards-compatibility.

PR #382, opened 2018-01-25
ADD **kwargs in fmin facade passed to SMAC facade

mlindauer:
see title

### DIFF
--- a/smac/facade/func_facade.py
+++ b/smac/facade/func_facade.py
@@ -19,7 +19,8 @@ def fmin_smac(func: callable,
               x0: list,
               bounds: list,
               maxfun: int=-1,
-              rng: np.random.RandomState=None):
+              rng: np.random.RandomState=None,
+              **kwargs):
     """ Minimize a function func using the SMAC algorithm.
     This function is a convenience wrapper for the SMAC class.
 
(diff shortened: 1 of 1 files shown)


smac/facade/func_facade.py
```diff
@@ -36,6 +37,8 @@ def fmin_smac(func: callable,
         Maximum number of function evaluations.
     rng : np.random.RandomState, optional
             Random number generator used by SMAC.
+    **kwargs:
+        arguments passed to SMAC facade
```
mfeurer: I suggest to not use the word facade here, it's quite confusing when you just want to minimize a function. How about `Arguments passed to the optimizer class.`, ideally with a link.

mlindauer:
@mfeurer please review again

smac/facade/func_facade.py
```diff
@@ -15,31 +16,36 @@
...
+    bounds : typing.List[typing.List[float]]
         ``(min, max)`` pairs for each element in ``x``, defining the bound on
         that parameters.
     maxfun : int, optional
         Maximum number of function evaluations.
     rng : np.random.RandomState, optional
             Random number generator used by SMAC.
+    scenario_args: typing.Mapping[str,typing.Any]
+        Arguments passed to the scenario
+        See smac.scenario.scenario.Scenario
```
mfeurer: I think you need to apply some special syntax to link to the class, for example like here: https://github.com/automl/SMAC3/blob/master/smac/facade/smac_facade.py#L78

mlindauer: hmm.. I would like to avoid to link to github. I think docstrings should be independent of github.

mlindauer: Done

PR #383, opened 2018-01-29
introduce_plateaut_walking

mfeurer:


smac/optimizer/ei_optimization.py
```diff
@@ -285,19 +273,27 @@ def _one_iter(
...
+                if acq_val == acq_val_incumbent:
+                    neighbors.append(neighbor)
                 if acq_val > acq_val_incumbent + self.epsilon:
                     self.logger.debug("Switch to one of the neighbors")
                     incumbent = neighbor
                     acq_val_incumbent = acq_val
                     changed_inc = True
                     break
 
+            if not changed_inc and n_no_improvements < 10 and len(neighbors) > 0:
```
mlindauer: please replace the `10` by a constant variable with a good name.

mfeurer:
Done. I think the failing unit test is not because this PR adds a bug, but because the unit test is broken and this addition reveals it.

mfeurer:
We should probably merge this and fix the unit test in a different PR?

mlindauer:
> We should probably merge this and fix the unit test in a different PR?

I had the impression that fixing the unit test is only a single line (adding intensification percentage). So I would prefer if we don't open another PR only for a single line.

mfeurer:
Fixing this unit test is only a single line. However, the actual problem is that we didn't fix #317, although the unit test assumes it.

smac/optimizer/ei_optimization.py
```diff
@@ -259,13 +243,15 @@ def _one_iter(
             start_point: Configuration,
             *args
     ) -> Tuple[float, Configuration]:
+        N_STEPS_PLATEAU_WALK = 10
```
KEggensperger: This should not be defined in here, but somewhere else, e.g. SMAC configuration.

smac/optimizer/ei_optimization.py
```diff
@@ -195,8 +195,8 @@ def _maximize(
 
         """
 
-        num_configurations_by_local_search = self._calculate_num_points(
-            num_points, stats, runhistory
+        num_configurations_by_local_search = min(
```
KEggensperger: Disclaimer: I do not really understand why this is the same as before, but I believe this could be true.

smac/optimizer/ei_optimization.py
```diff
@@ -285,19 +271,31 @@ def _one_iter(
...
                 incumbent, seed=self.rng.randint(MAXINT))
 
+            neighbors = []
             for neighbor in all_neighbors:
                 s_time = time.time()
                 acq_val = self.acquisition_function([neighbor], *args)
                 neighbors_looked_at += 1
                 time_n.append(time.time() - s_time)
 
+                if acq_val == acq_val_incumbent:
```
KEggensperger: Do we really expect exactly the same acq value? Shouldn't this rather have a precision that defines what is means to be the same value?

KEggensperger: Also, what happens if epsilon is > 0. Don't we miss some points here?

smac/optimizer/ei_optimization.py
```diff
@@ -285,19 +271,31 @@ def _one_iter(
...
                     changed_inc = True
                     break
 
+            if (
+                not changed_inc
+                and n_no_improvements < N_STEPS_PLATEAU_WALK
+                and len(neighbors) > 0
+            ):
+                n_no_improvements += 1
+                incumbent = neighbors[0]
```
KEggensperger: Can this be something else than all_neighbors[0]?

smac/optimizer/ei_optimization.py
mfeurer: I don't get your question - but what's happening here is that I actually remove the heuristic in `_calculate_num_points`.

KEggensperger: That wasn't a question, rather a remark that it is not clear to me why this code is changed.

smac/optimizer/ei_optimization.py
mfeurer: I think equal is the definition of a plateau here @mlindauer ? Also, maybe want to drop the epsilon?

smac/optimizer/ei_optimization.py
mfeurer: Yes, only if the acquisition values are equal something is added to `neighbors`.

smac/optimizer/ei_optimization.py
```diff
@@ -161,11 +161,13 @@ def __init__(
             config_space: ConfigurationSpace,
             rng: Union[bool, np.random.RandomState] = None,
             epsilon: float=0.0,
-            max_iterations: Optional[int]=None
+            max_iterations: Optional[int]=None,
+            n_steps_plateau_walk=10,
```
KEggensperger: *ahem* could we define this to be an INT?

smac/optimizer/ei_optimization.py
mlindauer: To be consistent, we should either drop epsilon or also consider the epsilon for our plateau walks.

mfeurer:
Removed the epsilon - it should be fine now (except for the unrelated unit test which I'll fix afterwards for cleanliness of PRs).

PR #389, opened 2018-01-31
Issue 366 path restore scenario

### LINKED ISSUE #366: Path to scenario folder to restore state
Hi,

I was running the example on restoring the state for the branin function and I run into this error

```
INFO:smac.scenario.scenario.Scenario:Output to restored
Output directory= restore_me/run_1
['traj_old.csv', 'scenario.txt', 'runhistory.json', 'param_config_space.pcs', 'stats.json', 'traj_aclib2.json']
  File "restore_state.py", line 77, in <module>
    main()
  File "restore_state.py", line 64, in main
    shutil.copy(traj_path, new_traj_path)
  File "/usr/lib/python3.5/shutil.py", line 235, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.5/shutil.py", line 115, in copyfile

    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: 'restored/traj_aclib2.json'
```

I figure the issue comes from this line:
https://github.com/automl/SMAC3/blob/56a5af9e3666ae54f7c126b62ae44852af65842d/examples/branin/restore_state.py#L46

The output scenario folder should redefined as `output_dir = os.path.join(original_scenario.output_dir, 'run_1')`, right?

shukon:
Fix #366, the example was not ported to the new output-structure yet.
Also resolved some through new output-folder-creation dependencies and simplified in CLI.
Also fixed outputfolder was created twice when restored.


examples/branin/restore_state.py
```diff
@@ -23,12 +23,14 @@ def main():
...
-    smac = SMAC(scenario=original_scenario)
+    smac = SMAC(scenario=original_scenario,
+                run_id=1)
     smac.optimize()
 
     print("\n########## BUDGET EXHAUSTED! Restoring optimization: ##########\n")
 
-    # Now the output is in the folder 'restore_me'
+    # Now the output is in the folder 'restore_me/run_1'
+    old_output_dir = os.path.join(original_scenario.output_dir, 'run_1')
```
mlindauer: it doesn't has to be in `run_1`; could also be `run_20`, right?

examples/branin/restore_state.py
```diff
@@ -43,30 +45,33 @@ def main():
...
     smac = SMAC(scenario=new_scenario,
                 runhistory=runhistory,
                 stats=stats,
-                restore_incumbent=incumbent)
+                restore_incumbent=incumbent,
+                run_id=1)
+    # Because we changed the output_dir, we might want to copy the old
+    # trajectory-file (runhistory and stats will be complete, but trajectory is
+    # written sequentially)
+    new_traj_path = os.path.join(new_scenario.output_dir, "run_1",  "traj_aclib2.json")
```
mlindauer: same as above

examples/branin/restore_state.py
shukon: Yep. Added a comment.

examples/branin/restore_state.py
shukon: See above.

smac/facade/smac_facade.py
```diff
@@ -128,7 +128,14 @@ def __init__(self,
         aggregate_func = average_cost
 
         self.scenario = scenario
-        self.output_dir = create_output_directory(scenario, run_id)
+        self.output_dir = ""
+        if not restore_incumbent:
+            self.output_dir = create_output_directory(scenario, run_id)
+        elif scenario.output_dir is not None:
+            # output-dir has been created in CLI because we are restoring.
```
mfeurer: I don't understand that comment.

shukon:
@mfeurer I agree with you, I'm not happy with the structure of the restoration. Maybe you have an idea to improve it, I will explain the main issue:
Theoretically, we could simply read in all the files in the CLI and pass them all to the facade. The facade should take care of creating the output-dir and writing all the necessary files.
The problem with this is the trajectory, because it [is written incrementally](https://github.com/automl/SMAC3/blob/5775eaf234730c8aaaa06e46d0eb6e0a8348929e/smac/utils/io/traj_logging.py#L94-L98). We cannot pass it to the facade, but we want it to be in the output-dir. So the output-dir has to be created before the `SMAC`-object and the trajectory is copied there. This solution is not optimal, but I'm not sure what better way there is to do this. Do you think, the trajectory can be efficiently reconstructed from the runhistory? Open for ideas!!

smac/facade/smac_facade.py
mfeurer: Could you maybe replace the comment by something similar to your more lengthy explanation? It is easier to understand.

mfeurer:
Currently I don't have any idea except for giving SMAC a global state object which can be initialized from a directory and then can dump itself to the new directory. However, that's probably not very simple to implement.

mfeurer:
@shukon there are some conflicts now, could you please check?

shukon:
@mfeurer I will think of something more elegant for restoration, but for now I think we can merge to get the example running again.
@mlindauer Did I address your comments sufficiently?

PR #391, opened 2018-02-01
FIX #317 - make deterministic if functions is deterministic, we optim…

### LINKED ISSUE #317: Non-Deterministic although random seed fixed

mfeurer:
…ize qualitiy and the budget is limited.

Fixes #317 and simplifies the func facade.

### DIFF
--- a/smac/runhistory/runhistory.py
+++ b/smac/runhistory/runhistory.py
@@ -4,6 +4,7 @@ import filelock
 import json
 import logging
 import numpy as np
+import os
 import typing
 
 import sys
@@ -27,7 +28,6 @@ InstSeedKey = collections.namedtuple(
 
 RunValue = collections.namedtuple(
     'RunValue', ['cost', 'time', 'status', 'additional_info'])
-
 
 class EnumEncoder(json.JSONEncoder):
     """Custom encoder for enum-serialization
@@ -335,11 +335,16 @@ class RunHistory(object):
                               conf.origin is not None)}
 
         # lock `fn` (writing)
-        with filelock.FileLock(fn + ".lock"):
+        lock_file_name = fn + ".lock"
+        with filelock.SoftFileLock(lock_file_name):
             with open(fn, "w") as fp:
                 json.dump({"data": data,
                            "config_origins": config_origins,
                            "configs": configs}, fp, cls=EnumEncoder, indent=2)
+        try:
+            os.remove(lock_file_name)
+        except FileNotFoundError:
+            pass
 
     def load_json(self, fn: str, cs: ConfigurationSpace):
         """Load and runhistory in json representation from disk.
@@ -355,9 +360,14 @@ class RunHistory(object):
         """
 
         # lock `fn` (writing)
-        with filelock.FileLock(fn + ".lock"):
+        lock_file_name = fn + ".lock"
+        with filelock.FileLock(lock_file_name):
             with open(fn) as fp:
                 all_data = json.load(fp, object_hook=StatusType.enum_hook)
+        try:
+            os.remove(lock_file_name)
+        except FileNotFoundError:
+            pass
 
         config_origins = all_data.get("config_origins", {})
 
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ clean-data:
 	# remove all files that could have been left by test cases or by manual runs
 	# feel free to add more lines
 	find . -maxdepth 3 -iname 'smac3-output_*-*-*_*' | tac | while read -r TESTDIR ; do rm -Rf "$${TESTDIR}" ; done
+	find . -maxdepth 3 -iname '*.lock' -exec rm {} \;
 	rm -Rf run_*
 	rm -Rf test/test_files/scenario_test/tmp_output_*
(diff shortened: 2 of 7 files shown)


mfeurer:
@shukon I think the unit tests fail for the same reason as for #389 , do you actually know why?

shukon:
This PR fails because of left-over files:

```
ERROR: tests did not clean up all files they generated:
  test/test_files/test_json_origin.json.lock
  test/test_files/validation/validated_runhistory.json.lock
  test/test_files/validation/validated_runhistory_EPM.json.lock
```

Since this is related to validation and json_origin (both of which I worked on last), I'll have a look into it. However, #389 fails because of deterministic-tests:

```
======================================================================
FAIL: Testing deterministic behaviour.
----------------------------------------------------------------------
  File "/home/travis/build/automl/SMAC3/test/test_cli/test_deterministic_smac.py", line 64, in test_deterministic
    self.assertEqual(h1, h2)
AssertionError: {'dat[1427 chars]0], [5.113903, -1.0, {'__enum__': 'StatusType.[4944 chars]44}}} != {'dat[1427 chars]0], [0.504017, -1.0, {'__enum__': 'StatusType.[4939 chars]28}}}
Diff is 14742 characters long. Set self.maxDiff to None to see it.
-------------------- >> begin captured logging << --------------------
smac.smac_cli.SMACCLI: INFO: SMAC call: scripts/smac --scenario test/test_files/test_deterministic_scenario.txt --verbose_level DEBUG --seed 1 --output_dir test/test_files/out_test_deterministic_1
--------------------- >> end captured logging << ---------------------
```

What makes you think they are related?

shukon:
@mfeurer so from what I see in [filelock's doc](https://filelock.readthedocs.io/en/latest/), the `.lock`-file is not supposed to be deleted/cleaned up by the module, please correct me if I'm wrong. This means, we have to delete it manually so the tests don't leave any files. I just wonder, why this didn't come up before?!
`FileLock`s were introduced in #353, I'd suggest deleting the [lockfile here](https://github.com/automl/SMAC3/blob/8618d6d1814d3c979b7a02b2dcd69cf8605db4a0/smac/runhistory/runhistory.py#L338-L342), but I don't know how the filelocking works exactly and if this contradicts with pSMAC racing conditions.

mfeurer:
We introduced `FileLock`s in order to avoid race conditions with pSMAC. `FileLock` apparently builds on native Univ file manipulation libraries, and I'm not familiar with them. I would suggest that we either use the `SoftFileLock`, which only looks whether a lockfile is present, or actually try to read a file and fail gracefully (i.e. not read data from different processes) if the file is written by some other process in the meantime. From a maintenance point it might actually be easier to do the second.

aaronki:
@shukon @mfeurer I support @mfeurer  's suggestion to use SoftFileLock. See the following commit.

mfeurer:
Thanks @aaronki for the fix. Ping @KEggensperger @mlindauer

KEggensperger:
Right now (and correct me if I am wrong), if we would acquire the lock, other processes would wait forever as Softlock does not implement a timeout.

smac/runhistory/runhistory.py
```diff
@@ -335,11 +335,16 @@ def save_json(self, fn: str="runhistory.json", save_external: bool=False):
...
 
         # lock `fn` (writing)
-        with filelock.FileLock(fn + ".lock"):
+        lock_file_name = fn + ".lock"
+        with filelock.SoftFileLock(lock_file_name):
             with open(fn, "w") as fp:
                 json.dump({"data": data,
                            "config_origins": config_origins,
                            "configs": configs}, fp, cls=EnumEncoder, indent=2)
+        try:
```
KEggensperger: This should not be necessary if using a SoftFileLock

mfeurer:
Good point - should we simply use a try/except around the read operation for the json file?

KEggensperger:
That's sounds good in general.
One remark: I don't know how often pSMAC tries to read in other files, but if takes a long time to try again, we should implement that it tries several times to read in the file.

mfeurer:
It tries in each iteration - in the worst case we'd loose one or two runs. Maybe we should use an [atomic write](https://stackoverflow.com/a/18706666), i.e. write to a temporary file and then rename the file such as we do in [auto-sklearn](https://github.com/automl/auto-sklearn/blob/master/autosklearn/util/backend.py#L245).

KEggensperger:
Yes, to me that sounds like the best option where least things can go wrong. I do not see any disadvantages with this method, except that we need to make sure to delete the temporary file?

What about doing that in another PR and merge this one?

smac/runhistory/runhistory.py
aaronki: It should not be necessary, but it is. Else, some test cases get stuck. I did not investigate this issue but simply fixed it by introducing the try/catch blocks.

mfeurer:
Fine by me. @aaronki and @KEggensperger if you agree, please accept this PR.

PR #393, opened 2018-02-07
Fix Issue #374: distinguish between developer and standard options

aaronki:
Only print standard options with --help
Print developer options with --help-all
A developer option starts with '--dev' and/or has a help message starting with '[dev]'


smac/utils/io/cmd_reader.py
```diff
@@ -10,8 +10,72 @@
...
+class ConfigurableHelpFormatter(ArgumentDefaultsHelpFormatter):
+    """
+    Configurable Help Formatter. Can filter out developer options.
+    """
+
+    def __init(self, *args, **kwargs):
+        super(ConfigurableHelpFormatter, self).__init__(*args, **kwargs)
+
+    def _add_item(self, func, args):
+        def filter_actions(actions):
```
shukon: could be shortened using list-comprehensions:

```
filtered_actions = [a for a in actions if isinstance(a, Action)]
# Filter out all help strings starting with '[dev]'
filtered_actions = [a for a in filtered_actions if not (isinstance(a.help, str) and a.help.startswith('[dev]'))]
# Filter out all option strings starting with '--dev'
filtered_actions = [a for a in filtered_actions if len([s for s in a.option_strings if s.startswith('--dev')])==0]
```

but i agree that there is no need to do so except for space-saving reasons. for me it's more clear using comprehensions, but i guess that's personal choice...

smac/utils/io/cmd_reader.py
```diff
@@ -10,8 +10,72 @@
...
+
+class ConfigurableHelpFormatter(ArgumentDefaultsHelpFormatter):
+    """
+    Configurable Help Formatter. Can filter out developer options.
+    """
+
+    def __init(self, *args, **kwargs):
+        super(ConfigurableHelpFormatter, self).__init__(*args, **kwargs)
+
+    def _add_item(self, func, args):
```
mlindauer: Please add type hints

smac/utils/io/cmd_reader.py
```diff
@@ -10,8 +10,72 @@
...
+        def filter_actions(actions):
+            filtered_actions = []
+            for action in actions:
+                if isinstance(action, Action):
```
mlindauer: is this type check here necessary? 
(in general, please try to avoid type checks if possible)

smac/utils/io/cmd_reader.py
mlindauer: ... everywhere

aaronki: Added type hints.
However, _add_item is an internal method of argparse's HelpFormatter class ([https://github.com/python/cpython/blob/6ea20fc719dcaf102d2cefa1786f0421514f7a58/Lib/argparse.py#L233](https://github.com/python/cpython/blob/6ea20fc719dcaf102d2cefa1786f0421514f7a58/Lib/argparse.py#L233)). Its type interface is not defined explicitly and may change in the future.

smac/utils/io/cmd_reader.py
aaronki: It is not necessary, I removed it.

aaronki:
@mlindauer @shukon This is the first step towards seperated SMAC and scenario options. The scenario options will be considered after they have been separated into scenario and SMAC options. An option qualifies as a developer option if it a normal SMAC user does not need it. Please tell me which options fulfill this criterion and I will make the developer options in another PR.

smac/utils/io/cmd_reader.py
```diff
@@ -10,8 +11,71 @@
...
+        self._current_section.items.append((func, args))
+
+
+class StandardHelpAction(Action):
+    """Action to only show standard options in help message"""
+
+    def __init__(self, *args, **kwargs):
+        super(StandardHelpAction, self).__init__(default=SUPPRESS, nargs=0, *args, **kwargs)
+
+    def __call__(self, parser, namespace, values, option_string=None):
```
mlindauer: Please add type hints also here.

smac/utils/io/cmd_reader.py
```diff
@@ -10,8 +11,71 @@
...
+        parser.exit()
+
+
+class DevHelpAction(Action):
+    """Action to show standard and developer options in help message"""
+
+    def __init__(self, *args, **kwargs):
+        super(DevHelpAction, self).__init__(default=SUPPRESS, nargs=0, *args, **kwargs)
+
+    def __call__(self, parser, namespace, values, option_string=None):
```
mlindauer: same issue again... type hints

PR #396, opened 2018-02-09
Fix #394

### LINKED ISSUE #394: Remove filelock
At the moment we use a soft filelock to make sure two SMACs sharing a directory do not read/write at the same time. This might cause one SMAC to wait infinitely long or other race conditions.

We would prefer to solve this via ``mv``, so SMAC should first write to a temporary file and then mv that to the final destination. See also comments in: https://github.com/automl/SMAC3/pull/391

mfeurer:
By removing lockfile, see #394.

### DIFF
--- a/smac/runhistory/runhistory.py
+++ b/smac/runhistory/runhistory.py
@@ -1,13 +1,13 @@
 import collections
 from enum import Enum
 import json
-import logging
 import typing
 
 import numpy as np
 
 from smac.configspace import Configuration, ConfigurationSpace
 from smac.tae.execute_ta_run import StatusType
+from smac.utils.logging import PickableLoggerAdapter
 
 __author__ = "Marius Lindauer"
 __copyright__ = "Copyright 2015, ML4AAD"
@@ -98,7 +98,7 @@ class RunHistory(object):
             algorithm-instance-seed were measured
             multiple times
         """
-        self.logger = logging.getLogger(
+        self.logger = PickableLoggerAdapter(
             self.__module__ + "." + self.__class__.__name__
         )
 
--- a/smac/runhistory/runhistory.py
+++ b/smac/runhistory/runhistory.py
@@ -1,13 +1,9 @@
 import collections
 from enum import Enum
-import filelock
 import json
-import logging
 import numpy as np
 import os
 import typing
-
-import sys
 
 from smac.configspace import Configuration, ConfigurationSpace
 from smac.tae.execute_ta_run import StatusType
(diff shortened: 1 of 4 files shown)


mlindauer:
Could you please summarize why we first used filelock, than softfilelock and now again no lock at all? That's somehow weird.

smac/runhistory/runhistory.py
```diff
@@ -358,22 +348,26 @@ def load_json(self, fn: str, cs: ConfigurationSpace):
...
-        with filelock.FileLock(lock_file_name):
+        try:
             with open(fn) as fp:
                 all_data = json.load(fp, object_hook=StatusType.enum_hook)
-        try:
-            os.remove(lock_file_name)
-        except FileNotFoundError:
-            pass
+        except Exception as e:
+            self.logger.warning(
```
KEggensperger: I can't figure out whether this is initialized but I assume so :-)
Also, is somewhere a logging output that we successfully read in some data points?

mfeurer: Good catch, of course it wasn't ;) There's some logger output in the pSMAC part of code - so you get the output if you're using pSMAC.

mfeurer:
We first used `FileLock` because we thought locking would be the right thing to do. It turned out that it leaves behind files on the file system. Therefore, we decided to move to the `SoftFileLock`, in which case the file is the lock (and not some OS functionality is used). However, this could lead to locks which are not visible to the end user. Simply skipping a file and reading it in the next iteration seemed to be the best solution for now.

mlindauer:
> It turned out that it leaves behind files on the file system.

And what's the problem? These files are tiny (or even empty), right?
Has lockfile fulfilled its purpose and prevented raise conditions?

mfeurer:
The content of these files is potentially OS dependent. There's no general problem with these files, they're just there after the optimization, and would have to delete them.

Whether they helped against a race condition is something I don't know because I never saw one.

mlindauer:
> Whether they helped against a race condition is something I don't know because I never saw one.

I encountered this issue once:
https://github.com/automl/SMAC3/issues/292
However, this happend very rarely. So, I cannot easily reproduce the issue.

Nevertheless, I actually prefer a solution based on a well-tested package rather than a `mv`-hack.

mfeurer:
The `mv` 'hack' is only for reducing the amount of time it takes to write a json file to disk and therefore reducing the danger of ending up in the `try/except` block in the PR. I didn't implement the actual `mv` hack yet, but if you prefer to use the `FileLock` package, we could also revert to that.

PR #402, opened 2018-03-06
FIX validation of capped runs

shukon:
When reusing runhistories, validation didn't reevaluate runs that were capped, crashed or aborted.
Fixed that, see added test-case for explanation.

### DIFF
--- a/smac/utils/validate.py
+++ b/smac/utils/validate.py
@@ -413,9 +413,13 @@ class Validator(object):
                     if len(inst_seed_config[i]) == 0:
                         inst_seed_config.pop(i)
                     # Add runs to runhistory
-                    for c in configs_evaluated:
+                    for c in configs_evaluated[:]:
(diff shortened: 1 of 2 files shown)


smac/utils/validate.py
```diff
@@ -413,9 +413,13 @@ def get_runs(self, configs: Union[str, typing.List[Configuration]],
                     if len(inst_seed_config[i]) == 0:
                         inst_seed_config.pop(i)
                     # Add runs to runhistory
-                    for c in configs_evaluated:
+                    for c in configs_evaluated[:]:
```
mlindauer: why?

shukon: Configurations in `configs_evaluated` for this specific inst-seed pair contain runs that have been capped. They are removed in this for-loop, but when removing from the list we are iterating, python doesn't throw a warning but instead just skips elements. So the list is copied for iteration. I could use `tuple()` instead, so it's clear that an immutable object is desired?

Issue #405, opened 2018-04-09
Could not make output directory 


chenhuims:
Hello, I'm trying to run through the rosenbrock_2d example provided here (https://github.com/automl/SMAC3/blob/master/examples/rosenbrock.py). But I get the following error
OSError: Could not make output directory: smac3-output_2018-04-09_21:43:45_268877/run_1.
Btw, the SMAC3 package is installed on the header node of a Spark cluster. Any thoughts about how to debug this issue? 
Thank you in advance!

mlindauer:
Hi,

Thanks for reporting this issue.
Could you please provide more details?

 * Have you checked that you have write permissions in the folder where you executed SMAC?
 * We use a simple `os.makedirs()`. Could you please try to run it manually?
 * Do you start SMAC several times in parallel at the same time with the same random seed? In theory, there is a small chance (race condition) that two SMAC processes will try to create the same directory at the same time which could raise this error. If that's the case, I would recommend to use different seeds for each SMAC run.

Best,
Marius

chenhuims:
Hi Marius,
To provide a bit more information, I'm able to run `os.makedirs()` on the Spark cluster. Besides, I only run SMAC once at a time.

mlindauer:
Hi,

We changed the error handling for this exception.
Please try our current dev branch.

Cheers,
Marius

alexandertornede:
Since the output directory is now called smac3_output, this should not be an issue anymore.

PR #406, opened 2018-04-15
MAINT improve oserror-message/traceback

shukon:
regarding issue #405, I think it makes sense to provide the user with more information about `OSError`s. I found different behaviours throughout SMAC, is this one good? Mainly, do you think `sys.exit` or simply reraising `OSError`s would be more appropriate?

### DIFF
--- a/smac/utils/io/output_writer.py
+++ b/smac/utils/io/output_writer.py
@@ -40,7 +40,6 @@ class OutputWriter(object):
                                   "created.")
             try:
                 os.makedirs(scenario.output_dir_for_this_run)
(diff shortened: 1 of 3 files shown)


mlindauer:
Hi,

Thanks for the quick PR. 
Answering your question is not easy. From my experience, experienced Python users often prefer to see the original raised error (instead of hand-designed). However for other users, python errors are often cryptic. 

I know that we use `sys.exit` sometimes. In hindsight, I think this was a wrong decision and we should raise a proper exception.

Cheers,
Marius

PR #407, opened 2018-04-19
FIX ROAR

mlindauer:
* FIX cli interface of ROAR
* FIX call of acq_optimizer.maximize() in smbo
* ADD wrapper method to directly call random search

### DIFF
--- a/smac/optimizer/ei_optimization.py
+++ b/smac/optimizer/ei_optimization.py
@@ -88,7 +88,6 @@ class AcquisitionFunctionMaximizer(object, metaclass=abc.ABCMeta):
             stats: Stats,
             num_points: int,
             **kwargs
-            
     ) -> Iterable[Tuple[float, Configuration]]:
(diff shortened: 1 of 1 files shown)


smac/optimizer/ei_optimization.py
```diff
@@ -381,6 +414,31 @@ def maximize(
             random_configuration_chooser,
             *args
```
mfeurer: Shouldn't this be `**kwargs`?

smac/optimizer/ei_optimization.py
```diff
@@ -59,31 +59,36 @@ def maximize(
...
 
     @abc.abstractmethod
     def _maximize(
             self,
             runhistory: RunHistory,
             stats: Stats,
-            num_points: int
+            num_points: int,
+            **kwargs
+
```
mfeurer: There's a newline too much.

PR #408, opened 2018-04-19
Fixing bugs of func_facade

jhfjhfj1:
Change hyperparameter name format.
If the length of x0 is larger than 10.
The original name format will mess up the order of the hyperparameters.


mlindauer:
Thank you for providing this PR.

Could you please try whether this bug also still exists in our development branch? I think we fixed this bug already Jan 11th.
If it is not fixed in the dev branch, please redirect your PR to the dev branch.

Best,
Marius

PS: We will most likely update our master branch with the next SMAC release within the next 2 weeks.

PR #417, opened 2018-04-27
#395/#411 Update travis; Run mypy & examples

### LINKED ISSUE #395: Setup static code checkers
Which run on travis-ci, otherwise we're constantly having issues with code checking platforms. This should prevent issues like #388.

### LINKED ISSUE #411: Run examples in unit tests
To make sure we don't break the examples.

KEggensperger:
* restructure .travis.yml
  * runs mypy (which currently fails)
  * runs examples with every travis-ci run

Fixes #395 
Fixed #411


travis/mypyer
```diff
@@ -0,0 +1,2 @@
+MYPYPATH="smac/configspace/ smac/epm/ smac/facade/ smac/initial_design smac/intensification smac/optimizer smac/runhistory smac/scenario smac/stats smac/tae smac/utils"
```
mfeurer: Is there a reason to not do `MYPYPATH="smac" ?

travis/mypyer
```diff
@@ -0,0 +1,2 @@
...
+mypy --ignore-missing-imports $MYPYPATH
```
mfeurer: We could try to be more strict: `--disallow-any`, `--disallow-untyped-defs`, `--strict`. What do you think?

mfeurer: I'm actually not sure if we want this. Additional run info is of type `Any`, right?

KEggensperger: Done

PR #421, opened 2018-04-27
MAINT move abort-on-first-run-crash to tae

### LINKED ISSUE #410: Crash in first run makes SMAC fail
In case `AbortOnFirstRunCrash == False` as reported here: https://github.com/automl/auto-sklearn/issues/456

shukon:
Fix #410, add `abort-on-first-run-crash` argument to tae, so that initial design is not interrupted.

### DIFF
--- a/smac/tae/execute_ta_run.py
+++ b/smac/tae/execute_ta_run.py
@@ -81,7 +81,8 @@ class ExecuteTARun(object):
 
     def __init__(self, ta, stats=None, runhistory=None,
                  run_obj: str="runtime", par_factor: int=1,
-                 cost_for_crash: float=float(MAXINT)):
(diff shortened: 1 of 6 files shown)


smac/tae/execute_ta_run.py
```diff
@@ -81,7 +81,8 @@ class ExecuteTARun(object):
 
     def __init__(self, ta, stats=None, runhistory=None,
                  run_obj: str="runtime", par_factor: int=1,
-                 cost_for_crash: float=float(MAXINT)):
+                 cost_for_crash: float=float(MAXINT),
+                 abort_on_first_run_crash: bool=False):
```
mlindauer: I would expect "True" as default

Issue #120, opened 2016-11-24
WC Limit on Pynisher can be limited to uint16


mlindauer:
Depending on the OS, the wc limit of the pynisher is maybe at most uint16, e.g. on Ubuntu -- on all Linux system uint16 is guaranteed.  
The Pynisher will throw an unintuitive error message and the TA function will return None.
So, in the end, a user will not know what went wrong.

This only affects our func TAE right now, since the other TAEs do not use the pynisher.

PR #424, opened 2018-04-27
ADD cutoff limit (uint16) when using pynisher

shukon:
#120


smac/tae/execute_func.py
```diff
@@ -116,6 +116,9 @@ def run(self, config, instance=None,
         # walltime for pynisher has to be a rounded up integer
         if cutoff is not None:
             cutoff = int(math.ceil(cutoff))
+            if cutoff > 65535:
+                raise ValueError("%d is outside the legal range of [0, 65535] "
```
mlindauer: Please mention that the limit is because of OS limitations

PR #434, opened 2018-05-02
Prospector

AndreBiedenkapp:


.prospector.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+pep8:
+  disable:
+    - N806              # ignore pep8 naming (func arg should be lowercase i.e. X_pred)
+  options:
+    max-line-length: 120
+
+pylint:
+  disable:
+    - C0326              # ignore pylint bad-whitespace  (func arguments)
```
mfeurer: Is there a reason to not check this one?

AndreBiedenkapp: We have a bunch of argument in functions like https://github.com/automl/SMAC3/blob/5775eaf234730c8aaaa06e46d0eb6e0a8348929e/smac/epm/rf_with_instances.py#L117
which all would create issues.

AndreBiedenkapp: So either we replace all these function arguments that have uppercase letters with lower case ones or we ignore this one rule

mfeurer: I actually meant `C0326              # ignore pylint bad-whitespace  (func arguments)`.

AndreBiedenkapp: Oh sorry my bad. I confused myself with the func arguments.
This should be removed. I accidentally left when testing if I configure prospector correctly.

PR #436, opened 2018-05-04
Hydra-MIP

AndreBiedenkapp:
Implementation of Hydra-MIP.
Multiple parallel runs of smac -> validate all incumbents -> keep the N best incumbents.

Small speedup of Hydra:
As input decide if Hydra has to validate on the whole training set or on a small subset.

Additional modifications to our prospector setup as I found some conflicting rules that were all turned on.


smac/facade/hydra_facade.py
```diff
@@ -33,56 +86,115 @@ class Hydra(object):
...
+            valX => train_set * 100/X
+        incs_per_round: int
+            Number of incumbents to keep per round
+        n_optimizers: int
+            Number of optimizers to run in parallel per round
+        rng: int/np.random.RandomState
+            The randomState/seed to pass to each smac run
+        run_id: int
+            run_id for this hydra run
+        tae: ExecuteTARun
```
mlindauer: can we also provide a function handle here? similar as in the SMAC facade?

smac/facade/hydra_facade.py
```diff
@@ -33,56 +86,115 @@ class Hydra(object):
...
         ----------
         scenario : ~smac.scenario.scenario.Scenario
             Scenario object
         n_iterations: int,
             number of Hydra iterations
-        """
+        val_set: str
+            Set to validate incumbent(s) on. [train, valX].
+            train => whole training set,
+            valX => train_set * 100/X
```
mlindauer: You should mention that X has to be between 0 and 100

smac/facade/hydra_facade.py
```diff
@@ -33,56 +86,115 @@ class Hydra(object):
...
+            self.logger.warning('Can not determine validation set size. Using full training-set!')
+            return self.scenario.train_insts
+        else:
+            size = int(val_set[3:])/100
+            if size < 0 or size > 1:
+                raise ValueError('X too large in valX')
+            insts = np.array(self.scenario.train_insts)
+            # just to make sure this also works with the small example we have to round up to 3
+            size = max(np.floor(insts.shape[0] * size).astype(int), 3)
+            ids = np.random.choice(insts.shape[0], size, replace=False)
```
mlindauer: why do you use `np.random.choice`? Since these are strings, `random.choice` should also do the trick.

smac/facade/hydra_facade.py
```diff
@@ -130,20 +252,124 @@ def optimize(self):
...
+
+    def _update_portfolio(self, incs: np.ndarray) -> typing.Union[np.float, float]:
+        """
+        Validates all configurations (in incs) and determines which ones to add to the portfolio
+
+        Parameters
+        ----------
+        incs: np.ndarray
+            List of Configurations
+
```
mlindauer: ADD "Returns"

smac/facade/hydra_facade.py
```diff
@@ -130,20 +252,124 @@ def optimize(self):
...
+                    raise ValueError('Num validated Instances mismatch!')
+                for key in cost_per_inst:
+                    self.cost_per_inst[key] = min(self.cost_per_inst[key], cost_per_inst[key])
+            else:
+                self.cost_per_inst = cost_per_inst
+
+        cur_cost = np.mean(list(self.cost_per_inst.values()))  # type: np.float
+        return cur_cost
+
+    def _get_rng(
```
mlindauer: looks like a function that should not live in Hydra exclusively.

smac/facade/hydra_facade.py
```diff
@@ -130,20 +252,124 @@ def optimize(self):
...
+        self.logger.info('*'*120)
+        self.logger.info('Determining best incumbents')
+        results = []
+        config_cost_per_inst = {}
+        for incumbent in incs:
+            cost_per_inst = new_rh.get_instance_costs_for_config(config=incumbent)
+            config_cost_per_inst[incumbent] = cost_per_inst
+            results.append(np.mean(list(self.cost_per_inst.values())))
+        to_keep_ids = list(map(lambda x: x[0],
+                               sorted(enumerate(results), key=lambda y: y[1])))[:self.incs_per_round]
```
mlindauer: Should `incs_per_round` be a fixed parameter? I would iteratively all configurations as long the VBS score improves.

smac/facade/hydra_facade.py
AndreBiedenkapp: Will do.

smac/facade/hydra_facade.py
AndreBiedenkapp: ids are actually ints and `random.choice` would only return one value.
If I were to use `random.choices` it would choose *with replacement* (see https://docs.python.org/3/library/random.html). As we don't want to have indices for the validation set that contains duplicates `np.random.choice(upper_limit, number_of_choices, replace=False)` is the easiest way to do it.

`random.sample` would sample without replacement but wouldn't give me the IDs, which I use to remove the validation data from the training set.

smac/facade/hydra_facade.py
AndreBiedenkapp: SMAC also uses this _get_rng function.
Should they both inherit it from a super class?

smac/facade/hydra_facade.py
AndreBiedenkapp: Yes I'd say so.
`incs_per_round` gives an upper bound on the number of performed iterations.
If a user were to use Hydra on a scenario where each individual SMAC run takes a day, they could set a limit of `incs_per_round=6` to spend at most 6 days.
So the default value for `incs_per_round` should be rather large but fixed over the hydra call

smac/facade/hydra_facade.py
AndreBiedenkapp: Now the ExecuteTARunHydra can also handle function handles (which also inherit from ExecuteTARun)

AndreBiedenkapp:
@mlindauer This PR is ready to be reviewed again

.prospector.yaml
```diff
@@ -1,24 +1,32 @@
...
+    - N802               # ignore pep8 naming (func name should be lowercase i.e. get_X_y())
+    - N806               # ignore pep8 naming (func arg should be lowercase i.e. X_pred)
   options:
     max-line-length: 120
 
+pep257:
+  disable:
+    - D203               # 1 blank line required before class docstring (conflicts with D211 [no blanks allowed])
+    - D212               # Multi-line docstring summary should start at the first line(conflicts with D213 [first line])
+    - D205               # 1 blank line required between summary line and description
```
mfeurer: I suggest to leave that in.

AndreBiedenkapp: Well we would just have to decide if we want to use D211 or D203 and D212 or D213.
I just had to choose one of the conflicting rules.

PR #442, opened 2018-06-04
FIX catch writing of pcs

mlindauer:
Writing the cs in pcs can fail if not all parameter values of cat parameter are strings. The json format can deal with it. Therefore, I added a catch and try block to prevent SMAC from crashing.

### DIFF
--- a/smac/utils/io/output_writer.py
+++ b/smac/utils/io/output_writer.py
@@ -1,6 +1,6 @@
 import os
 import shutil
-import logging
+import traceback
 import typing
 
 from smac.configspace import pcs_new, json, ConfigurationSpace
@@ -10,7 +10,7 @@ class OutputWriter(object):
     """Writing scenario to file."""
 
     def __init__(self):
-        self.logger = logging.getLogger(self.__module__ + "." + self.__class__.__name__) 
+        pass
 
     def write_scenario_file(self, scenario):
         """Write scenario to a file (format is compatible with input_reader).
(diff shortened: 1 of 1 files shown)


smac/utils/io/output_writer.py
```diff
@@ -90,8 +90,11 @@ def _parse_argument(self, scenario, key: str, value):
                 except shutil.SameFileError:
                     return value  # File is already in output_dir
             elif key == 'pcs_fn' and scenario.cs is not None:
-                new_path = os.path.join(scenario.output_dir_for_this_run, 'configspace.pcs')
-                self.save_configspace(scenario.cs, new_path, 'pcs_new')
+                try:
+                    new_path = os.path.join(scenario.output_dir_for_this_run, 'configspace.pcs')
+                    self.save_configspace(scenario.cs, new_path, 'pcs_new')
+                except TypeError:
+                    traceback.print_exc()
```
mfeurer: I think this output should be sent to the logger with level `ERROR`.

mlindauer: that's what I tried but unfortunately, somehow at some point we try to serialize this class such that we cannot have a logger here.

mfeurer: You can use this helper class to make the logger pickleable: https://github.com/automl/SMAC3/blob/development/smac/utils/logging.py

mlindauer:
@mfeurer Ready for merge?

PR #443, opened 2018-06-12
ADD option to use only time reported by TA

mlindauer:
required for use with surrogate benchmarks since wallclock time should be much smaller than time reported by TA


KEggensperger:
@mlindauer Could you please also check whether the issues CodeFactor/Codacy found are critical and fix them?

mlindauer:
imo, not all a fixable (Access to member before its definition) and the remaining are not critical

KEggensperger:
Okay, then everything is fine from my side.

PR #447, opened 2018-08-09
Update install

KEggensperger:
Update documentation:
  * Requires Python >3.5.2
  * Add cmd to install requirements

Update setup.py:
  * Requires Python >3.5.2
  * Raises error if used version is lower


setup.py
```diff
@@ -36,6 +40,7 @@
     ],
     platforms=['Linux'],
     install_requires=requirements,
+    python_requires='>=3.5.2',
```
mfeurer: If it is python 3.5.2 here, shouldn't it be python 3.5.2 everywhere?

setup.py
```diff
@@ -12,6 +13,9 @@
     version = fh.readlines()[-1].split()[-1].strip("\"'")
 
 
+if sys.version_info < (3, 5):
```
mfeurer: We could add some more dependency checks here to be sure, for example cython and numpy or add them to `install_requires`. I would prefer the first to have one single way of installing this package.

setup.py
KEggensperger: That's in here because of the typing.type we use somewhere (don't remember exactly). But I am not sure whether we wanted to keep that or change the line to be compatible with 3.5.*?

mfeurer: I don't think there's a reason why anyone should use python>=3.5.0 && python<3.5.2 so we can be strict here.

setup.py
mfeurer: That these packages must be installed prior to installation. Then there's one less way to install SMAC3 (namely to not have those packages installed yet).

mfeurer:
Would it make sense to check for the requirements during import like we do for Auto-sklearn? This would make sure that all necessary packages are installed during installation time and that no weird issues happen at that point in time.

KEggensperger:
Sounds like a good idea. I had a look, but it seems like auto-sklearn maintains at least two lists of requirements. Which one do you mean makes most sense here?
  * setup.py using `install_requires` (is checked once during installation)
  * __init__.py manually checking a list (is checked on import, after installation)

What do you think about parsing requirements.txt for doing the checks in both, setup.py and __init__.py?

mfeurer:
> What do you think about parsing requirements.txt for doing the checks in both, setup.py and init.py

This probably won't work because the requirements.txt will not be installed

> Which one do you mean makes most sense here?

To prevent runtime issues we need to check this in `__init__.py`

KEggensperger:
To follow up on this: You mean maintaining two lists of dependencies? One in requirements.txt and one in __init__.py?

KEggensperger:
@mfeurer I added the same checks as in auto-sklearn. If you are happy with this PR, it is ready to be merged.

PR #456, opened 2018-10-25
parser argument use_ta_time type to 'truthy'

brenting:
I switched to the development version of SMAC since I started using AutoFolio. I noticed that my computing cluster started cancelling jobs because of time limits. For some reason, SMAC had set my 'wallclock_limit' time to 'algo_runs_timelimit' without me specifying the 'use_ta_time' argument.

I think this was the cause, as the standard value for 'use_ta_time' ((str) False) was sent to the argparser and interpreted as (Bool)True.


mlindauer:
Thanks for finding the bug and providing the fix.

Issue #469, opened 2018-12-17
Unable to install when dependencies are not preinstalled


KOLANICH:
#### Steps/Code to Reproduce
```bash
pip3 install git+https://github.com/automl/SMAC3.git
```

#### Expected Results
Should be installed

#### Actual Results
```
Traceback (most recent call last):
File "/tmp/pip-install-v4p5bk3x/smac/smac/utils/dependencies.py", line 34, in _verify_package
```

#### Versions
7f15a59e8851157d2a8b0647cc9d3b184c60e894

mlindauer:
Hi @KOLANICH 

Thank you for reporting this issue. I can reproduce your error, but I don't understand why you want to install SMAC without its dependencies. You won't be able to run SMAC without its dependencies anyway. Could you please explain what you are trying to do?

Cheers,
Marius

KOLANICH:
>I don't understand why you want to install SMAC without its dependencies.

I want the dependencies be installed **automatically** by pip (as they were before someone had added into `setup.py` the code importing `smac` for just fetching a list of contributors from inside of the package), but currently `setup.py` depends on smac be installed. Though it is not a fatal problem: when running from the clone of this repo, not installed smac will be imported, but it would be a problem if we cd into another folder and run `setup.py` by full path. This issue is a major one and must be fixed, but it is not fatal.

The fatal problem results from the fact the lib also checks presence of its dependencies in runtime on import. So `setup.py` imports `smac`, `smac` imports `.utils.dependencies` which checks presence of dependencies explicitly and fails.

BTW, the issue is fixed by #471 .

mlindauer:
Hi, 

Thank you for your thorough explanation. Now, I have a better understanding of the problem. We will discuss this internally (probably after new year) and will find a solution for the next release. Thank you again for reporting the problem.

Cheers,
Marius

mfeurer:
I suggest to do a 0.9.1 bugfix release once we merged an updated version of #471.

mlindauer:
@KOLANICH could you please check our development branch whether your issue is now solved?

shukon:
I'm not KOLANICH, but for me it does solve the problem. I'd love to see a patch-release (0.9.1) as well, since it's a little annoying to install a handful of packages manually every time I set up an environment where smac is just a dependency...

KOLANICH:
The issue is fixed. Thank you all.

PR #471, opened 2018-12-17
Fixed the installation issue

KOLANICH:


mlindauer:
@KOLANICH Thank you for the PR.

@mfeurer could you please look into it since you know more about installation scripts than I do.

mfeurer:
Thanks for the PR and pointing out this problem. 

While your proposed solution would fix the installation problem it removes the authors list from the setup (which I'm unhappy about). I'd suggest to add a file __authors__ which contains the authors list and can then be parsed by `ast.literal_eval()`. Let me know if you want to update your PR or pass this to us/me.

mfeurer:
Fixes #469.

KOLANICH:
>add a file authors which contains the authors list and can then be parsed by ast.literal_eval(). Let me know if you want to update your PR or pass this to us/me.

I suggest to
1 move all the metadata into setup.cfg (the info is automatically imported by setuptools, for examples how it may look like see any repo authored by me in my GH profile)
2 keep the list of authors there too. Or use an external file and use `file: `  to put it there (again see the examples in my repos).

mlindauer:
@KEggensperger This was fixed in your PR (already merged into the dev). So, we can close this PR?

PR #477, opened 2019-01-03
Fix 469 470 installation issues

### LINKED ISSUE #470: should smac.utils.dependencies be removed
Is it really needed? IMHO it only causes problems.

KEggensperger:
Fixes #470 and #469.

  * does no longer require smac to be installed before running setup.py
  * all dependencies are at one place: requirements.txt
  * all dependencies are checked when smac is imported
  * version number and author list is now in setup.cfg


setup.py
```diff
@@ -1,46 +1,37 @@
...
-        "Topic :: Scientific/Engineering",
-        "Topic :: Scientific/Engineering :: Artificial Intelligence",
-        "License :: OSI Approved :: BSD License",
-    ],
-    platforms=['Linux'],
+
+def get_version():
+    version_file = os.path.join(os.path.dirname(os.path.abspath(__file__)), "smac", "__init__.py")
+    for line in open(version_file):
+        if line.startswith("__version__"):
```
KOLANICH: ```ast.parse```

setup.py
```diff
@@ -1,46 +1,37 @@
-import setuptools
-import sys
-
-import smac
+#!/usr/bin/env python3
+import os
+from setuptools import setup
 
 
 with open('requirements.txt') as fh:
```
KOLANICH: You may want to use `pathlib.Path` instead of old API.

setup.cfg
```diff
@@ -1,2 +1,24 @@
 [metadata]
-description-file = README.md
\ No newline at end of file
+name = smac
+author_email = "fh@cs.uni-freiburg.de"
+description = SMAC3, a Python implementation of 'Sequential Model-based Algorithm Configuration'.
+long_description = file: README.md, CHANGELOG.md, LICENSE
+license = 3-clause BSD
```
KOLANICH: It would be better to put SPDX identifier (`BSD-3-Clause`) here.

setup.py
```diff
@@ -1,46 +1,37 @@
...
+    python_requires=">=3.5.2",
     install_requires=requirements,
-    python_requires='>=3.5.2',
-    tests_require=['mock',
-                   'nose'],
-    test_suite='nose.collector'
+    author=get_author(),
+    version=get_version(),
+    test_suite="nose.collector",
+    tests_require=["mock", "nose"]
```
KOLANICH: as far as I remember, it can be also put into `setup.cfg`

PR #486, opened 2019-01-17
MAINT SMAC facade to accept acq class and arguments instead of obj

mlindauer:
(Edited by @mfeurer ): Refactors the main SMAC interface to accept class objects and the constructor arguments instead of accepting instances and trying to inject dependencies into the existing instances. This has the advantage that objects like the model, which require types and bounds, can be passed as class objects and the SMAC builder/facade will take care of providing the required types and bounds. This is most important for classes which require the stats and runhistory object (such as target algorithm executors).

Side effects:
* Simple interface for changing the arguments to SMAC, see examples/rosenbrock_lcb.py
* Unified EPM interface
* More consistent BORF, func and ROAR facades
* Simpler class hierarchy for the initial design (removed the single and multiple initial design classes to have a single constructor signature)


mlindauer:
not ready for merging!

smac/facade/smac_facade.py
```diff
@@ -226,6 +229,9 @@ def __init__(self,
                 acquisition_function = LogEI(model=model)
             else:
                 acquisition_function = EI(model=model)
+        else:
+            acquisition_function_kwargs["model"] = model
+            acquisition_function = acquisition_function(**acquisition_function_kwargs)
```
mlindauer: @mfeurer does that looks like what you wanted for a better SMAC interface?

mfeurer: Yes :+1: 

Three suggestions:
1. Use None as the default argument for the acquisition_function_kwargs
2. It would be great if you could check that there's no model in the acquisition_function_kwargs
3. Maybe use the acquisition function kwargs for LogEI and EI, too?

mlindauer: Thanks for the feedback. I did it now for all arguments.

regarding your suggestions:

 1. I prefer empty dicts for the kwargs, since I cannot do `dict.update(None)`
 2. I define now default arguments which are overwritten by user arguments
 3. Done.

test/test_smbo/test_smbo.py
```diff
@@ -194,7 +194,7 @@ def side_effect_predict(X):
         # For each configuration it is randomly sampled whether to take it from the list of challengers or to sample it
         # completely at random. Therefore, it is not guaranteed to obtain twice the number of configurations selected
         # by EI.
-        self.assertEqual(len(challengers), 9913)
+        self.assertEqual(len(challengers), 9971)
```
mlindauer: I don't know why this number changed. In fact, I don't understand the meaning of this check given the the comment above this line.

smac/facade/smac_facade.py
```diff
@@ -74,20 +76,29 @@ class SMAC(object):
     def __init__(self,
                  scenario: Scenario,
                  tae_runner: typing.Optional[typing.Union[ExecuteTARun, typing.Callable]]=None,
+                 tae_runner_kwargs : dict={},
```
mfeurer: [Here's a reason](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python) to not use a dictionary as the default.

smac/facade/smac_facade.py
```diff
@@ -102,28 +113,47 @@ def __init__(self,
             :class:`~smac.tae.execute_func.ExecuteTAFuncDict`.
             If not set, it will be initialized with the
             :class:`~smac.tae.execute_ta_run_old.ExecuteTARunOld`.
+        tae_runner_kwargs:
```
mfeurer: please add the dict type to the docstring

smac/facade/smac_facade.py
```diff
@@ -102,28 +113,47 @@ def __init__(self,
...
+            arguments passed to constructor of '~tae_runner'
         runhistory : RunHistory
             runhistory to store all algorithm runs
+        runhistory_kwargs : RunHistory
```
mfeurer: please change the type

smac/facade/smac_facade.py
```diff
@@ -188,15 +220,29 @@ def __init__(self,
             self.scenario.transform_y = "LOG"
 
         # initialize empty runhistory
+        runhistory_def_kwargs = {'aggregate_func': aggregate_func}
+        runhistory_def_kwargs.update(runhistory_kwargs)
```
mfeurer: And here you'd only have to add an `if runhistory_kwargs:` to check if the update is legal.

smac/facade/smac_facade.py
```diff
@@ -207,44 +253,64 @@ def __init__(self,
 
         # initial EPM
         types, bounds = get_types(scenario.cs, scenario.feature_array)
+        model_def_kwargs = {
```
mfeurer: Could you please add whitespace after the colons?

smac/facade/smac_facade.py
```diff
@@ -207,44 +253,64 @@ def __init__(self,
...
-                                              log_y=scenario.transform_y in ["LOG", "LOGS"],
-                                              num_trees=scenario.rf_num_trees,
-                                              do_bootstrapping=scenario.rf_do_bootstrapping,
-                                              ratio_features=scenario.rf_ratio_features,
-                                              min_samples_split=scenario.rf_min_samples_split,
-                                              min_samples_leaf=scenario.rf_min_samples_leaf,
-                                              max_depth=scenario.rf_max_depth)
+            model = RandomForestWithInstances(**model_def_kwargs)
+        elif inspect.isclass(model):
+            model = model(**model_def_kwargs)
```
mfeurer: I'm wondering what happens if the model does not accept the number of trees as a hyperparameter? Maybe we need to add the RF-specific hyperparameters to the `if model is None` statement?

smac/facade/smac_facade.py
```diff
@@ -207,44 +253,64 @@ def __init__(self,
...
+        #TODO since not all our models implement the AbstractEPM,
+        #no check of object done here
+            
         # initial acquisition function
+        acq_def_kwargs = {'model': model}
+        acq_def_kwargs.update(acquisition_function_kwargs)
```
mfeurer: I think we should add a check here that model is not overwritten from the outside.

smac/facade/smac_facade.py
```diff
@@ -207,44 +253,64 @@ def __init__(self,
 
         # initial EPM
         types, bounds = get_types(scenario.cs, scenario.feature_array)
+        model_def_kwargs = {
+            'types':types,
```
mfeurer: I think we should add a check here that types and bounds and the feature_array and the seed are not overwritten from the outside.

smac/facade/smac_facade.py
```diff
@@ -207,44 +253,64 @@ def __init__(self,
...
 
         # initialize optimizer on acquisition function
+        acq_func_opt_kwargs = {
+            'acquisition_function':acquisition_function,
+            'config_space':scenario.cs,
+            'rng':np.random.RandomState(seed=rng.randint(MAXINT)),
+            'max_steps':scenario.sls_max_steps,
+            'n_steps_plateau_walk':scenario.sls_n_steps_plateau_walk
+            }
+        acq_func_opt_kwargs.update(acquisition_function_optimizer_kwargs)
```
mfeurer: I think we need a check here, too.

mfeurer: And colons in the dictionary definition.

smac/facade/smac_facade.py
```diff
@@ -253,27 +319,31 @@ def __init__(self,
...
+            tae_runner = ExecuteTAFuncDict(**tae_def_kwargs)
+        elif inspect.isclass(tae_runner):
+            tae_runner = tae_runner(**tae_def_kwargs)
+        elif isinstance(tae_runner, ExecuteTARun):
+            if tae_runner.stats is None:
+                tae_runner.stats = self.stats
+            if tae_runner.runhistory is None:
+                tae_runner.runhistory = runhistory
+            if tae_runner.crash_cost != scenario.cost_for_crash:
+                tae_runner.crash_cost = scenario.cost_for_crash
```
mfeurer: Could you please add a call to the logger here?

smac/facade/smac_facade.py
```diff
@@ -287,170 +357,145 @@ def __init__(self,
...
+        elif inspect.isclass(intensifier):
+            intensifier = intensifier(**intensifier_def_kwargs)
+        elif isinstance(intensifier, Intensifier):
+            if intensifier.tae_runner is None:
+                intensifier.tae_runner = tae_runner
+            if intensifier.stats is None:
+                intensifier.stats = self.stats
+            if intensifier.traj_logger is None:
+                intensifier.traj_logger = traj_logger
+        #TODO: no further check since we have no abstract intensifier
```
mfeurer: How about an else which raises an exception?

smac/utils/io/cmd_reader.py
```diff
@@ -6,12 +6,14 @@
 import re
 import shlex
 import sys
+import time
+import typing
```
mfeurer: I think there should be a whitespace between built-ins and 3rd-party packages.

test/test_smbo/test_smbo.py
mfeurer: I think these numbers are simply to ensure that nothing breaks out of a sudden. Since you did some major changes, I think these changes in numbers are okay.

examples/rosenbrock_lcb.py
```diff
@@ -0,0 +1,69 @@
...
+# Import ConfigSpace and different types of parameters
+from smac.configspace import ConfigurationSpace
+from ConfigSpace.hyperparameters import UniformFloatHyperparameter
+
+# Import SMAC-utilities
+from smac.scenario.scenario import Scenario
+from smac.facade.smac_facade import SMAC
+from smac.optimizer.acquisition import LCB
+from smac.initial_design.latin_hypercube_design import LHDesign
+from smac import initial_design
```
mfeurer: What is this used for?

examples/rosenbrock_lcb.py
```diff
@@ -0,0 +1,69 @@
...
+        initial_design_kwargs={'n_configs_x_params':4,
+                               'max_config_fracs':1.0},
+        runhistory2epm=RunHistory2EPM4InvScaledCost,
+        model_kwargs={'num_trees': 42},
+        acquisition_function_optimizer_kwargs={'max_steps':100},
+        acquisition_function=LCB,
+        acquisition_function_kwargs={'par':0.01}
+        )
+
+#smac = SMAC(scenario=scenario, rng=np.random.RandomState(42),
```
mfeurer: Could you please remove this piece of code?

examples/smac_smac.py
```diff
@@ -0,0 +1,194 @@
...
+                              'min_samples_split': smac_config['min_samples_split'],
+                              'min_samples_leaf': smac_config['min_samples_leaf']},
+                acquisition_function=acq[smac_config['acq_func']],
+                acquisition_function_kwargs={'par':smac_config['par'] if smac_config['par'] is not None else smac_config['lcb_par']}
+            )
+    
+    # hard coded for artificial functions
+    smac.solver.scenario.intensification_percentage = 1e-10
+    smac.solver.intensifier.min_chall = 1
+    smac.solver.acq_optimizer.n_sls_iterations = 10
```
mfeurer: We could tune these values, too

examples/smac_smac.py
```diff
@@ -0,0 +1,194 @@
...
+    smac.solver.scenario.acq_opt_challengers = 1000
+    smac.solver.predict_incumbent = True
```
mfeurer: Especially this one :)

smac/epm/rf_with_instances.py
```diff
@@ -117,7 +117,7 @@ def __init__(self, types: np.ndarray,
 
         self.logger = logging.getLogger(self.__module__ + "." +
                                         self.__class__.__name__)
-
+
```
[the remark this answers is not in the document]
mlindauer: actually that was a codacy recommendation to have an empty line at the end. I think we also do this in nearly every module.

smac/facade/smac_facade.py
mlindauer: okay, good point

smac/epm/rf_with_instances.py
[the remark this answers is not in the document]
mfeurer: Weird that codacy suggests changing files which are not affected by this commit.

smac/facade/smac_facade.py
mlindauer: the code assumes so far that all constructors can accept all arguments even if they will not use them. I don't want to implement a complicated logic only to prevent that some arguments are not used.

smac/facade/smac_facade.py
mlindauer: why should we? I don't see a reason why someone should do that but I thought our code should be more flexible. So, I would not additional constraints.

smac/facade/smac_facade.py
mlindauer: No, if someone has implemented his/her own intensification approach (with the same API as we use), he/she should be able to still use the SMAC facade

examples/smac_smac.py
mlindauer: I had the feeling from our BORF experiments that these parameters have no big impact on the performance. To make it not too hard for SMACSMAC, I would not touch it for now.

smac/facade/smac_facade.py
mfeurer: But shouldn't that be covered by the `elif isinstance(...)`? Otherwise, there's no/little guarantee that the objects actually fit together.

mfeurer:
@mlindauer @AndreBiedenkapp @KEggensperger I think the rewriting of the SMAC interface itself is done and at least I would be very happy with the current state. Please let me know what you think (the files rosenbrock_lcb.py an smac_facade.py should suffice for this), then I could work on adding a few more tests and fixing the remaining failing unit test (hydra facade).

smac/facade/borf_facade.py
```diff
@@ -47,55 +48,61 @@ class BORF(SMAC):
...
         # only 1 configuration per SMBO iteration
-        self.solver.scenario.intensification_percentage = 1e-10
-        self.solver.intensifier.min_chall = 1
-        
+        intensifier_kwargs = kwargs.get('intensifier_kwargs', dict())
+        intensifier_kwargs['min_chall'] = 1
+        kwargs['intensifier_kwargs'] = intensifier_kwargs
+        scenario.intensification_percentage = 1e-10
+
+        model = RandomForestWithInstances
```
mlindauer: maybe call it `model_class`?

smac/facade/smac_facade.py
```diff
@@ -184,19 +214,35 @@ def __init__(self,
...
+           'rng': rng
+        }
+        if random_configuration_chooser_kwargs is not None:
+            rand_conf_chooser_kwargs.update(random_configuration_chooser_kwargs)
+        if random_configuration_chooser is None:
+            if 'prob' not in rand_conf_chooser_kwargs:
+                rand_conf_chooser_kwargs['prob'] = scenario.rand_prob
+            random_configuration_chooser = ChooserProb(**rand_conf_chooser_kwargs)
+        elif inspect.isclass(random_configuration_chooser):
+            random_configuration_chooser = random_configuration_chooser(**rand_conf_chooser_kwargs)
```
mlindauer: Why is the injection of `prob` here missing, but done in line 239?

smac/facade/smac_facade.py
```diff
@@ -207,75 +253,100 @@ def __init__(self,
...
+                'log_y': scenario.transform_y in ["LOG", "LOGS"],
+                'num_trees': scenario.rf_num_trees,
+                'do_bootstrapping': scenario.rf_do_bootstrapping,
+                'ratio_features': scenario.rf_ratio_features,
+                'min_samples_split': scenario.rf_min_samples_split,
+                'min_samples_leaf': scenario.rf_min_samples_leaf,
+                'max_depth': scenario.rf_max_depth,
+            }.items():
+                if key not in model_def_kwargs:
+                    model_def_kwargs[key] = value
```
mlindauer: That's looks more complicated than my solution with an `update()`. Why did you use a for loop here?

smac/facade/borf_facade.py
mfeurer: Makes sense, I will update this

smac/facade/smac_facade.py
mfeurer: `prob` is not a mandatory argument for a `RandomConfigurationChooser`, therefore, it is only passed to the chooser which requires it.

smac/facade/smac_facade.py
mfeurer: If we would use update here we would overwrite arguments provided by the user which were added above in line 264. This is in line with adding a single argument as for example done in line 238/239 for the random configuration chooser.

smac/facade/smac_facade.py
mlindauer: but we don't know whether the class `random_configuration_chooser` will require it or not

smac/facade/smac_facade.py
mlindauer: > If we would use update here we would overwrite arguments

That depends on how you apply the update. You could simply define your default arguments and overwrite them by the user provided arguments.

smac/facade/smac_facade.py
mfeurer: Correct. In fact, we know for no method except the default method what kind of hyperparameters it requires. To rephrase my statement, the abstract class `RandomConfigurationChooser` requires only the random state and therefore, only the random state is passed by default.

If you think that's not good we should rather add the default value to constructor of the respective class instead of adding it to this dictionary. Ideally, the dictionaries only add new "default" values which are constructed inside the smac facade. At some point in the future it would be good if we move all other arguments somewhere else.

smac/facade/smac_facade.py
mfeurer: Like:
```
           new_kwargs = {
                'log_y': scenario.transform_y in ["LOG", "LOGS"],
                'num_trees': scenario.rf_num_trees,
                'do_bootstrapping': scenario.rf_do_bootstrapping,
                'ratio_features': scenario.rf_ratio_features,
                'min_samples_split': scenario.rf_min_samples_split,
                'min_samples_leaf': scenario.rf_min_samples_leaf,
                'max_depth': scenario.rf_max_depth,
            }
            new_kwargs.update(model_def_kwargs)
            model_def_kwargs = new_kwargs
```
?

examples/rosenbrock_lcb.py
```diff
@@ -0,0 +1,65 @@
...
+from smac.initial_design.latin_hypercube_design import LHDesign
+from smac.runhistory.runhistory2epm import RunHistory2EPM4InvScaledCost
+
+def rosenbrock_2d(x):
+    """ The 2 dimensional Rosenbrock function as a toy model
+    The Rosenbrock function is well know in the optimization community and
+    often serves as a toy problem. It can be defined for arbitrary
+    dimensions. The minimium is always at x_i = 1 with a function value of
+    zero. All input parameters are continuous. The search domain for
+    all x's is the interval [-5, 5].
```
KEggensperger: I think more common is the interval [-5, 10]

examples/rosenbrock_lcb.py
```diff
@@ -0,0 +1,65 @@
...
+
+x0 = UniformFloatHyperparameter("x0", -5, 5, default_value=-3)
+x1 = UniformFloatHyperparameter("x1", -5, 5, default_value=-4)
+cs.add_hyperparameters([x0,x1])
+
+# Scenario object
+scenario = Scenario({"run_obj": "quality",   # we optimize quality (alternatively runtime)
+                     "runcount-limit": 200,  # maximum function evaluations
+                     "cs": cs,               # configuration space
+                     "deterministic": "true"
```
KEggensperger: Can this also be a bool?

examples/rosenbrock_lcb.py
```diff
@@ -0,0 +1,65 @@
...
+
+# Optimize, using a SMAC-object
+print("Optimizing! Depending on your machine, this might take a few minutes.")
+smac = SMAC(scenario=scenario, rng=np.random.RandomState(42),
+        tae_runner=rosenbrock_2d,
+        initial_design=LHDesign,
+        initial_design_kwargs={'n_configs_x_params':4,
+                               'max_config_fracs':1.0},
+        runhistory2epm=RunHistory2EPM4InvScaledCost,
+        model_kwargs={'num_trees': 42},
```
KEggensperger: Would any of these complain if given wring kwargs, speicificially typos or not specifying a parameter?

examples/rosenbrock_lcb.py
```diff
@@ -0,0 +1,65 @@
...
+                     })
+
+# Example call of the function
+# It returns: Status, Cost, Runtime, Additional Infos
+def_value = rosenbrock_2d(cs.get_default_configuration())
+print("Default Value: %.2f" % (def_value))
+
+# Optimize, using a SMAC-object
+print("Optimizing! Depending on your machine, this might take a few minutes.")
+smac = SMAC(scenario=scenario, rng=np.random.RandomState(42),
```
KEggensperger: Can seed also be an int?

smac/epm/gaussian_process.py
```diff
@@ -57,21 +59,16 @@ def __init__(
         self,
         types: np.ndarray,
         bounds: typing.List[typing.Tuple[float, float]],
+        seed: int,
```
KEggensperger: Why do we switch back to forwarding integer seeds?

KEggensperger: Is the seed even used in here? If the GP deterministic given a seed?

smac/epm/random_epm.py
```diff
@@ -20,17 +20,10 @@ class RandomEPM(AbstractEPM):
     rng : np.random.RandomState
```
KEggensperger: Not sure whether we use this anymore, but shouldn't this switch to seed as well?

smac/epm/rf_with_instances.py
```diff
@@ -39,6 +39,7 @@ class RandomForestWithInstances(AbstractEPM):
 
     def __init__(self, types: np.ndarray,
                  bounds: typing.List[typing.Tuple[float, float]],
+                 seed: int = 42,
```
KEggensperger: GPs do not have a default seed. Why do RFs have one?

smac/facade/borf_facade.py
KEggensperger: agree

smac/facade/func_facade.py
```diff
@@ -18,9 +18,9 @@
 
 def fmin_smac(func: typing.Callable,
               x0: typing.List[float],
-              bounds: typing.List[typing.List[float]],
-              maxfun: int=-1,
-              rng: np.random.RandomState=None,
+              bounds: typing.List[typing.Iterable[float]],
+              maxfun: int = -1,
+              rng: typing.Union[np.random.RandomState, int] = None,
               scenario_args: typing.Mapping[str, typing.Any]=None,
```
KEggensperger: If you add spaces for the lines above, you should also add them here :-)
scenario_args: typing.Mapping[str, typing.Any]X=XNone

smac/facade/smac_facade.py
```diff
@@ -102,29 +110,48 @@ def __init__(self,
...
             If not set, it will be initialized with the
             :class:`~smac.tae.execute_ta_run_old.ExecuteTARunOld`.
+        tae_runner_kwargs: Optional[dict]
+            arguments passed to constructor of '~tae_runner'
         runhistory : RunHistory
             runhistory to store all algorithm runs
+        runhistory_kwargs : Optional[dict]
+            arguments passed to constructor of runhistory.
+            We strongly advise against changing the aggregation function,
+            since it will break some code assumption
```
KEggensperger: assumptions

smac/facade/smac_facade.py
```diff
@@ -207,75 +253,100 @@ def __init__(self,
...
+                "Argument acquisition_function must be None or an object implementing the "
+                "AbstractAcquisitionFunction, not %s."
+                % type(acquisition_function)
+            )
 
         # initialize optimizer on acquisition function
+        acq_func_opt_kwargs = {
+            'acquisition_function': acquisition_function,
+            'config_space': scenario.cs,
+            'rng': np.random.RandomState(seed=rng.randint(MAXINT)),
```
KEggensperger: Why is it necessary to generate a new rs here? In line 233 we simply use directly the rng

smac/facade/smac_facade.py
```diff
@@ -287,170 +358,139 @@ def __init__(self,
...
-        # inject cost_for_crash
-        if tae_runner.crash_cost != scenario.cost_for_crash:
-            tae_runner.crash_cost = scenario.cost_for_crash
-
         # initialize intensification
+        intensifier_def_kwargs = {
+            'tae_runner': tae_runner,
+            'stats': self.stats,
+            'traj_logger': traj_logger,
+            'rng': rng,
```
KEggensperger: see above

smac/facade/smac_facade.py
```diff
@@ -287,170 +358,139 @@ def __init__(self,
...
             raise ValueError(
                 "Either use initial_design or initial_configurations; but not both")
 
+        init_design_def_kwargs = {
+            'tae_runner': tae_runner,
+            'scenario': scenario,
+            'stats': self.stats,
+            'traj_logger': traj_logger,
+            'runhistory': runhistory,
+            'rng': rng,
```
KEggensperger: ^

smac/facade/smac_facade.py
```diff
@@ -504,9 +544,12 @@ def optimize(self):
         return incumbent
 
     def validate(self,
-                 config_mode: typing.Union[typing.List[Configuration], np.ndarray, str]='inc',
-                 instance_mode: typing.Union[typing.List[str], str]='train+test',
-                 repetitions: int=1, use_epm: bool=False, n_jobs: int=-1, backend: str='threading'):
+                 config_mode: Union[List[Configuration], np.ndarray, str] = 'inc',
+                 instance_mode: Union[List[str], str]='train+test',
```
KEggensperger: space missing

smac/facade/smac_facade.py
```diff
@@ -504,9 +544,12 @@ def optimize(self):
...
+                 repetitions: int = 1,
+                 use_epm: bool = False,
+                 n_jobs: int=-1, backend:
```
KEggensperger: space missing

smac/facade/borf_facade.py
AndreBiedenkapp: :+1

smac/optimizer/smbo.py
```diff
@@ -298,15 +298,15 @@ def _get_incumbent_value(self):
...
                 raise ValueError("Runhistory is empty and the cost value of "
                                  "the incumbent is unknown.")
             incumbent_value = self.runhistory.get_cost(self.incumbent)
             if isinstance(self.rh2EPM,RunHistory2EPM4LogCost):
                 incumbent_value = np.log(incumbent_value)
-        
+
         return incumbent_value
 
     def validate(self, config_mode='inc', instance_mode='train+test',
```
AndreBiedenkapp:
```suggestion
    def validate(self, config_mode='inc', instance_mode= 'train+test',
```

smac/facade/func_facade.py
```diff
@@ -18,9 +18,9 @@
 
 def fmin_smac(func: typing.Callable,
               x0: typing.List[float],
-              bounds: typing.List[typing.List[float]],
-              maxfun: int=-1,
-              rng: np.random.RandomState=None,
+              bounds: typing.List[typing.Iterable[float]],
+              maxfun: int = -1,
```
AndreBiedenkapp:
```suggestion
              maxfun: int=-1,
```

smac/facade/func_facade.py
```diff
@@ -18,9 +18,9 @@
...
+              rng: typing.Union[np.random.RandomState, int] = None,
```
AndreBiedenkapp:
```suggestion
              rng: typing.Union[np.random.RandomState, int]=None,
```

smac/epm/rf_with_instances.py
AndreBiedenkapp:
```suggestion
                 seed: int=42,
```

smac/epm/rf_with_instances_hpo.py
```diff
@@ -47,11 +47,11 @@ def __init__(
         self,
         types: np.ndarray,
         bounds: typing.List[typing.Tuple[float, float]],
+        seed: int = 42,
```
AndreBiedenkapp:
```suggestion
        seed: int=42,
```

smac/optimizer/ei_optimization.py
```diff
@@ -167,8 +167,8 @@ def __init__(
             acquisition_function: AbstractAcquisitionFunction,
             config_space: ConfigurationSpace,
             rng: Union[bool, np.random.RandomState] = None,
-            max_steps: Optional[int]=None,
-            n_steps_plateau_walk: int=10,
+            max_steps: Optional[int] = None,
+            n_steps_plateau_walk: int = 10,
```
AndreBiedenkapp: The two lines above also should change the " = " to "="

smac/optimizer/random_configuration_chooser.py
```diff
@@ -54,7 +59,13 @@ def check(self, iteration) -> bool:
 
 class ChooserLinearCoolDown(RandomConfigurationChooser):
 
-    def __init__(self, start_modulus: float=2.0, modulus_increment: float=0.3, end_modulus: float=np.inf):
+    def __init__(
+        self,
+        rng: np.random.RandomState,
+        start_modulus: float = 2.0,
+        modulus_increment: float = 0.3,
+        end_modulus: float = np.inf,
```
AndreBiedenkapp: See all my other comments so far

smac/utils/io/output_directory.py
```diff
@@ -8,9 +7,9 @@
 
 
 def create_output_directory(
-        scenario: typing.Type[Scenario],
+        scenario: Scenario,
         run_id: int,
-        logger: Logger=None,
+        logger: Logger = None,
```
AndreBiedenkapp:
```suggestion
        logger: Logger=None,
```

AndreBiedenkapp:
@mfeurer looks very good. I've only found issues related to " = "  vs "=" to complain ;)

examples/rosenbrock_lcb.py
mfeurer: Maybe, but I think that's a more general issue with the scenario.

examples/rosenbrock_lcb.py
mfeurer: Yes. These are passed to the different classes. If they do not allow for that argument, there will be an exception.

smac/epm/gaussian_process.py
mfeurer: We don't switch back, we adapt the GP to follow the interface of the base_epm. And yes, the GP is deterministic give the seed.

examples/rosenbrock_lcb.py
mfeurer: Thanks, just changed that.

examples/rosenbrock_lcb.py
mfeurer: Yes, I just changed that.

smac/epm/random_epm.py
mfeurer: I removed the comment as it should be inherited in the documentation.

smac/epm/rf_with_instances.py
mfeurer: I just removed the rf default seed.

smac/facade/func_facade.py
mfeurer: Done

smac/facade/smac_facade.py
mfeurer: Done

smac/facade/smac_facade.py
mfeurer: It's not. I just changed that.

smac/facade/smac_facade.py
mfeurer: I just corrected that.

smac/facade/smac_facade.py
mfeurer: That too.

mfeurer:
@AndreBiedenkapp I think I also corrected all your suggestions.

mfeurer:
Tests are still to come.

smac/facade/bogp_facade.py
```diff
@@ -61,41 +66,46 @@ def __init__(self, model_type='gp_mcmc', **kwargs):
...
+
         super().__init__(**kwargs)
 
         if self.solver.scenario.n_features > 0:
             raise NotImplementedError("BOGP cannot handle instances")
 
         self.logger.info(self.__class__)
 
-        self.solver.random_configuration_chooser.prob = 0.0
+        # assumes random chooser for random configs
```
mfeurer: Shouldn't this happen before the call to `super().__init__(**kwargs)`?

smac/facade/bogp_facade.py
```diff
@@ -61,41 +66,46 @@ def __init__(self, model_type='gp_mcmc', **kwargs):
...
+        random_config_chooser_kwargs['prob'] = random_config_chooser_kwargs.get('prob', 0.0)
+        kwargs['random_configuration_chooser_kwargs'] = random_config_chooser_kwargs
 
         # only 1 configuration per SMBO iteration
-        self.solver.scenario.intensification_percentage = 1e-10
-        self.solver.intensifier.min_chall = 1
+        intensifier_kwargs = kwargs.get('intensifier_kwargs', dict())
+        intensifier_kwargs['min_chall'] = 1
+        kwargs['intensifier_kwargs'] = intensifier_kwargs
+        scenario.intensification_percentage = 1e-10
```
mfeurer: I'm really wondering how we can avoid such hacking of the scenario object. Maybe we can also alter it earlier? Or make all such "scenario" arguments which change the behavior of SMAC (and therefore aren't scenario arguments) be passed via the kwargs, and then handled by the respective facades? Then, only the "real" scenario arguments are handed around in the SMAC code base.

PR #492, opened 2019-02-21
rename facades

KEggensperger:


examples/rf.py
```diff
@@ -108,4 +106,4 @@ def rmse(y, y_pred):
     incumbent = smac.solver.incumbent
 
 inc_value = smac.get_tae_runner().run(incumbent, 1)[1]
-print("Optimized Value: %.2f" % (inc_value))
+print("Optimized Value: %.2f" % inc_value)
```
mlindauer: I want my paranthesis back T___T

smac/utils/io/cmd_reader.py
```diff
@@ -136,6 +136,8 @@ class ReadPCSFileAction(Action):
     def __call__(self, parser: ArgumentParser, namespace: Namespace, values: list, option_string: str=None):
         fn = values
         if fn:
+            print(fn)
```
mlindauer: please rm prints

mlindauer:
Please check whether this PR still fails after merging with the new dev branch

mlindauer:
the travis test on the examples fails with the message: `The job exceeded the maximum time limit for jobs, and has been terminated.`  How do we fix that? I would like to merge that PR asap such that I can use the new facades in SMACSMAC

mlindauer:
The folder `doc/apidoc` is outdated such that not all files are rendered properly. Unfortunately, I don't know how to update the folder.

KEggensperger:
Okay, sorry, I'll fix the remaining issue.

mfeurer:
@shukon it seems that there's a problem with the docs, especially the apidocs directory. Is that generated automatically, or should we update that manually to reflect changes in the classes locations we're making?

KEggensperger:
@mfeurer I already talked to André, the apidoc folder can be generated automatically. However, by looking at `create_doc.sh` I would guess it is not created automatically with every commit. I would expect adding ```make buildapi``` in line 23 should fix this issue. What do you think?

mfeurer:
I think you're right and I think it would be great to have the apidoc directory be generated automatically during every build.

PR #497, opened 2019-03-01
MAINT speed up rf predictions for log transformed data

mfeurer:
by reshuffling the for loop for computing the mean performance.

### DIFF
--- a/smac/epm/rf_with_instances.py
+++ b/smac/epm/rf_with_instances.py
@@ -1,4 +1,5 @@
 import logging
+import math
 import typing
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


mfeurer:
can reduce runtime by a factor of 2.

smac/runhistory/runhistory2epm.py
```diff
@@ -461,6 +461,7 @@ def transform_response_values(self, values: np.ndarray) -> np.ndarray:
         """
 
         min_y = self.min_y - (self.perc - self.min_y)  # Subtract the difference between the percentile and the minimum
+        min_y -= 1e-5  # Minimal value to avoid numerical issues in the log scaling below
```
mlindauer: This implies that we are always at least 1e-5 away from the optimum. Therefore, it could be harder to hit the optimum with higher precision. In practice, this might not matter; however in our BORF experiments, we always wanted to achieve a regret of 1e-6 (which is of course also an arbitrary number).
In the end, I wonder whether 1e-5 is a reasonable gap to be used here.

smac/epm/rf_with_instances.py
```diff
@@ -40,16 +40,16 @@ class RandomForestWithInstances(AbstractEPM):
...
-                 num_trees: int=N_TREES,
-                 do_bootstrapping: bool=True,
-                 n_points_per_tree: int=-1,
-                 ratio_features: float=5. / 6.,
-                 min_samples_split: int=3,
-                 min_samples_leaf: int=3,
-                 max_depth: int=2**20,
-                 eps_purity: float=1e-8,
-                 max_num_nodes: int=2**20,
+                 log_y: bool = False,
```
mlindauer: you did it again.... harder to find the actual changes.

smac/epm/rf_with_instances.py
```diff
@@ -195,21 +195,36 @@ def _predict(self, X: np.ndarray) -> typing.Tuple[np.ndarray, np.ndarray]:
...
+                third_dimension = max(max_num_leaf_data, third_dimension)
+
+            # Transform list of 2d arrays into a 3d array
+            preds_as_array = np.empty((X.shape[0], self.rf_opts.num_trees, third_dimension)) * np.NaN
+            for i, preds_per_tree in enumerate(all_preds):
+                for j, pred in enumerate(preds_per_tree):
+                    preds_as_array[i, j, :len(pred)] = pred
+
+            # Do all necessary computation with vectorized functions
+            preds_as_array = np.log(np.nanmean(np.exp(preds_as_array), axis=2) + 1e-10)
```
mlindauer: why have you used here 1e-10 and not again 1e-5?

smac/runhistory/runhistory2epm.py
mfeurer: I'd be happy to change this to 1e-10 or something else which is small enough. On the other hand I was wondering why you want to use a scaled version on the first place.

smac/epm/rf_with_instances.py
mfeurer: unclear ;)

smac/runhistory/runhistory2epm.py
mlindauer: Yes, please change it to 1e-10.
We can discuss the scaled version offline if you want.

mfeurer:
> Actually, I'm surprised that this is actually faster than the old code because it has more for-loops and you copy data quite a bit between arrays.

One more for-loop ;) It appears that calling mean on small arrays is more expensive than copying data around.

The more interesting question is what lower value for numerical stability would be acceptable? You just suggested 1e-10, it appears that the examples also work with 1e-14, but let's be conservative. I'll update the PR in a second.

PR #500, opened 2019-03-05
Vectorize local search

mfeurer:
This PR vectorizes the local search.

### DIFF
--- a/smac/optimizer/ei_optimization.py
+++ b/smac/optimizer/ei_optimization.py
@@ -214,6 +214,8 @@ class LocalSearch(AcquisitionFunctionMaximizer):
 
         # sort according to acq value
         configs_acq.sort(reverse=True, key=lambda x: x[0])
+        for _, inc in configs_acq:
+            inc.origin = 'Local Search'
 
         return configs_acq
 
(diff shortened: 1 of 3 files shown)


mfeurer:
Okay, so I commented the code. Feedback welcome @mlindauer @AndreBiedenkapp @KEggensperger

AndreBiedenkapp:
@mfeurer Code looks correct to me

smac/optimizer/ei_optimization.py
```diff
@@ -247,75 +252,139 @@ def _get_initial_points(self, num_points, runhistory):
...
+                            if acq_val[acq_index] > acq_val_incumbents[i]:
+                                incumbents[i] = neighbors[acq_index]
+                                acq_val_incumbents[i] = acq_val[acq_index]
+                                new_neighborhood[i] = True
+                                improved[i] = True
+                                local_search_steps[i] += 1
+                                neighbors_w_equal_acq[i] = []
+                                obtain_n[i] = 1
+                            # Found an equally well performing configuration, keeping it for plateau walking
+                            elif acq_val[acq_index] == acq_val_incumbents[i]:
```
mlindauer: I don't get why you distinguish between plateau steps and improving steps here.

mfeurer: It was the same before, I did not change the behavior here.

mfeurer:
Yeah, it's speed vs readability :(

I checked that the rosenbrock example is quantitatively similar. 

I also used the local search procedure to actually maximize (-1 * rosenbrock4d) from the unit tests, and the new versions actually performs better :open_mouth: no idea why...

mfeurer:
I added an additional unit test to make sure that the local search procedure actually does something.

PR #509, opened 2019-05-07
Change local search starting points

mfeurer:
Local search now starts from:

* the `num_points` observed configurations with the highest EI
* the `num_points` observed configurations with the lowest cost
* the `num_points` unobserved configurations with the highest EI, configurations obtained by random search

To account for the increased necessary computation (because of the up to three fold increase of initial points), we reduce the number of local searches from 100 to 10.


mfeurer:
The examples unit test actually depends on #508 being merged first.

PR #511, opened 2019-05-13
Fix error linalg error handling of the Gaussian process

mfeurer:


smac/epm/gaussian_process.py
```diff
@@ -114,9 +114,9 @@ def _train(self, X: np.ndarray, y: np.ndarray, do_optimize: bool=True):
                 if i == n_tries:
                     raise e
                 # Assume that the last entry of theta is the noise
-                theta = self.kernel.theta
+                theta = np.exp(self.kernel.theta)
```
KEggensperger: Any reason why you add 1 and not any other number?

KEggensperger: Also compared to before: Why is this now trying harder to fix non-PSD kernels?

PR #515, opened 2019-06-06
Successive Halving - SMAC

ashraaghav:
* Added successive halving as an intensification procedure
* Added & modified examples (MLP example for 'cutoff' as a budget is still pending)


smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+
+        if time_bound < self._min_time:
+            raise ValueError("time_bound must be >= %f" % self._min_time)
+
+        if isinstance(challengers, ChallengerList):
+            challengers = challengers.challengers
+
+        # selecting first 'n' challengers for the current run, based on eta
+        curr_challengers = challengers[:self.initial_challengers]
+        # challengers can repeated only if optimizing across multiple seeds
```
mlindauer: `can be repeated`

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+            top challenger configurations, sorted in increasing costs
+        """
+        # extracting costs for each given configuration
+        config_costs = {}
+        for c in configs:
+            runs = run_history.get_runs_for_config(c)
+            perf = run_history.aggregate_func(c, run_history, runs)
+            config_costs[c] = perf
+
+        configs_sorted = sorted(config_costs, key=config_costs.get)
```
mlindauer: what would happen if we consider a configuration for a second time? I worry that we will compare configurations based on a different kinds of sets. I think we have to limit the considered `runs` performed in the current SH run

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+        chal_sum_cost = sum_cost(config=challenger,
+                                 instance_seed_pairs=chall_inst_seeds,
+                                 run_history=run_history)
+        cutoff = min(self.cutoff,
+                     inc_sum_cost * self.adaptive_capping_slackfactor -
+                     chal_sum_cost
+                     )
+        return cutoff
+
+    def _compare_configs(self, incumbent: Configuration, challenger: Configuration,
```
mlindauer: that's the same function as in the original intensification, right?
if yes, please avoid code duplication. 
(the same applies to the adaptive capping code)

smac/optimizer/smbo.py
```diff
@@ -153,9 +154,28 @@ def start(self):
...
         if self.stats.ta_runs == 0 and self.incumbent is None:
-            self.incumbent = self.initial_design.run()
+            logging.info('Running initial design')
+            # Run the initial deisgn only if intensifying using SMAC's procedure
+            # SH or HB methods get initialized by 1 run with random configurations
+            if not isinstance(self.intensifier, SuccessiveHalving):
+                # Intensifier initialization
+                self.incumbent = self.initial_design.run()
+            else:
+                # SH/HB initialization - get default configuration + random configurations
```
mlindauer: shouldn't that be handled inside of the initial design?
E.g., if we evaluate multiple configurations in the "old" initial design, we also use intensification to compare these.

smac/utils/io/cmd_reader.py
```diff
@@ -426,6 +426,12 @@ def _add_smac_options(self):
                                default=True, type=truthy,
                                help="If true, *SMAC* will abort if the first run of "
                                     "the target algorithm crashes.")
+        smac_opts.add_argument("--limit-resources", "--limit_resources",
+                               dest='limit_resources',
+                               default=True, type=truthy,
+                               help="If true, *SMAC* will use pynisher to limit time and memory for "
+                                    "the target algorithm. Allows SMAC to use all resources available. "
+                                    "Set to 'True' by default. (Use with caution!)")
```
mlindauer: please also mention that this is only applicable for the func TAE.

examples/SMAC4HPO_rf.py
```diff
@@ -1,4 +1,13 @@
+"""
```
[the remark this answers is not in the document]
ashraaghav: I edited some of the examples to include different ways of passing instances, cutoffs and objectives shown in SMAC, for which I added the description on what it does at the beginning of the file. 

For example, 
RF - multiple instances; quality objective; no cutoff limits
MLP - single instance; quality objective; cutoff passed to target algorithm -> no limits from SMAC

Does this lead to edits in the SMAC-documentation pipeline?

examples/SMAC4HPO_rf.py
```diff
@@ -13,10 +22,22 @@
 from smac.scenario.scenario import Scenario
 from smac.facade.smac_hpo_facade import SMAC4HPO
 
-boston = load_boston()
+INSTANCE_PATH = 'data/instances/'
+
+
+def load_data_from_file(container_path: str, delimiter=','):
```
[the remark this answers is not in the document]
ashraaghav: I used the local data that is already a part of the examples. I only split it into multiple instances. It can serve as an example for passing multiple instances as budget to successive halving too.

examples/SMAC4HPO_svm.py
```diff
@@ -103,7 +105,8 @@ def svm_from_cfg(cfg):
 # Optimize, using a SMAC-object
 print("Optimizing! Depending on your machine, this might take a few minutes.")
 smac = SMAC4HPO(scenario=scenario, rng=np.random.RandomState(42),
-        tae_runner=svm_from_cfg)
+                tae_runner=svm_from_cfg,
+                intensifier_type='intensify')  # intensifier to use - 'intensify' or 'sh'
```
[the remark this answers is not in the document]
ashraaghav: No. It defaults to 'intensify'. I can remove this if required.

examples/spear_qcp/SMAC4AC_spear_qcp.py
```diff
@@ -0,0 +1,33 @@
+"""
```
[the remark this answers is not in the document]
ashraaghav: This is the same as the `run.sh` file, but uses SMAC4AC interface. I added it since there was no python example for running SMAC4AC so far.

examples/spear_qcp/scenario.txt
```diff
@@ -5,6 +5,7 @@ deterministic = 0
 run_obj = runtime
 overall_obj = PAR10
 cutoff_time = 5
-wallclock-limit = 30
+wallclock-limit = 100
```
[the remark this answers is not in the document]
ashraaghav: I mainly added it to let successive halving complete a couple of iterations since it evaluates a lot more configurations (so takes some time). I will change this back if it is not necessary.

smac/facade/smac_ac_facade.py
```diff
@@ -356,6 +362,7 @@ def __init__(self,
             tae_runner = tae_runner(**tae_def_kwargs)
         elif callable(tae_runner):
             tae_def_kwargs['ta'] = tae_runner
+            tae_def_kwargs['use_pynisher'] = scenario.limit_resources  # TODO check where to add this param - not relevant to ExecuteTARunOld
```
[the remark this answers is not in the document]
ashraaghav: It is resolved. I will remove the TODO.

smac/facade/smac_ac_facade.py
```diff
@@ -75,6 +76,7 @@ def __init__(self,
                  runhistory: Optional[Union[Type[RunHistory], RunHistory]] = None,
                  runhistory_kwargs: Optional[dict] = None,
                  intensifier: Optional[Type[Intensifier]] = None,
+                 intensifier_type: Optional[str] = 'intensify',
```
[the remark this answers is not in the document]
ashraaghav: Ok, I'll change this. Would you recommend using bool types instead, for each new intensifier that is added?

smac/intensification/successive_halving.py
ashraaghav: In the current implementation, we remove configurations that have already been run at least once before starting the successive halving iterations. So we are not considering a configuration for the 2nd time in SH. 

Configurations being compared at this stage are only those that have run this given set of instance-seed pairs in the current budget. So `runs` would be the same for all configs. (I could/should take it out of the loop)

However, if cutoff is used as a budget, configurations will repeat i.e., on larger budgets, we will run a configuration on the same instance-seed pair again. But `RunHistory` overwrites the cost for that key in its dictionary, so the configurations stay comparable.

smac/utils/io/cmd_reader.py
```diff
@@ -546,7 +552,29 @@ def _add_smac_options(self):
                default=0.5, type=float,
                help="[dev] probablity to run a random configuration"
                " instead of configuration optimized on the acquisition function")
-
+        ## Successive Halving
+        smac_opts.add_argument("--sh-min-budget", "--sh_min_budget",
+                dest='sh_min_budget',
+                default=None, type=int,
```
AndreBiedenkapp: The help text says that the default is 1 but its actual default is set to None

smac/utils/io/cmd_reader.py
```diff
@@ -546,7 +552,29 @@ def _add_smac_options(self):
...
+                help="[dev] Set minimum budget for successive halving. Default is 1 (min. instances in 1st run)")
+        smac_opts.add_argument("--sh-max-budget", "--sh_max_budget",
+               dest='sh_max_budget',
+               default=None, type=int,
+               help="[dev] Set maximum budget for successive halving. Default is total instances")
```
AndreBiedenkapp: Maybe if you can't set the default value here you could change the help text to something like: "If not specified, i.e. None; all instances are used"

smac/utils/io/cmd_reader.py
```diff
@@ -546,7 +552,29 @@ def _add_smac_options(self):
...
+        smac_opts.add_argument("--sh-n-seeds", "--sh_n_seeds",
+               dest='sh_n_seeds',
+               default=None, type=int,
+               help="[dev] Set number of seeds to consider for successive halving run if algorithm is "
+                    "non-deterministic. Default is None")
+        smac_opts.add_argument("--sh-instance-order", "--sh_instance_order",
+               dest='sh_instance_order',
+               choices=["random", "budget_random", "none"],
+               default='random', type=str,
+               help="[dev] Set order of instances for successive halving run. Default is 'random'")
```
AndreBiedenkapp: I think it would be good if you specify what random and budget_random do.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+
+        # scenario info
+        self.deterministic = deterministic
+        self.run_obj_time = run_obj_time
+        self.tae_runner = tae_runner
+
+        # instance related info
+        if instances is None:
+            instances = []
+        self.instances = list(set(instances))
```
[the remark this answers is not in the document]
ashraaghav: There was a similar line in intensification.py, which I assume is to remove duplicates from the given list?

https://github.com/automl/SMAC3/blob/a8832fc34e6f19074f4d3eb82c8f4558637a0fd4/smac/intensification/intensification.py#L89

But yes, this is wrong. I should retain the order provided even when removing duplicates.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+        if instance_specifics is None:
+            self.instance_specifics = {}
+        else:
+            self.instance_specifics = instance_specifics
+
+        # determine instance order
+        self.instance_order = instance_order
+        if instance_order == 'random':
```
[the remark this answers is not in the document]
ashraaghav: `instance_order` parameter says in which order should SMAC consider instances. Can be set to:
* None - use as is given by the user
* random - shuffle once and use across all SH run (default)
* budget_random - shuffle before every SH run

I have added this to the docstring, so it should be clearer now.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+            # randomize once
+            np.random.shuffle(self.instances)
+
+        # set seed(s) for all SH runs
+        # - currently user gives the number of seeds to consider
+        self.n_seeds = n_seeds
+        if self.deterministic:
+            seeds = [0]
+        else:
+            if self.n_seeds is None:
```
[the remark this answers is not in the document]
ashraaghav: This is already being handed in the previous line, L98?

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+        if self.n_seeds is None or self.n_seeds == 1:
+            curr_challengers = set(curr_challengers) - set(run_history.get_all_configs())
+
+        # randomize instances per successive halving run, if user specifies
+        all_instances = self.instances
+        if self.instance_order == 'budget_random':
+            np.random.shuffle(all_instances)
+
+        # calculating the incumbent's performance for adaptive capping
+        # this check is required because there is no initial run to get the incumbent performance
```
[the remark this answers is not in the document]
ashraaghav: The initial design involves 1 full SH run with default + random configurations. Since it is essentially a `intensify()` call from there, that will not have an incumbent.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+        if incumbent is not None:
+            inc_runs = run_history.get_runs_for_config(incumbent)
+            inc_sum_cost = sum_cost(config=incumbent, instance_seed_pairs=inc_runs, run_history=run_history)
```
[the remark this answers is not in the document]
ashraaghav: If I am not wrong, `run_history.get_cost(incumbent)` will perform the aggregation that the user has given in the parameter `aggregate_func`. If that is an averaging or another aggregation method, then the cost computed will not be a sum.

This was done in `intensification.py` too. So I followed the same.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+                        # Returning the final incumbent selected so far because we ran out of optimization budget
+                        self.logger.debug("Budget exhausted; "
+                                          "Interrupting optimization run and returning current incumbent")
+                        return incumbent, None
+
+                    self._num_run += 1
+
+                if not first_run:
+                    first_run = True
+                    self._chall_indx += 1
```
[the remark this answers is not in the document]
ashraaghav: The challengers are repeated in the larger budget, so counting is performed only in the first budget iteration where all challengers are attempted. `_chall_idx` is later used only to update stats using `stats.update_average_configs_per_intensify()`, so it made sense only to count configs and not every time it was evaluated. It was similarly done in `intensification.py` too.

But thanks for pointing that out. There was a bug in that if-statement!

smac/optimizer/smbo.py
ashraaghav: It is handled inside the initial design now. I had to modify the initial design to accommodate it though.

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
AndreBiedenkapp: > This is already being handed in the previous line, L98?

Yes you are right. Sorry I overlooked that

examples/SMAC4HPO_rf.py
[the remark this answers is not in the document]
mlindauer: > Does this lead to edits in the SMAC-documentation pipeline?

Could be.

I think it is a very good idea to extend our examples, however please don't change the existing examples, but create new examples.

examples/SMAC4HPO_rf.py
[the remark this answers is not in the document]
mlindauer: I see your point. However, it looks more complicated than before and thus, it is more unlikely that new users will really read these. Please try to make examples short and sweet.

smac/initial_design/initial_design.py
```diff
@@ -199,15 +218,15 @@ def _transform_continuous_designs(self,
                 # add a vector with zeros
                 design_ = np.zeros(np.array(design.shape) + np.array((0, 1)))
                 design_[:, :idx] = design[:, :idx]
-                design_[:, idx+1:] = design[:, idx:]
+                design_[:, idx + 1:] = design[:, idx:]
                 design = design_
             elif isinstance(param, CategoricalHyperparameter):
                 v_design = design[:, idx]
-                v_design[v_design == 1] = 1 - 10**-10
+                v_design[v_design == 1] = 1 - 10 ** -10
```
mlindauer: that's a weird format and hard to read. please convert it back how it was.

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
mlindauer: Originally we wanted to support different kinds of aggregation functions, but now we have many code assumptions assuming that we always use `average_cost`. So, we don't support other aggregation functions anymore and you can safely assume average_cost

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,440 @@
...
+            stores all runs we ran so far
+        k: int
+            number of configurations to select
+
+        Returns
+        -------
+        challengers: typing.List[Configuration]
+            top challenger configurations, sorted in increasing costs
+        """
+        # NOTE run_history overwrites the cost for runs since it doesnt record each budget (cutoff) separately
```
mlindauer: I wonder whether this will hurt us quite a bit when we will combine HB with BO. Nothing to do for now, but we have to keep that in mind and have to change the runhistory at some point

ashraaghav: But in this section, we have to use `sum_cost` to compute the incumbent runtime for adaptive capping. The incumbent cost is calculated the same way in intensification.py too.

https://github.com/automl/SMAC3/blob/6f77eafaafe36d5c59c88689bfaea5d9505c3003/smac/intensification/intensification.py#L342-L348

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
mfeurer: Interesting, I believe this needs to be documented.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,440 @@
...
+        'halving' factor after each iteration in a successive halving run
+    run_obj_time : bool
+        whether the run objective is runtime or not (if true, apply adaptive capping)
+    n_seeds : int
+        Number of seeds to use, if TA is not deterministic. Defaults to None, i.e., seed is set as 0
+    instance_order : str
+        how to order instances. Can be set to:
+        None - use as is given by the user
+        random - shuffle once and use across all SH run (default)
+        budget_random - shuffle before every SH run
```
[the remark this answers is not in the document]
ashraaghav: Yes, that does make it more intuitive. I will update that.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,440 @@
...
+            raise ValueError("time_bound must be >= %f" % self._min_time)
+
+        if isinstance(challengers, ChallengerList):
+            challengers = challengers.challengers
+
+        # selecting first 'n' challengers for the current run, based on eta
+        curr_challengers = challengers[:self.initial_challengers]
+        # challengers can be repeated only if optimizing across multiple seeds
+        if self.n_seeds is None or self.n_seeds == 1:
+            curr_challengers = set(curr_challengers) - set(run_history.get_all_configs())
```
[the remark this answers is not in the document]
ashraaghav: Yes. That has been changed but I haven't pushed the changes yet.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,440 @@
...
+
+        if time_bound < self._min_time:
+            raise ValueError("time_bound must be >= %f" % self._min_time)
+
+        if isinstance(challengers, ChallengerList):
+            challengers = challengers.challengers
+
+        # selecting first 'n' challengers for the current run, based on eta
+        curr_challengers = challengers[:self.initial_challengers]
+        # challengers can be repeated only if optimizing across multiple seeds
```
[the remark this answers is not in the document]
ashraaghav: The current condition for a configuration to be executed a 2nd time is only if the TA is non-deterministic. But you do make a good point about the case where instance order is shuffled every SH run. I will include that in the condition too.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,440 @@
...
+        self.start_time = time.time()
+        self._ta_time = 0
+        self._chall_indx = 0
+        self._num_run = 0
+
+        if time_bound < self._min_time:
+            raise ValueError("time_bound must be >= %f" % self._min_time)
+
+        if isinstance(challengers, ChallengerList):
+            challengers = challengers.challengers
```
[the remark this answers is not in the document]
ashraaghav: It was intentional because it will be easier to parallelize in the future if we filter challengers before racing them. But yes, this approach does not interleave random configurations. Will fix it. :)

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,457 @@
...
+                                  "Interrupting optimization run and returning current incumbent")
+                inc_perf = run_history.get_cost(incumbent)
+                return incumbent, inc_perf
+
+            # selecting the top 'k' challengers for the next iteration
+            if next_n_chal > 0:
+                curr_challengers = self._top_k(curr_challengers, run_history, k=next_n_chal)
+            else:
+                # returning at least 1 configuration for next iteration since the incumbent has to run on all instances
+                self.logger.info('Ran out of configurations! Returning the best from what was seen so far.')
```
mfeurer: I find this message counterintuitive as it so far was only emitted after successive halving successfully raced all configurations. I think this message should only be emitted in case that the number of challengers is unequal to the natural number of things to compare against each other

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,457 @@
...
+            self.logger.info("incumbent configuration: %s" % str(best_challenger))
+            if log_traj:
+                # adding incumbent entry
+                self.stats.inc_changed += 1  # first incumbent
+                self.traj_logger.add_entry(train_perf=inc_perf,
+                                           incumbent_id=self.stats.inc_changed,
+                                           incumbent=new_incumbent)
+
+        else:
+            new_incumbent = self._compare_configs(incumbent, best_challenger,
```
mfeurer: I'm wondering how we would define incumbent best. I think only returning an incumbent after the successive halving procedure is completed is too conservative. Therefore, the question is whether the incumbent is defined as a configuration+budget pair or only as a configuration? Only in the latter case we want to only compare in the end, otherwise we want to compare it in between.

smac/optimizer/objective.py
```diff
@@ -35,8 +35,8 @@ def _runtime(config: Configuration, run_history: RunHistory,
         instance_seed_pairs = run_history.get_runs_for_config(config)
 
     runtimes = []
-    for i, r in instance_seed_pairs:
-        k = RunKey(id_, i, r)
+    for i, r, b in instance_seed_pairs:
```
mfeurer: These are definitively no instance-seed pairs any more, but rather instance-seed-budget pairs. I'm really not sure how to call these now, `instance_seed_budget_triplet`?

mfeurer: This issue obviously also applies to the whole runhistory.

examples/SMAC4HPO_mlp.py
```diff
@@ -0,0 +1,144 @@
...
+        seed: int or RandomState
+            used to initialize the rf's random generator
+        instance: str
+            used to represent the instance to use (just a placeholder for this example)
+        budget: float
+            used to set max iterations for the MLP
+
+        Returns:
+        -----------
+        np.mean(accuracy): float
```
mfeurer: That's not correct, you return `accuracy * -1`. How about you return the misclassification error (`1 - accuracy`)? I'd find that far more intuitive. Also, it is sufficient to state the type of the return value here.

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
mfeurer: This is better, thanks, but this will introduce unnecessary overhead (sampling random configurations costs time). Please only retrieve as many challengers as you actually require below.

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
mfeurer: Thanks, this is much better. However, could you please format this as a list (instead of not a paragraph)?

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
mfeurer: Thanks a lot. In addition to the changes you'll have to do to make this efficient (see comment above) you can then also simplify this and don't use the `islice` function.

smac/optimizer/smbo.py
```diff
@@ -32,6 +32,7 @@
 from smac.utils.validate import Validator
 from smac.utils.constants import MAXINT
 
+from smac.intensification.successive_halving import SuccessiveHalving
```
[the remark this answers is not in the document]
mfeurer: I just removed this line as the imported is actually not needed.

test/test_intensify/test_successive_halving.py
```diff
@@ -0,0 +1,240 @@
...
+
+        self.scen = Scenario({"cutoff_time": 2, 'cs': self.cs,
+                              "run_obj": 'runtime',
+                              "output_dir": ''})
+        self.stats = Stats(scenario=self.scen)
+        self.stats.start_timing()
+
+        self.logger = logging.getLogger(self.__module__ + "." + self.__class__.__name__)
+
+    def test_top_k(self):
```
mfeurer: I think this test needs to be updated. In case of instances, one needs to make sure that all configurations are compared on the same instances, and the instances must be only the ones which are so far chosen by successive halving. The current test setting should therefore actually fail.

test/test_intensify/test_successive_halving.py
```diff
@@ -0,0 +1,240 @@
...
+                    status=StatusType.SUCCESS, instance_id=1,
+                    seed=0, budget=3,
+                    additional_info=None)
+
+        inc, _ = intensifier.intensify(challengers=[self.config2],
+                                       incumbent=self.config1,
+                                       run_history=self.rh,
+                                       aggregate_func=average_cost)
+
+        self.assertEqual(inc, self.config2)
```
mfeurer: Could you please change all these tests to either check the full run history or mock out the comparison function and check the call arguments?

test/test_intensify/test_successive_halving.py
```diff
@@ -0,0 +1,240 @@
...
+    cs.add_hyperparameter(UniformIntegerHyperparameter(name='a',
+                                                       lower=0,
+                                                       upper=100))
+    cs.add_hyperparameter(UniformIntegerHyperparameter(name='b',
+                                                       lower=0,
+                                                       upper=100))
+    return cs
+
+
+class TestSuccessiveHalving(unittest.TestCase):
```
mfeurer: Please add a unit test for the init method as well.

test/test_intensify/test_intensify.py
```diff
@@ -18,7 +18,7 @@
 from smac.tae.execute_ta_run import StatusType
 from smac.stats.stats import Stats
 from smac.utils.io.traj_logging import TrajLogger
-
+from smac.utils.constants import MAX_CUTOFF
```
mfeurer: This appears to be an unused import.

smac/runhistory/runhistory.py
```diff
@@ -17,14 +17,24 @@
 __version__ = "0.0.1"
 
 
-RunKey = collections.namedtuple(
-    'RunKey', ['config_id', 'instance_id', 'seed'])
+# RunKey = collections.namedtuple(
```
mfeurer: Please don't leave commented code.

smac/runhistory/runhistory.py
```diff
@@ -17,14 +17,24 @@
...
+#     'RunKey', ['config_id', 'instance_id', 'seed', 'budget'])
+# NOTE class instead of collection to have a default value for budget in RunKey
+class RunKey(collections.namedtuple('RunKey', ['config_id', 'instance_id', 'seed', 'budget'])):
+    __slots__ = ()
+
+    def __new__(cls, config_id, instance_id, seed, budget=0.0):
+        return super().__new__(cls, config_id, instance_id, seed, budget)
 
 
 InstSeedKey = collections.namedtuple(
```
mfeurer: Is this one still necessary now that we also store the budget?

smac/runhistory/runhistory.py
```diff
@@ -114,7 +124,8 @@ def __init__(
 
         # for fast access, we have also an unordered data structure
         # to get all instance seed pairs of a configuration
-        self._configid_to_inst_seed = {}  # type: typing.Dict[int, InstSeedKey]
+        # self._configid_to_inst_seed = {}  # type: typing.Dict[int, InstSeedKey]
```
mfeurer: no comments here, too.

test/test_intensify/test_hyperband.py
```diff
@@ -0,0 +1,129 @@
...
+    cs.add_hyperparameter(UniformIntegerHyperparameter(name='a',
+                                                       lower=0,
+                                                       upper=100))
+    cs.add_hyperparameter(UniformIntegerHyperparameter(name='b',
+                                                       lower=0,
+                                                       upper=100))
+    return cs
+
+
+class TestHyperband(unittest.TestCase):
```
mfeurer: Could you please add some tests for the init method?

smac/runhistory/runhistory.py
ashraaghav: It isn't. But I thought I will let it be in case we need it again in the future.

test/test_intensify/test_hyperband.py
ashraaghav: I added the tests as a part of the intensify test cases because testing the parameter updates would mean running intensify separately for that.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,475 @@
...
+            # budget with cutoff
+            # cannot run successive halving with cutoff as budget if budget limits are not provided!
+            if min_budget is None or \
+                    (max_budget is None and self.cutoff is None):
+                raise ValueError("Successive Halving with runtime-cutoff as budget (i.e., only 1 instance) "
+                                 "requires parameters min_budget and max_budget/cutoff for intensification!")
+
+            # if both cutoff and max_budget are provided, then warn user
+            if self.cutoff is not None and max_budget is not None:
+                warnings.warn('Successive Halving with runtime-cutoff as budget: '
```
mfeurer: Could you please use the logger anstead of `warning.warn`?

smac/runhistory/runhistory.py
mfeurer: Please don't keep code because one could need it in the future. If we need it again we can retrieve it from the git history.

examples/SMAC4HPO_mlp.py
mfeurer: Please only state the return type (and remove the two following lines).

test/test_intensify/test_successive_halving.py
```diff
@@ -0,0 +1,313 @@
...
+        self.rh.add(config=self.config1, cost=1, time=1,
+                    status=StatusType.SUCCESS, instance_id=1,
+                    seed=None,
+                    additional_info=None)
+        self.rh.add(config=self.config2, cost=10, time=10,
+                    status=StatusType.SUCCESS, instance_id=2,
+                    seed=None,
+                    additional_info=None)
+
+        with self.assertRaises(AssertionError):
```
mfeurer: Could you please use `self.assertRaisesRegex` so that it is clear which error message is raised?

smac/intensification/hyperband.py
```diff
@@ -0,0 +1,169 @@
+import sys
+import time
+import logging
+import typing
+
+import numpy as np
+
+from smac.intensification.successive_halving import SuccessiveHalving
+from smac.stats.stats import Stats
+from smac.utils.constants import MAXINT, MAX_CUTOFF
```
mfeurer: This appears to be an unused import.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,475 @@
...
+                self.rs.shuffle(self.instances)
+
+            # set seed(s) for all SH runs
+            # - currently user gives the number of seeds to consider
+            if self.deterministic:
+                seeds = [0]
+            else:
+                seeds = self.rs.randint(low=0, high=MAXINT, size=self.n_seeds)
+            # storing instances & seeds as tuples
+            self.instances = [InstSeedKey(i, s) for s in seeds for i in self.instances]
```
mfeurer: This behavior is different than the one in the parent class, could you please make the two look alike?

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,475 @@
...
+        self.eta = eta
+
+        # BUDGETS
+
+        # - if only 1 instance was provided & quality objective, then use cutoff as budget
+        # - else, use instances as budget
+        if not self.run_obj_time and len(self.instances) <= 1:
+            # budget with cutoff
+            # cannot run successive halving with cutoff as budget if budget limits are not provided!
+            if min_budget is None or \
```
mfeurer: Above you define them as float, while here you also allow them to be `None`. Please update the method signature to reflect this.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,475 @@
...
+                              (self.max_budget, len(self.instances)))
+
+            self.min_budget = min_budget
+            self.max_budget = max_budget if max_budget else self.cutoff
+            self.cutoff_as_budget = True
+
+        else:
+            # budget with instances
+            if self.run_obj_time and len(self.instances) <= 1:
+                warnings.warn("Successive Halving has objective 'runtime' but only 1 instance-seed pair.")
```
mfeurer: Please use the logger here.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,475 @@
...
+            self.min_budget = 1 if min_budget is None else int(min_budget)
+            self.max_budget = len(self.instances) if max_budget is None else int(max_budget)
+            self.cutoff_as_budget = False
+
+            # max budget cannot be greater than number of instance-seed pairs
```
mfeurer: The same holds for min budget. Also min budget cannot be larger than max budget. Could you please add tests for that?

test/test_intensify/test_successive_halving.py
```diff
@@ -0,0 +1,313 @@
...
+            tae_runner=None, stats=self.stats,
+            traj_logger=TrajLogger(output_dir=None, stats=self.stats),
+            rng=np.random.RandomState(12345), deterministic=False, run_obj_time=False,
+            instances=[1, 2, 3], n_seeds=2, min_budget=None, max_budget=None, eta=2)
+
+        self.assertEqual(len(intensifier.instances), 6)  # since instance-seed pairs
+        self.assertEqual(intensifier.min_budget, 1)
+        self.assertEqual(intensifier.max_budget, 6)
+        self.assertEqual(intensifier.init_chal, 4)  # 2 iterations
+        self.assertEqual(intensifier.cutoff_as_budget, False)
```
mfeurer: Please use `self.assertFalse` here.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,475 @@
...
+            if self.max_budget > len(self.instances):
+                raise ValueError('Max budget cannot be greater than the number of instance-seed pairs')
+            if self.max_budget < len(self.instances):
+                warnings.warn('Max budget (%d) does not include all instance-seed pairs (%d)' %
+                              (self.max_budget, len(self.instances)))
+
+        self.logger.debug("Running Successive Halving with '%s' as budget" % self.cutoff_as_budget)
+
+        # number configurations to consider for a full successive halving iteration
+        self.max_sh_iter = np.floor(np.log(self.max_budget / self.min_budget) / np.log(eta))
```
mfeurer: Could you please add a reference to the line in Algorithm 1 of [Li et al.](http://jmlr.org/papers/volume18/16-558/16-558.pdf)? This would make it easier to read (I'm actually right now not sure which line this belongs to and which variable you are computing here).

mfeurer: Also, please add a comment which variable each of the variables here corresponds to in Algorithms 1 of [Li et al.](http://jmlr.org/papers/volume18/16-558/16-558.pdf).

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+        stats object
+    traj_logger: TrajLogger
+        TrajLogger object to log all new incumbents
+    rng : np.random.RandomState
+    instances : typing.List[str]
+        list of all instance ids
+    instance_specifics : typing.Mapping[str,np.ndarray]
+        mapping from instance name to instance specific string
+    cutoff : int
+        runtime cutoff of TA runs
```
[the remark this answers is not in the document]
mlindauer: Please fix this docstring. Simply drop "runtime"

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,475 @@
...
+from smac.runhistory.runhistory import RunHistory, InstSeedKey
+from smac.tae.execute_ta_run import BudgetExhaustedException, CappedRunException, ExecuteTARun
+from smac.utils.io.traj_logging import TrajLogger
+
+__author__ = "Ashwin Raaghav Narayanan"
+__copyright__ = "Copyright 2018, ML4AAD"
+__license__ = "3-clause BSD"
+
+
+class SuccessiveHalving(Intensifier):
```
mlindauer: `Intensifier` might not be a good base class to inherit from. Please create an Abstract Racing class and let `Intensifier` and `SuccessiveHalving` inherit from that

mlindauer:
I looked over all my previous comments again and resolved all of them (as long as there was a button for "resolve conversation". I only discovered that the inheritance between Intensifier and SH might be sub-optimal and thus I would recommend a refactoring. But this could also be done in a further PR s.t. this PR does not get even larger. Therefore, I will approve the PR.

smac/intensification/successive_halving.py
ashraaghav: Is it ok if I'll leave this open for now, since Marius had also mentioned we might have to redesign the inheritance of intensifiers?

mfeurer: If you create a new issue with open TODOs inherited from this PR, then yes.

smac/intensification/successive_halving.py
ashraaghav: The 'max_sh_iter' variable is just a precomputed variable used for sampling configurations when SuccessiveHalving is used independently.
If configurations come from hyperband, then this variable is not used.

smac/runhistory/runhistory.py
ashraaghav: Apologies, I was wrong. We still use in another variable where we map configs, inst-seeds to budget.

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
mfeurer: This is not addressed yet!

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
mfeurer: This is not addressed yet!

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,467 @@
...
+        Number of seeds to use, if TA is not deterministic. Defaults to None, i.e., seed is set as 0
+    instance_order : str
+        how to order instances. Can be set to:
+        None - use as is given by the user
+        shuffle_once - shuffle once and use across all SH run (default)
+        shuffle - shuffle before every SH run
+    adaptive_capping_slackfactor : float
+        slack factor of adpative capping (factor * adpative cutoff)
+
+    Returns
```
mfeurer: The init actually hasn't any return values. Could you please drop those?

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,440 @@
...
+        aggregate_func: typing.Callable
+            aggregate error across instances
+        time_bound : float, optional (default=2 ** 31 - 1)
+            time in [sec] available to perform intensify
+        log_traj: bool
+            whether to log changes of incumbents in trajectory
+
+        Returns
+        -------
+        incumbent: Configuration()
```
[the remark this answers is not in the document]
mfeurer: This is not addressed yet!

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,440 @@
...
+        else:
+            self.adaptive_capping = False
+        self.adaptive_capping_slackfactor = adaptive_capping_slackfactor
+
+    def intensify(self, challengers: typing.List[Configuration],
+                  incumbent: typing.Optional[Configuration],
+                  run_history: RunHistory,
+                  aggregate_func: typing.Callable,
+                  time_bound: float = float(MAXINT),
+                  log_traj: bool = True):
```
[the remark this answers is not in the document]
mfeurer: This is not addressed yet!

smac/intensification/successive_halving.py
ashraaghav: The implementation was based on BOHB paper and hpbandster package. I'll add these as reference.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+        # adaptive capping
+        if not self.cutoff_as_budget and self.instance_order != 'shuffle' and self.run_obj_time:
+            self.adaptive_capping = True
+        else:
+            self.adaptive_capping = False
+
+    def _init_sh_params(self, min_budget: float,
+                        max_budget: typing.Union[float, None],
+                        eta: typing.Union[float, None],
+                        init_chal: int):
```
mfeurer: Could you please add a return type here?

smac/intensification/successive_halving.py
mfeurer: This will be part of a later pull request

smac/runhistory/runhistory.py
```diff
@@ -114,7 +122,7 @@ def __init__(
 
         # for fast access, we have also an unordered data structure
         # to get all instance seed pairs of a configuration
-        self._configid_to_inst_seed = {}  # type: typing.Dict[int, InstSeedKey]
+        self._configid_to_inst_seed_budget = {}  # type: typing.Dict[int, typing.Dict[InstSeedKey, typing.List[float]]]
```
mfeurer: Isn't the type annotation wrong now?

smac/intensification/successive_halving.py
mfeurer: Okay, I have noted this.

smac/intensification/hyperband.py
```diff
@@ -0,0 +1,167 @@
...
+        slack factor of adpative capping (factor * adpative cutoff)
+    """
+
+    def __init__(self, tae_runner: ExecuteTARun,
+                 stats: Stats,
+                 traj_logger: TrajLogger,
+                 rng: np.random.RandomState,
+                 instances: typing.List[str],
+                 instance_specifics: typing.Mapping[str, np.ndarray] = None,
+                 cutoff: int = None,
```
mfeurer: Please update all the signatures here to reflect that they allow `None` by annotating them with `typing.Optional[float]` etc...

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+        self._init_sh_params(min_budget, max_budget, eta, init_chal)
+
+        # adaptive capping
+        if not self.cutoff_as_budget and self.instance_order != 'shuffle' and self.run_obj_time:
+            self.adaptive_capping = True
+        else:
+            self.adaptive_capping = False
+
+    def _init_sh_params(self, min_budget: float,
+                        max_budget: typing.Union[float, None],
```
mfeurer: Please simplify to `typing.Optional[float]`.

smac/intensification/successive_halving.py
mfeurer: This will be fixed in a future PR.

test/test_initial_design/test_single_config_initial_design.py
```diff
@@ -81,6 +81,46 @@ def test_multi_config_design(self):
...
+            configs=configs,
+            intensifier=intensifier,
+            aggregate_func=average_cost,
+            n_configs_x_params=2,
+            fill_random_configs=True,
+        )
+
+        inc = dc.run()
+        self.assertTrue(stats.ta_runs == 2)  # two runs per config
+        self.assertTrue(len(rh.data) == 2)  # two runs per config
```
mfeurer: But how many configs in total? Could you please add a check for that, too?

smac/facade/smac_ac_facade.py
```diff
@@ -222,6 +223,11 @@ def __init__(self,
 
         # initialize empty runhistory
         runhistory_def_kwargs = {'aggregate_func': aggregate_func}
+        if intensifier is not None and isinstance(intensifier, SuccessiveHalving):
+            # if Successive Halving / Hyperband is used, then RunHistory overwrites data for
+            # same instance-seed pairs (to save runs of higher budgets)
+            # - will only be used when cutoff is used as a budget
```
mfeurer: Could you please add an explanation why? I currently don't get this.

smac/intensification/hyperband.py
```diff
@@ -0,0 +1,167 @@
...
+        # setting initial running budget for future iterations (s & s_max from Algorithm 1)
+        self.s_max = np.floor(np.log(self.max_budget / self.min_budget) / np.log(self.eta))
+        self.s = np.floor(np.log(self.max_budget / self.min_budget) / np.log(self.eta))
+
+    def intensify(self, challengers: typing.List[Configuration],
+                  incumbent: typing.Optional[Configuration],
+                  run_history: RunHistory,
+                  aggregate_func: typing.Callable,
+                  time_bound: float = float(MAXINT),
+                  log_traj: bool = True):
```
mfeurer: Could you please add the return types here?

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,462 @@
...
+    rng : np.random.RandomState
+    instances : typing.List[str]
+        list of all instance ids
+    instance_specifics : typing.Mapping[str,np.ndarray]
+        mapping from instance name to instance specific string
+    cutoff : typing.Union[int, None]
+        cutoff of TA runs
+    deterministic : bool
+        whether the TA is deterministic or not
+    min_budget : typing.Union[float, None]
```
mfeurer: Could you please rename this to `initial_budget` as it is called in Algorithm 1 of the Appendix of Falkner et al.?

smac/intensification/successive_halving.py
mfeurer: Is this `s_max` from Algorithm 1 of the BOHB paper?

mfeurer: Could you please add a comment to each of the three variables you pre-compute what they're doing exactly?

smac/intensification/successive_halving.py
ashraaghav: In "hpbandster" implementations they had named it as `min_budget`. So I thought it would be easier to relate between the two implementations this way. If that is not needed, I'll change it.

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,464 @@
...
+
+            self.logger.info('Running with budget [%.2f / %d] with %d challengers' %
+                             (curr_budget, self.max_budget, len(curr_challengers)))
+            # selecting instance subset for this budget, depending on the kind of budget
+            prev_budget = self.budgets[i - 1] if i > 0 else 0
+            available_insts = all_instances[int(prev_budget):int(curr_budget)] if not self.cutoff_as_budget \
+                else all_instances
+
+            # determine 'k' for the next iteration - at least 1
+            next_n_chal = max(1, int(np.round(len(curr_challengers) / self.eta)))
```
mfeurer: Could you please use `np.floor` instead of `np.round` to make the code closer to the pseudo-code in the paper?

test/test_intensify/test_successive_halving.py
```diff
@@ -0,0 +1,347 @@
...
+        """
+
+        with self.assertRaisesRegex(ValueError, 'eta must be greater than 1'):
+            SuccessiveHalving(
+                tae_runner=None, stats=self.stats,
+                traj_logger=TrajLogger(output_dir=None, stats=self.stats),
+                rng=np.random.RandomState(12345), deterministic=True, run_obj_time=False,
+                cutoff=10, instances=[1], eta=0)
+
+    def test_init_5(self):
```
mfeurer: What's the difference to the test above?

test/test_intensify/test_successive_halving.py
```diff
@@ -0,0 +1,347 @@
...
+            instances=[1, 2], initial_budget=1, max_budget=2, eta=2, instance_order=None)
+
+        self.rh.add(config=self.config1, cost=.001, time=0.001,
+                    status=StatusType.SUCCESS, instance_id=1, seed=0,
+                    additional_info=None)
+        self.rh.add(config=self.config1, cost=.001, time=0.001,
+                    status=StatusType.SUCCESS, instance_id=2, seed=0,
+                    additional_info=None)
+
+        # config2 should be capped and config1 should still be the incumbent
```
mfeurer: What about config3, should it also be capped?

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,464 @@
...
+            self.max_budget = len(self.instances) if max_budget is None else int(max_budget)
+            self.cutoff_as_budget = False
+
+            if self.max_budget > len(self.instances):
+                raise ValueError('Max budget cannot be greater than the number of instance-seed pairs')
+            if self.max_budget < len(self.instances):
+                self.logger.warning('Max budget (%d) does not include all instance-seed pairs (%d)' %
+                                    (self.max_budget, len(self.instances)))
+
+        self.logger.debug("Running Successive Halving with '%s' as budget" % self.cutoff_as_budget)
```
mfeurer: Could you the initial budget and the max budget to the log message?

mfeurer: Could you please also add the decision process on how to race to the docstring of this class? Otherwise, I believe that it will be hard for users to understand what's happening.

smac/runhistory/runhistory.py
```diff
@@ -282,24 +296,33 @@ def get_cost(self, config: Configuration):
         config_id = self.config_ids[config]
         return self.cost_per_config.get(config_id, np.nan)
 
-    def get_runs_for_config(self, config: Configuration):
+    def get_runs_for_config(self, config: Configuration, max_budget: bool = True):
```
mfeurer: Please write a unit test for this new flag.

smac/runhistory/runhistory.py
```diff
@@ -282,24 +296,33 @@ def get_cost(self, config: Configuration):
...
+        # select only the max budget run if specified
+        if max_budget:
+            for k, v in runs.items():
+                runs[k] = [max(v)]
+
+        # convert to inst-seed-budget key
+        runs = [InstSeedBudgetKey(k.instance, k.seed, budget) for k, v in runs.items() for budget in v]
+        return runs
 
     def get_instance_costs_for_config(self, config: Configuration):
```
mfeurer: Shouldn't this also get a flag to only return the cost on the maximum budget?

smac/intensification/successive_halving.py
```diff
@@ -0,0 +1,464 @@
...
+
+        # BUDGETS
+
+        if max_budget is not None and initial_budget is not None \
+                and max_budget < initial_budget:
+            raise ValueError('Max budget has to be larger than min budget')
+
+        # - if only 1 instance was provided & quality objective, then use cutoff as budget
+        # - else, use instances as budget
+        if not self.run_obj_time and len(self.instances) <= 1:
```
mfeurer: I believe there is an issue here. If the user wants to optimize a non-deterministic algorithm for multiple budgets, this is accidentally sorted into the multi-instance case. I actually don't know what to do in this situation. @mlindauer any suggestions? Until we have a solution for this we should maybe simply catch this case and raise an exception.

examples/spear_qcp/SMAC4AC_spear_qcp.py
```diff
@@ -0,0 +1,33 @@
...
+                   intensifier_kwargs=intensifier_kwargs,  # parameters for Successive Halving
+                   intensifier=SuccessiveHalving)  # change intensifier to successive halving by passing the class
+
+    # Start optimization
+    try:
+        incumbent = smac.optimize()
+    finally:
+        incumbent = smac.solver.incumbent
+
+    # inc_value = smac.get_tae_runner().run(incumbent)[1]
```
mfeurer: Please remove this line.

test/test_intensify/test_successive_halving.py
ashraaghav: Yes. I wanted to make it another test case for wrong initializations. I will also combine these similar cases into a single suite.

test/test_intensify/test_successive_halving.py
ashraaghav: Yes. That should be too. I'll mention that.

smac/runhistory/runhistory.py
ashraaghav: I'm not sure if that will be right, because we cannot combine the costs of different (runtime) budgets into 1. So this by default takes only the max budget. Should I give an option of returning instance costs for each instance-budget combination?

test/test_runhistory/test_runhistory.py
```diff
@@ -74,32 +74,54 @@ def test_get_config_runs(self):
...
+               instance_id=1, seed=1, budget=2)
+
+        rh.add(config=config2, cost=5, time=10, status=StatusType.SUCCESS,
+               instance_id=1, seed=1, budget=1)
+        rh.add(config=config2, cost=10, time=20, status=StatusType.SUCCESS,
+               instance_id=1, seed=1, budget=2)
+
+        ist = rh.get_runs_for_config(config=config1, max_budget=False)
+
+        self.assertEqual(len(ist), 2)
```
mfeurer: Shouldn't this be four (two budgets for two configs)?

smac/intensification/successive_halving.py
mfeurer: This isn't resolved yet. Could you please add an `elif` here and fail the optimization?

smac/runhistory/runhistory.py
mfeurer: Ah, because the `self.get_runs_for_config` only returns the cost on the highest budget, right?

test/test_runhistory/test_runhistory.py
ashraaghav: `get_runs_for_config` returns all runs for a single configuration. So this will have 2 budgets of 1 config.

mfeurer: Sorry for this stupid question...

smac/runhistory/runhistory.py
ashraaghav: Yes, that is the default behaviour.

smac/intensification/successive_halving.py
ashraaghav: Adding that as an `elif` was not feasible because the instances would have already been processed by then. I check for this case in `__init__`. I have also added a "NOTE" for reference.

PR #516, opened 2019-06-25
Reduce requirements

gsakkis:
Currently SMAC3 includes more requirements than necessary:
1. Dev only, e.g. nose, sphinx
2. Not (directly) used, e.g. statsmodels, cython
3. Required only for specific features, e.g. george, emcee

This PR drops (1) and (2) from requirements.txt and changes (3) to ["extra"](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) dependencies that can be installed optionally.


smac/facade/bogp_facade.py
```diff
@@ -38,7 +39,12 @@ def __init__(self, model_type='gp_mcmc', **kwargs):
         """
         scenario = kwargs['scenario']
         if scenario.initial_incumbent not in ['LHD', 'FACTORIAL', 'SOBOL']:
-            scenario.initial_incumbent = 'SOBOL'
+            if 'sobol' in smac.extras_installed:
```
mlindauer: The behavior of SMAC should not depend on which packages are installed. We assume that all required packages are installed.

mlindauer: same applies to several other changes of the code

gsakkis:
@mlindauer thanks for the quick feedback! No rush about waiting for other contributors to chime in.

smac/facade/bogp_facade.py
gsakkis: As a general comment, SMAC already has an (unofficial?) split between required and optional packages via the `__MANDATORY_PACKAGES__` definition and the check that these (and only these) are installed at runtime.

As for the specific change here, I assumed that the setting to `SOBOL` here was a reasonable default value but all other available options are also fine. If that's not true or we want to be stricter for backwards compatibility reasons, a runtime exception or warning are easy to add.

KEggensperger:
Hi @gsakkis, thanks a lot for the PR. Making use of _extra requirements_ that are not required for the default is great and would simplify installation for users quite a bit!

However, I do have two remarks:

1) I agree with @mlindauer that the default behavior should not depend on which packages are installed. I would propose to make `sobol` a general dependency and raise an exception if GPs are not installed when using the GP interface.
2) Could you rebase the PR to the dev branch? I would hope that there are no conflicts.

gsakkis:
@KEggensperger I rebased the PR to the development branch (there were a couple conflicts as apparently several things have changed) and made `sobol` a required dependency as suggested.

PR #517, opened 2019-07-11
Update README.md

abhaikollara:
Use `pip install -r` to install requirements


mlindauer:
The reason why we used such a complicated statement is actually to ensure that the dependencies are installed in the correct order.

PR #521, opened 2019-08-06
Fix GPMCMC normalization of y values

mfeurer:
This PR unifies the normalization of y values for GPMCMC. y values are now normalized only on the level of the base class instead of sometimes by sklearn and sometimes by the GPML/GPMAP class (which implement different ways of normalizing the data).

### DIFF
--- a/smac/epm/gaussian_process_mcmc.py
+++ b/smac/epm/gaussian_process_mcmc.py
@@ -122,7 +122,7 @@ class GaussianProcessMCMC(BaseModel):
 
         self.gp = GaussianProcessRegressor(
             kernel=self.kernel,
-            normalize_y=self.normalize_y,
+            normalize_y=False,
             optimizer=None,
             n_restarts_optimizer=-1,  # Do not use scikit-learn's optimization routine
             alpha=0,  # Governed by the kernel
--- a/smac/epm/gaussian_process_mcmc.py
+++ b/smac/epm/gaussian_process_mcmc.py
@@ -117,6 +117,9 @@ class GaussianProcessMCMC(BaseModel):
             hyperparameter specified in the kernel.
         """
         X = self._impute_inactive(X)
+        if self.normalize_y:
+            y = self._normalize_y(y)
+
         self.gp = GaussianProcessRegressor(
             kernel=self.kernel,
             normalize_y=self.normalize_y,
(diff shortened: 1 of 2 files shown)


mfeurer:
I can debug the failing unit test for 3.5 once the tests for 3.6 and 3.7 are fixed.

smac/epm/gaussian_process_mcmc.py
```diff
@@ -215,6 +218,11 @@ def _train(self, X: np.ndarray, y: np.ndarray, do_optimize: bool=True):
         self.models = []
         for sample in self.hypers:
 
+            if (sample < -50).any():
```
KEggensperger: Where do these numbers come from? 
Also, this should work without the if statements

smac/epm/gaussian_process_mcmc.py
```diff
@@ -236,15 +244,22 @@ def _train(self, X: np.ndarray, y: np.ndarray, do_optimize: bool=True):
...
                 kernel=kernel,
-                normalize_y=self.normalize_y,
+                normalize_y=False,
                 seed=self.rng.randint(low=0, high=10000),
             )
             model._train(X, y, do_optimize=False)
             self.models.append(model)
 
+        if self.normalize_y:
+            for model in self.models:
```
KEggensperger: Don't get this. First you are normalizing outside of model, now the mode itself should normalize. Why?

KEggensperger: Also is this now the same as for the other GP class?

smac/epm/gaussian_process_mcmc.py
mfeurer: > Where do these numbers come from?
RoBO used 20 here, but I found that the code also works with 50, so I thought I generalize this a bit.
> Also, this should work without the if statements
Yes, but I found it to be faster like this.

smac/epm/gaussian_process_mcmc.py
mfeurer: > First you are normalizing outside of model
This makes sure that the data is exactly the same when passed to scikit-learn to compute the likelihoods, and when passed to the SMAC GP class.
> now the mode itself should normalize
No, the model is already trained at this point in time. This basically makes sure that the base models denormalize the data themselves, and that this is not done by the GPMCMC class.

If that makes sense I'll add a few more comments to avoid such confusion in the future.

PR #522, opened 2019-08-07
Development

ricky151192:
Hi,
I added the probabilistic smv-C model in LCB function, creating a new acquisition function named "probCLCB".
I put two examples in "./examples" directory one that uses a RF as surrogate model and the other one using GP as surrogate.
Thanks,
Riccardo


mlindauer:
By now that PR is unfortunately quite old, the tests fail and we have conflicting files. If no one takes care of that again, I think we should close the PR.

PR #525, opened 2019-08-13
Improve CI

mfeurer:
Improve CI by requiring code changes to adhere to PEP8. I see this as a first stop towards uniformly formatted code.


mfeurer:
The output of the newly added script is now visible in the succeeding unit tests: https://travis-ci.org/automl/SMAC3/jobs/571394398

mlindauer:
How strict is PEP8 here? I think we agreed at some point that we allow for some more characters per line.

mfeurer:
Good that you're asking. I already catered for that and this change allows for 120 lines: https://github.com/automl/SMAC3/pull/525/files#diff-8572f4480db0a9c3d9b26e97c77de7d1R129

mlindauer:
Okay, I approved it. Nevertheless, I hope I won't be too annoyed by it in the future ;-)

mfeurer:
> I hope I won't be too annoyed by it in the future ;-)

Me too, but for OpenML, this works really well.

PR #529, opened 2019-09-11
update Gaussian process hyperparameters

mfeurer:
by using the hyperparameters suggested by https://arxiv.org/pdf/1908.06674.pdf

This also removes unnecessary print statements.

### DIFF
--- a/smac/facade/smac_bo_facade.py
+++ b/smac/facade/smac_bo_facade.py
@@ -5,7 +5,7 @@ from smac.epm.gaussian_process_mcmc import GaussianProcessMCMC, GaussianProcess
 from smac.epm.gp_base_prior import HorseshoePrior, LognormalPrior
 from smac.epm.util_funcs import get_types, get_rng
 from smac.initial_design.sobol_design import SobolDesign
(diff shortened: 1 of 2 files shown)


smac/facade/smac_bo_facade.py
```diff
@@ -20,6 +20,10 @@ class SMAC4BO(SMAC4AC):
     see smac.facade.smac_Facade for API
     This facade overwrites options available via the SMAC facade
 
+    Hyperparameters are chosen according to the best configuration for Gaussian process maximum likelihood found in
+    "Towards Assessing the Impact of Bayesian Optimization's Own Hyperparameters" by Lindauer et al., presented at the
+    DSO workshop 2019 (https://arxiv.org/abs/1908.06674).
+
```
KEggensperger: Could you briefly add which hyperparameters were chosen and where in that paper one would find the table?

mfeurer: Unfortunately, there is no table in the paper itself, I got the table via Slack. The only thing which resembles the information is inside the ablation plots. I could reference them and add the changed hyperparameters to the docstring. Would that be sufficient? I would directly commit those to the development branch if that is okay.

Issue #437, opened 2018-05-09
SMAC does not gracefully handle exceptions if `use_pynisher=False`


mfeurer:
```
/home/feurerm/projects/2018_nips_working_paper/2018_nips_working_paper/metalearning/greedy_portfolio.py:683: RuntimeWarning: invalid value encountered in greater_equal
  np.all((normalized_matrix >= 0) | (~np.isfinite(normalized_matrix)))
/home/feurerm/projects/2018_nips_working_paper/2018_nips_working_paper/metalearning/greedy_portfolio.py:684: RuntimeWarning: invalid value encountered in less_equal
  and np.all((normalized_matrix <= 1) | (~np.isfinite(normalized_matrix)))
/home/feurerm/projects/2018_nips_working_paper/2018_nips_working_paper/metalearning/resampling_strategies.py:233: RuntimeWarning: invalid value encountered in less
  loc = len(lc_runtime[(lc_runtime < cutoff) == True]) - 1
Traceback (most recent call last):
  File "/home/feurerm/environments/2018_nips_working_paper/envs/2018_nips_working_paper/lib/python3.6/site-packages/smac/intensification/intensification.py", line 162, in intensify
    self._add_inc_run(incumbent=incumbent, run_history=run_history)
  File "/home/feurerm/environments/2018_nips_working_paper/envs/2018_nips_working_paper/lib/python3.6/site-packages/smac/intensification/intensification.py", line 263, in _add_inc_run
    instance_specific=self.instance_specifics.get(next_instance, "0"))
  File "/home/feurerm/environments/2018_nips_working_paper/envs/2018_nips_working_paper/lib/python3.6/site-packages/smac/tae/execute_ta_run.py", line 154, in start
    "Skip target algorithm run due to exhausted "
smac.tae.execute_ta_run.BudgetExhaustedException: Skip target algorithm run due to exhausted configuration budget

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/feurerm/mhome/projects/2018_nips_working_paper/2018_nips_working_paper/scripts/run_metalearning.py", line 133, in <module>
    automl=automl,
  File "/home/feurerm/projects/2018_nips_working_paper/2018_nips_working_paper/metalearning/util.py", line 177, in evaluate_performance
    portfolio_size=portfolio_size)
  File "/home/feurerm/projects/2018_nips_working_paper/2018_nips_working_paper/metalearning/global_portfolio.py", line 522, in build
    losses=losses,
[...]

PR #534, opened 2019-09-18
Execute func issues

ashraaghav:
Fixed issues on `exec_func.py` 
* #437
* #453

### DIFF
--- a/smac/tae/execute_func.py
+++ b/smac/tae/execute_func.py
@@ -165,7 +165,7 @@ class AbstractTAFunc(ExecuteTARun):
             runtime = float(obj.wall_clock_time)
         else:
             start_time = time.time()
-            result = self.ta(config, **obj_kwargs)
+            result = self._call_ta_raw(config, **obj_kwargs)
             
             if result is not None:
                 status = StatusType.SUCCESS
@@ -185,6 +185,14 @@ class AbstractTAFunc(ExecuteTARun):
 
     def _call_ta(self, obj, config, instance, seed):
         raise NotImplementedError()
+
+    def _call_ta_raw(self, config, **kwargs):
+        try:
+            result = self.ta(config, **kwargs)
+        except Exception as e:
+            self.logger.exception(e)
+            result = None
+        return result
 
     def serialize(self, x):
         """ to serialize results from the TA if required """
--- a/smac/tae/execute_func.py
+++ b/smac/tae/execute_func.py
@@ -2,6 +2,7 @@ import logging
 import inspect
 import math
 import time
+import json
 
 import numpy as np
 import pynisher
@@ -76,6 +77,9 @@ class AbstractTAFunc(ExecuteTARun):
         
         self.use_pynisher = use_pynisher
 
+        self.logger = logging.getLogger(
+            self.__module__ + '.' + self.__class__.__name__)
+
     def run(self, config, instance=None,
             cutoff=None,
             seed=12345,
@@ -143,6 +147,7 @@ class AbstractTAFunc(ExecuteTARun):
             else:
                 result = rval
                 additional_run_info = {}
+
     
             if obj.exit_status is pynisher.TimeoutException:
                 status = StatusType.TIMEOUT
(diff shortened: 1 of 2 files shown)


smac/tae/execute_func.py
```diff
@@ -172,11 +177,37 @@ def run(self, config, instance=None,
...
         return status, cost, runtime, additional_run_info
 
     def _call_ta(self, obj, config, instance, seed):
         raise NotImplementedError()
 
+    def _call_ta_raw(self, config, **kwargs):
+        try:
+            result = self.ta(config, **kwargs)
+        except Exception as e:
+            self.logger.exception(e)
```
mfeurer: Cool, I didn't know about this one!

smac/tae/execute_func.py
```diff
@@ -160,7 +165,7 @@ def run(self, config, instance=None,
             runtime = float(obj.wall_clock_time)
         else:
             start_time = time.time()
-            result = self.ta(config, **obj_kwargs)
+            result = self._call_ta_raw(config, **obj_kwargs)
```
mfeurer: Could you please make the two ways to call this more similar by also allowing additional run information if the pynisher is not used?

smac/tae/execute_func.py
```diff
@@ -172,11 +177,37 @@ def run(self, config, instance=None,
...
+        # check serializability of results - if not, then attempt to convert it
+        cost = self.serialize(cost)
+        additional_run_info = self.serialize(additional_run_info)
+
         return status, cost, runtime, additional_run_info
 
     def _call_ta(self, obj, config, instance, seed):
         raise NotImplementedError()
 
+    def _call_ta_raw(self, config, **kwargs):
```
mfeurer: I think this function needs to go to the two subclasses to allow both the dict and the list-like version to work.

test/test_tae/test_exec_func.py
```diff
@@ -76,12 +79,12 @@ def test_run_wo_pynisher(self):
         self.assertGreaterEqual(rval[2], 0.0)
         self.assertEqual(rval[3], dict())
 
-
     @unittest.mock.patch.object(Configuration, 'get_dictionary')
     def test_run_execute_func_for_fmin(self, mock):
+        def target(x):
```
mfeurer: Please add a test without the pynisher here, too.

smac/tae/execute_func.py
```diff
@@ -172,11 +177,37 @@ def run(self, config, instance=None,
             runtime = time.time() - start_time
             additional_run_info = {}
 
+        # check serializability of results - if not, then attempt to convert it
```
mfeurer: I don't think we want to convert data without communicating this to the user. Also, I think the user should convert the data herself to know about the consequences. Could you please remove this functionality?

Also, please only try to serialize it only to check if it is possible. The current implementation creates a json, which is than later on formatted as a json.

Finally, the current implementation wouldn't note which value cannot be serialized. Could you please make sure that this is part of the error message?

smac/tae/execute_func.py
ashraaghav: This function is not needed actually. The `_call_ta` method can be reused here. I'll update it.

smac/tae/execute_func.py
```diff
@@ -170,7 +187,17 @@ def run(self, config, instance=None,
                 cost = self.crash_cost
             
             runtime = time.time() - start_time
-            additional_run_info = {}
+
+        # check serializability of results
+        try:
+            json.dumps(cost)
+        except TypeError:
+            raise TypeError("TA returned result of type %s but it should be a serializable type." % type(cost))
```
mfeurer: It would be great if you could

1. print the actual return value as well
2. explicitly note that it must be a type serializable by json and ask the user to convert it.

smac/tae/execute_func.py
```diff
@@ -170,7 +187,17 @@ def run(self, config, instance=None,
...
+        try:
+            json.dumps(additional_run_info)
+        except TypeError:
+            raise TypeError("TA returned 'additional_run_info' with some non-serializable items. "
+                            "Please ensure all objects returned are serializable.")
```
mfeurer: Could you please print the output here, too? Also, could you please somehow add the original error message so that the users sees what kind of object cannot be serialized?

Also, please do not abbreviate the word TA here.

test/test_tae/test_exec_func.py
```diff
@@ -76,12 +103,23 @@ def test_run_wo_pynisher(self):
         self.assertGreaterEqual(rval[2], 0.0)
         self.assertEqual(rval[3], dict())
 
+        def target(x):
```
mfeurer: This test is now duplicated, isn't it?

test/test_tae/test_exec_func.py
```diff
@@ -53,10 +56,22 @@ def test_run(self):
         self.assertEqual(rval[1], 4)
         self.assertGreaterEqual(rval[2], 0.0)
         self.assertEqual(rval[3], {'key': 12345, 'instance': 'test'})
-        
-    
+
```
mfeurer: This test should be in the function `test_run_wo_pynisher`, right? It appears that it is already there, so this is a duplicate, isn't it?

ashraaghav: That is a mistake. It is not supposed to have `use_pynisher` set to false. The reason why I added this test case to both was to test for an exception in the TA in both with & without pynisher runs.

smac/tae/execute_func.py
mfeurer: Sorry for not being not clear enough here. Issue #453 shows that SMAC complains with the error message `TypeError: Object of type 'int64' is not JSON serializable`. Could you please parse the message out of the exception object and print that, too? Then, the user would know which object cannot be serialized.

PR #542, opened 2019-09-30
Hyperband documentation

ashraaghav:
- [x] Documenting budgets and instances
- [x] 2nd example with instance (no. of folds) as budget in MLP example
- [ ] Rename cutoff to budget


examples/SMAC4HPO_hyperband_mlp.py
```diff
@@ -104,9 +107,9 @@ def mlp_from_cfg(cfg, seed, instance, budget, **kwargs):
...
 # SMAC scenario object
-scenario = Scenario({"run_obj": "quality",     # we optimize quality (alternative runtime)
-                     "wallclock-limit": 100,   # max duration to run the optimization (in seconds)
-                     "cs": cs,                 # configuration space
+scenario = Scenario({"run_obj": "quality",      # we optimize quality (alternative runtime)
+                     "wallclock-limit": 100,    # max duration to run the optimization (in seconds)
+                     "cs": cs,                  # configuration space
                      "deterministic": "true",
                      "limit_resources": False,  # Disables pynisher to pass cutoff directly to the target algorithm.
                                                 # Timeouts have to be taken care within the TA
```
mfeurer: Doesn't this contradict the logger statement above saying `"Running MLP example for SMAC. If you experience difficulties, try to decrease the memory-limit."`?

examples/SMAC4HPO_hyperband_mlp_instances.py
```diff
@@ -0,0 +1,156 @@
...
+
+# Defining instances (number of folds for CV)
+instances = [['2'], ['3'], ['5'], ['10']]
+
+# SMAC scenario object
+scenario = Scenario({"run_obj": "quality",      # we optimize quality (alternative runtime)
+                     "wallclock-limit": 100,    # max duration to run the optimization (in seconds)
+                     "cs": cs,                  # configuration space
+                     "deterministic": "true",
+                     "limit_resources": False,  # Disables pynisher to pass cutoff directly to the target algorithm.
```
mfeurer: As we use instances as the intensification target and optimize quality, couldn't we set this to `True`?

examples/SMAC4HPO_hyperband_mlp_instances.py
```diff
@@ -0,0 +1,156 @@
...
+                                                # Timeouts have to be taken care within the TA
+                     "cutoff": 10,              # Cutoff denotes the number of epochs for MLP.
```
mfeurer: Hm, I find this unintuitive. Shouldn't this be a hyperparameter or a constant, and `cutoff` be a time limit/memory limit?

examples/SMAC4HPO_hyperband_mlp_instances.py
```diff
@@ -0,0 +1,156 @@
...
+                                                # It constant across all instances
+                     "instances": instances     # Optimize across all given instances
+                     })
+
+# intensifier parameters
+# if no argument provided for budgets, hyperband decides them based on the number of instances available
+intensifier_kwargs = {'initial_budget': 1, 'max_budget': 4, 'eta': 2,
+                      'instance_order': None}  # You can also shuffle the order of using instances by this parameter
```
mfeurer: Could you please explicitly note the different options?

examples/SMAC4HPO_hyperband_mlp_instances.py
```diff
@@ -0,0 +1,156 @@
...
+                                                        # This example currently uses Hyperband intensification,
+                intensifier_kwargs=intensifier_kwargs)  # all parameters related to intensifier can be passed like this
+
+# Example call of the function with default values
+# It returns: Status, Cost, Runtime, Additional Infos
+def_costs = []
+for i in instances:
+    cost = smac.get_tae_runner().run(cs.get_default_configuration(), i[0], 10, 0)[1]
+    def_costs.append(cost)
+print("Value for default configuration: %.4f" % (np.mean(def_costs)))
```
mfeurer: I don' t think computing the mean across different budgets is what we want, but instead seeing how the value changes with different budgets?

examples/SMAC4HPO_hyperband_mlp_instances.py
```diff
@@ -0,0 +1,156 @@
...
+            solver=cfg['solver'],
+            batch_size=batch_size,
+            activation=cfg['activation'],
+            learning_rate=lr,
+            learning_rate_init=lr_init,
+            max_iter=int(np.ceil(budget)),
+            random_state=seed)
+
+        # returns the cross validation accuracy
+        score = cross_val_score(mlp, digits.data, digits.target, cv=int(instance))
```
mfeurer: Isn't the `cross_val_score` non-deterministic if you don't pass a CV object with a fixed random state? If yes, could you please change that (also for the other example)?

examples/SMAC4HPO_hyperband_mlp_instances.py
```diff
@@ -0,0 +1,156 @@
...
+try:
+    incumbent = smac.optimize()
+finally:
+    incumbent = smac.solver.incumbent
+
+inc_costs = []
+for i in instances:
+    cost = smac.get_tae_runner().run(incumbent, i[0], 10, 0)[1]
+    inc_costs.append(cost)
+print("Optimized Value: %.4f" % (np.mean(inc_costs)))
```
mfeurer: I guess we want to show the final 10-fold CV performance here?

smac/intensification/successive_halving.py
```diff
@@ -26,6 +26,25 @@ class SuccessiveHalving(Intensifier):
     Implementation from "BOHB: Robust and Efficient Hyperparameter Optimization at Scale" (Falkner et al. 2018)
     (refer supplementary)
```
mfeurer: Can we add this link http://proceedings.mlr.press/v80/falkner18a/falkner18a-supp.pdf here? That'll make it easier to find what we're referring to. Also, could you please write "Implementation following the description in ..."?

smac/intensification/successive_halving.py
```diff
@@ -26,6 +26,25 @@ class SuccessiveHalving(Intensifier):
...
 
+    Successive Halving intensifier can operate on two kinds of budgets:
+    1. **'Instances' as budget**:
+        When multiple instances are provided or when run objective is "runtime", this is the criterion used as budgets
+        for successive halving iterations i.e., for each budget, the challengers are tested across an
+        increasing number of instances and their combined performance across all instances is used for evaluation.
```
mfeurer: Could you please explicitly note that in every iteration one instance is added and the users have to make sure themselves that this actually doubles the time it takes in practice?

@mlindauer is this the expected behavior? Should we add a mode where the number of instances are explicitly doubled - following more the BOHB implementation?

smac/intensification/successive_halving.py
```diff
@@ -26,6 +26,25 @@ class SuccessiveHalving(Intensifier):
...
+
+        If `initial_budget` and `max_budget` are not provided, then they are set to 1 and total number
+        of available instances respectively by default.
+    2. **'Runtime' as budget**:
+        Runtime/Cutoff is used when there is only one instance provided or when run objective is "quality".
```
mfeurer: I guess this line should end with a comma.

smac/intensification/successive_halving.py
```diff
@@ -26,6 +26,25 @@ class SuccessiveHalving(Intensifier):
...
+        i.e., the budget determines how long the challenger is allowed to run.
+        Keep in mind that this need not necessarily be wallclock time.
+        Since the cutoff can be passed to the target algorithm as an argument, you can set anything as the cutoff,
+        (eg: number of epochs for training a neural network) as long as you disable SMAC's internal resource limiter
+        by setting `limit_resources=False` in scenario.
```
mfeurer: Could you please make this more detailed and describe the two cases (runtime limit handled by SMAC and not passed to the target algorithm and cutoff passed to the target algorithm) in more detail, for example by explicitly noting that there are two ways to set this up. Maybe we want one more example showing how to do so?

smac/intensification/successive_halving.py
```diff
@@ -26,6 +26,25 @@ class SuccessiveHalving(Intensifier):
...
+
+        `initial_budget` is a required parameter for this type of budget.
+        If `max_budget` is not provided, then `cutoff` is used as the default maximum budget.
```
mfeurer: I'm wondering whether we should raise an exception if the user gives both. This will make it harder to make a mistake here.

examples/SMAC4HPO_hyperband_mlp_instances.py
ashraaghav: Right now, since `cutoff` is not strictly coupled to pynisher runtime limits, I thought this will be an example of using an arbitrary cutoff (no. of epochs in this case). That is also why this example doesn't use pynisher. But I agree, this does make it a bit unintuitive.

examples/SMAC4HPO_hyperband_mlp_instances.py
ashraaghav: I initially thought it would be the average performance across all instances, but I can make it 10-fold CV too.

smac/intensification/successive_halving.py
ashraaghav: The number of instances actually doubles (increases `eta` folds to be precise) like how it is described in the algorithm. We don't do anything different with increasing budgets for runtime or instances. I'll make it clearer in the docstring.

smac/intensification/successive_halving.py
ashraaghav: Would it make it easier if we decouple `cutoff` and `max_budget` altogether? If yes, then when using runtime as budget, should we raise an exception/warning when both `cutoff` and `max_budget` is provided?

mfeurer: Yes, that would be great. Let's make it as hard as possible to misuse this software.

examples/SMAC4HPO_hyperband_mlp_instances.py
mfeurer: Thanks for the update, but I think you forgot to change it for the other example.

examples/SMAC4HPO_svm_hyperband_instances.py
```diff
@@ -0,0 +1,165 @@
+"""
+An example for the usage of SMAC within Python.
+We optimize a SVM on the MNIST digits dataset as multiple binary classification problems
+using "Hyperband" intensification. We split the MNIST digits dataset (10 classes) into 45 binary datasets.
+
+In this example, we use instances as the budget in hyperband and optimize the average cross validation accuracy.
+An "Instance" represents a specific scenario/condition (eg: different datasets, subsets, transformations)
+for the algorithm to run. SMAC then returns the algorithm that had the best performance across all the instances.
+In this case, an instance here is a binary dataset i.e., digit-2 vs digit-3.
```
mfeurer: You can drop the `here` for the comment.

examples/SMAC4HPO_svm_hyperband_instances.py
```diff
@@ -0,0 +1,165 @@
...
+                                               # 'shuffle_once' will shuffle instances once before the 1st
+                                               # SH iteration begins
+                      }
+
+# To optimize, we pass the function to the SMAC-object
+smac = SMAC4HPO(scenario=scenario, rng=np.random.RandomState(42),
+                tae_runner=svm_from_cfg,
+                intensifier=Hyperband,                  # you can also change the intensifier to use like this!
+                                                        # This example currently uses Hyperband intensification,
+                intensifier_kwargs=intensifier_kwargs)  # all parameters related to intensifier can be passed like this
```
mfeurer: could you please rename `parameters` to `arguments` to avoid confusion here?

examples/SMAC4HPO_svm_hyperband_instances.py
```diff
@@ -0,0 +1,165 @@
...
+try:
+    incumbent = smac.optimize()
+finally:
+    incumbent = smac.solver.incumbent
+
+inc_costs = []
+for i in instances:
+    cost = smac.get_tae_runner().run(incumbent, i[0])[1]
+    inc_costs.append(cost)
+print("Optimized Value: %.4f" % (np.mean(inc_costs)))
```
mfeurer: Great example!

smac/intensification/successive_halving.py
```diff
@@ -27,26 +27,31 @@ class SuccessiveHalving(Intensifier):
...
-        Runtime/Cutoff is used when there is only one instance provided and when run objective is "quality",
+        Runtime is used when there is only one instance provided and when run objective is "quality",
         i.e., budget determines how long the challenger is allowed to run.
-        This is interpreted as wallclock time by SMAc's internal resource limiter by default.
-        But the cutoff can also be passed to the target algorithm as an argument to make it an arbitrary runtime limit,
-        (eg: number of epochs for training a neural network) as long as you disable SMAC's internal resource limiter
-        by setting `limit_resources=False` in scenario.
+        The budget is passed to the target algorithm as an argument to make it an arbitrary runtime limit,
+        (eg: number of epochs for training a neural network).
+        Since this deals with runtime, SMAC's internal resource limiter has to be disabled by setting
```
mfeurer: Does it? Can't they both be active? I agree that it's suboptimal that the cutoff isn't scaled, but in principle, there's nothing holding us back from doing so.

smac/intensification/successive_halving.py
```diff
@@ -27,26 +27,31 @@ class SuccessiveHalving(Intensifier):
...
+        `initial_budget` and `max_budget` are required parameters for this type of budget.
+        `cutoff` should NOT be provided.
 
-        `initial_budget` is a required parameter for this type of budget.
-        If `max_budget` is not provided, then `cutoff` is used as the default maximum budget.
-        If both are provided, `cutoff` is ignored.
+    Examples for successive halving (and hyperband) can be found here:
+    * Runtime objective and multiple instances *(instances as budget)*: `examples/spear_qcp/SMAC4AC_spear_qcp.py`
+    * Quality objective and multiple instances *(instances as budget)*: `examples/SMAC4HPO_svm_hyperband_instances.py`
+    * Quality objective and single instance *(runtime as budget)*: `examples/SMAC4HPO_mlp_hyperband.py`
```
mfeurer: I still find this confusing as the budget is not really the runtime, but a positive, real-valued number. Can we rename this to something which reflects this?

smac/intensification/successive_halving.py
```diff
@@ -186,35 +191,33 @@ def _init_sh_params(self, initial_budget: typing.Optional[float],
...
-                                 "requires parameters initial_budget and max_budget/cutoff for intensification!")
+            if initial_budget is None or max_budget is None:
+                raise ValueError("Successive Halving with runtime-limit as budget (i.e., only 1 instance) "
+                                 "requires parameters initial_budget and max_budget for intensification!")
 
-            if self.cutoff is not None and max_budget is not None:
-                self.logger.warning('Successive Halving with runtime-cutoff as budget: '
-                                    'Both max budget (%d) and runtime-cutoff (%d) were provided. '
-                                    'Max budget will be used.' % (max_budget, self.cutoff))
+            if self.cutoff is not None:
```
mfeurer: I don't think this is necessary

smac/intensification/successive_halving.py
```diff
@@ -420,7 +423,7 @@ def _run_challengers(self, challengers: typing.List[Configuration],
                     break
 
                 # setting cutoff based on the type of budget & adaptive capping
-                tae_cutoff = budget if self.runtime_as_budget else cutoff
+                tae_cutoff = cutoff if self.instance_as_budget else budget
```
mfeurer: Shouldn't the cutoff always be the cutoff?

smac/tae/execute_ta_run.py
```diff
@@ -213,7 +213,7 @@ def start(self, config: Configuration,
 
         if self.runhistory:
             # NOTE: cutoff is not adapted if run objective is 'quality', hence, recording it as 'budget'
-            # if there is no budget/budget as instance, cutoff still remains constant throughout the optimization
+            # if budget is instance, cutoff still remains constant throughout the optimization
```
mfeurer: I think it would be better if the TAEs would support both cutoff and budget.

mfeurer: Not sure how to handle this for old-style TAEs (the command-line ones). @mlindauer would that be an issue if they can't handle budgets?

smac/intensification/successive_halving.py
ashraaghav: Can we call it "arbitrary" budget? Since it can represent anything the user wants it to in the TA.

mfeurer: How about calling them real-valued? "arbitrary" has a negative connotation for me.

smac/intensification/successive_halving.py
mfeurer: Could you please add to the docs that the number of instances is multiplied by eta?

examples/SMAC4HPO_mlp_hyperband.py
```diff
@@ -1,13 +1,16 @@
 """
 An example for the usage of SMAC within Python.
 We optimize a simple MLP on the MNIST digits dataset using "Hyperband" intensification.
+
+In this example, we use runtime/cutoff (here training epochs) as the budget in hyperband and
```
mfeurer: I think you forgot to update this comment.

smac/intensification/successive_halving.py
```diff
@@ -23,8 +23,31 @@
...
+    Supplementary reference: http://proceedings.mlr.press/v80/falkner18a/falkner18a-supp.pdf
+
+    Successive Halving intensifier (and Hyperband) can operate on two kinds of budgets:
+    1. **'Instances' as budget**:
+        When multiple instances are provided or when run objective is "runtime", this is the criterion used as budget
+        for successive halving iterations i.e., the budget determines how many instances the challengers are evaluated
+        on at a time. Top challengers for the next iteration are selected based on the combined performance across
+        all instances used.
+
+        If `initial_budget` and `max_budget` are not provided, then they are set to 1 and total number
```
mfeurer: You need to use double backticks here (this needs to be formatted as rst).

examples/SMAC4HPO_svm_hyperband_instances.py
```diff
@@ -0,0 +1,165 @@
...
+    cv = StratifiedKFold(n_splits=4, random_state=seed)  # to make CV splits consistent
+    scores = cross_val_score(clf, data, target, cv=cv)
+    return 1-np.mean(scores)  # Minimize!
+
+
+logger = logging.getLogger("SVM-example")
+logging.basicConfig(level=logging.INFO)  # logging.DEBUG for debug output
+
+# Build Configuration Space which defines all parameters and their ranges
+cs = ConfigurationSpace()
```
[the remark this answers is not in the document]
mfeurer: Could you please simplify further to only to the rbf kernel? Also, could you please use the hyperparameters given in the official tuning guide (https://www.csie.ntu.edu.tw/~cjlin/papers/guide/guide.pdf)?

smac/tae/execute_func.py
```diff
@@ -94,18 +100,18 @@ def run(self, config, instance=None,
...
         Parameters
         ----------
-            config : dictionary (or similar)
+            config : Configuration, dictionary (or similar)
                 Dictionary param -> value
-            instance : str
+            instance : str, optional
                 Problem instance
-            cutoff : int, optional
+            cutoff : float, optional
```
mfeurer: Are you sure about this? The pynisher can't handle float cutoffs.

smac/tae/execute_ta_run.py
```diff
@@ -131,10 +133,13 @@ def start(self, config: Configuration,
                 Mainly a dictionary param -> value
             instance : string
                 Problem instance
-            cutoff : float
+            cutoff : float, optional
```
mfeurer: This should be `int`, too.

smac/tae/execute_ta_run.py
```diff
@@ -251,11 +255,14 @@ def run(self, config: Configuration, instance: str,
                 dictionary param -> value
             instance : string
                 problem instance
-            cutoff : int, optional
+            cutoff : float, optional
```
mfeurer: Same as above.

smac/tae/execute_ta_run_old.py
```diff
@@ -33,12 +34,15 @@ def run(self, config:Configuration,
...
-            instance : string
+            instance : string, optional
                 Problem instance
             cutoff : float
                 Runtime cutoff
             seed : int
                 Random seed
+            budget : float, optional
+                A positive, real-valued number representing an arbitrary limit to the target algorithm
+                Handled by the target algorithm internally
```
mfeurer: Could you please add a comment that the budget is actually ignored?

smac/tae/execute_func.py
ashraaghav: The signature in `execute_func` (int) and `execute_ta_run` (float) were different. I thought this will make it more consistent. Since we convert cutoff to int before using in pynisher anyway, I thought this wouldn't affect it.

mfeurer: Since there's a cast, how about you leave these as `float` and update the docstring.

Issue #545, opened 2019-10-13
Segfault on Archlinux (probably swig related)


gamer01:
#### Description
running `https://github.com/automl/SMAC3/blob/master/examples/fmin_rosenbrock.py` results in a segfault.

I also tried running my own program first and there the error occures inside `SMAC4HPO.__init__()`

#### Steps/Code to Reproduce
1. Install `swig` on archlinux via `pacman`
1. Install smac via `pip`
1. execute example
    => segfault (this is swig 4)

replace step 1 with (continue with smac install):
1. Install `swig3` on archlinux via `pacman`
1. `sudo ln -s /usr/bin/swig-3 /usr/bin/swig` 
    => same segfault

#### Expected Results

no segfault

#### Actual Results

```
/usr/bin/python3.7 /home/gamer01/Projects/Whatsapp_analysis/testsmac.py
/usr/lib/python3.7/site-packages/pyparsing.py:3168: FutureWarning: Possible set intersection at position 3
  self.re = re.compile(self.reString)
INFO:smac.utils.io.cmd_reader.CMDReader:Output to smac3-output_2019-10-13_19:35:02_943499
INFO:smac.facade.smac_hpo_facade.SMAC4HPO:Optimizing a deterministic scenario for quality without a tuner timeout - will make SMAC deterministic and only evaluate one configuration per iteration!

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

#### Versions

0.11.1

gamer01:
I tracked the segfault down to line 71 in `regression.py`

Changing the facade to `SMAC4BO` fixes the issue. Therefore its quite obvious that its related to the random forest implementation.

mlindauer:
Thank you for trying to look into this issue.
Unfortunately, we are not familiar with archlinux s.t. we cannot provide support for it.
Furthermore, our main developer of the pyrfr package left our group quite some time ago; therefore we lack manpower to provide support for the pyrfr package in general. 

If you find a way to fix the issue, we would highly appreciate a PR.

Best,
Marius

gamer01:
I fixed the issue by reinstalling `pyrfr`
```
sudo pip uninstall pyrfr
sudo pip install pyrfr
```

with swig3 installed and linked to 
```
sudo pacman -Syu swig3
sudo ln -s /usr/bin/swig-3 /usr/bin/swig
```

[...]

PR #549, opened 2019-10-14
Update README.md

gamer01:
More detailed Requirements: Swig must be version 3 and is only a build dependency of pyrfr, which can be uninstalled and unlinked afterwards  #545

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ we refer to
 
     Hutter, F. and Hoos, H. H. and Leyton-Brown, K.
     Sequential Model-Based Optimization for General Algorithm Configuration
-    In: Proceedings of the conference on Learning and Intelligent OptimizatioN (LION 5)
+    In: Proceedings of the conference on Learning and Intelligent Optimization (LION 5)
 
 
 SMAC v3 is written in Python3 and continuously tested with python3.5 and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ we refer to
 
     Hutter, F. and Hoos, H. H. and Leyton-Brown, K.
     Sequential Model-Based Optimization for General Algorithm Configuration
-    In: Proceedings of the conference on Learning and Intelligent OptimizatioN (LION 5)
+    In: Proceedings of the conference on Learning and Intelligent Optimization (LION 5)
```
mlindauer: The capital N is in fact intended, since it it LIO*N*

README.md
```diff
@@ -47,10 +47,15 @@ is written in C++.
 ## Requirements
 
 Besides the listed requirements (see `requirements.txt`), the random forest
-used in SMAC3 requires SWIG (>= 3.0).
+used in SMAC3 requires SWIG (<4.0, >= 3.0) as a build dependency. After successful installation of `pyrfr`, SWIG can be uninstalled.
```
mlindauer: Actually, I would like to avoid to tell people what they can uninstall. The next time they try to install SMAC they might wonder why last time it worked and now it does not

gamer01: The formulation is up to you, but i think its crucial to tell, that it IS and Only IS a build dependency. It took me 3 hours to figure out why it has no effect whatsoever i do with swig unless i actually reinstall pyrfr with that configuration.

README.md
gamer01: oh, i even *tried* to look it up :D

gamer01:
Will you do the rest from here? I didnt think much when i requested the merge to master, but as a developer its actually obvious :man_facepalming:. Understood :)

README.md
mlindauer: I agree that we should mention that it is a build dependency. But could you please drop the second sentence.

mlindauer:
I will take care of the PR, but before that you would need to redirect the PR to the dev branch. I think there is no way that I can do that --- at least I'm not aware of any.

gamer01:
just reject it. its just one file that changed. i think its easier if you just copy the two lines by hand into dev and commit it the normal way, right?

mlindauer:
okay, done. Thanks again.

README.md
tunnuz: Just hit this from Ubuntu 20.04. Good thing it was documented.

PR #551, opened 2019-10-25
Pull architecture (v0)

ashraaghav:
The basic design of the pull architecture for sampling challengers on-the-fly.
* Implemented by passing the optimizer (SMBO) as an argument to `intensify()` method
* Currently supports sending both a list of challengers and optimizer as arguments to the intensifier, since `InitialDesign` sends its own configurations for intensification.

**UPDATE**
* `InitialDesign` only returns the configurations to run and all runs happen only through the main BO loop in SMBO. This is to avoid sampling additional configurations (and redundant code) during initial design. 

~**Note**: Testcases have not been modified yet.~

### DIFF
--- a/smac/facade/smac_ac_facade.py
+++ b/smac/facade/smac_ac_facade.py
@@ -426,7 +426,7 @@ class SMAC4AC(object):
             }
         if isinstance(intensifier, SuccessiveHalving):
             # If running successive halving, then you need multiple configurations for the initial design
-            n_configs = intensifier.num_initial_challengers / len(self.scenario.cs.get_hyperparameters())
+            n_configs = intensifier.n_configs_in_stage[0] / len(self.scenario.cs.get_hyperparameters())
             init_design_def_kwargs['n_configs_x_params'] = n_configs
             init_design_def_kwargs['run_first_config'] = False
             init_design_def_kwargs['fill_random_configs'] = True
--- a/smac/facade/smac_ac_facade.py
+++ b/smac/facade/smac_ac_facade.py
@@ -428,8 +428,6 @@ class SMAC4AC(object):
             # If running successive halving, then you need multiple configurations for the initial design
             n_configs = intensifier.n_configs_in_stage[0] / len(self.scenario.cs.get_hyperparameters())
             init_design_def_kwargs['n_configs_x_params'] = n_configs
-            init_design_def_kwargs['run_first_config'] = False
-            init_design_def_kwargs['fill_random_configs'] = True
         if initial_design_kwargs is not None:
             init_design_def_kwargs.update(initial_design_kwargs)
         if initial_configurations is not None:
--- a/smac/facade/smac_ac_facade.py
+++ b/smac/facade/smac_ac_facade.py
@@ -1,7 +1,7 @@
 import inspect
 import logging
 import os
-from typing import  List, Union, Optional, Type, Callable
+from typing import List, Union, Optional, Type, Callable
 
 import numpy as np
 
@@ -30,7 +30,6 @@ from smac.initial_design.sobol_design import SobolDesign
 # intensification
 from smac.intensification.intensification import Intensifier
 from smac.intensification.abstract_racer import AbstractRacer
-from smac.intensification.successive_halving import SuccessiveHalving
 # optimizer
 from smac.optimizer.smbo import SMBO
 from smac.optimizer.objective import average_cost
--- a/smac/facade/smac_ac_facade.py
+++ b/smac/facade/smac_ac_facade.py
@@ -412,22 +412,14 @@ class SMAC4AC(object):
                 "Either use initial_design or initial_configurations; but not both")
 
         init_design_def_kwargs = {
-            'tae_runner': tae_runner,
-            'scenario': scenario,
-            'stats': self.stats,
+            'cs': scenario.cs,
             'traj_logger': traj_logger,
-            'runhistory': runhistory,
             'rng': rng,
+            'ta_run_limit': scenario.ta_run_limit,
             'configs': initial_configurations,
-            'intensifier': intensifier,
-            'aggregate_func': aggregate_func,
             'n_configs_x_params': 0,
             'max_config_fracs': 0.0
             }
-        if isinstance(intensifier, SuccessiveHalving):
-            # If running successive halving, then you need multiple configurations for the initial design
-            n_configs = intensifier.n_configs_in_stage[0] / len(self.scenario.cs.get_hyperparameters())
-            init_design_def_kwargs['n_configs_x_params'] = n_configs
         if initial_design_kwargs is not None:
             init_design_def_kwargs.update(initial_design_kwargs)
         if initial_configurations is not None:
--- a/smac/initial_design/initial_design.py
+++ b/smac/initial_design/initial_design.py
@@ -178,13 +178,30 @@ class InitialDesign:
             # intensifiy will not do any configuration runs
             # (also not on the incumbent)
             # therefore, at least two different configurations have to be in <configs>
-            inc, _ = self.intensifier.intensify(
-                challengers=configs,
-                optimizer=None,
-                incumbent=inc,
-                run_history=self.runhistory,
-                aggregate_func=self.aggregate_func,
-            )
+
+            # running intensification
+            while not self.intensifier.iteration_done:
+                # sample next configuration for intensification
+                config = self.intensifier.next_challenger(
+                    challengers=configs,
+                    chooser=None,
+                    run_history=self.runhistory,
+                    repeat_configs=self.intensifier.repeat_configs
+                )
+
+                # stop if no more configs are generated
+                if not config:
+                    break
+                # remove config from <configs> to not repeat it again
+                configs = [c for c in configs if c != config]
+
+                # evaluate configuration
+                inc, _ = self.intensifier.eval_challenger(
+                    challenger=config,
+                    incumbent=inc,
+                    run_history=self.runhistory,
+                    aggregate_func=self.aggregate_func,
+                )
 
         return inc
 
@@ -256,7 +273,7 @@ class InitialDesign:
                 v_design[v_design == 1] = 1 - 10**-10
                 design[:, idx] = np.array(v_design * len(param.sequence), dtype=np.int)
             else:
-                raise ValueError("Hyperparamer not supported in LHD")
+                raise ValueError("Hyperparameter not supported in LHD")
 
         self.logger.debug("Initial Design")
         configs = []
--- a/smac/initial_design/initial_design.py
+++ b/smac/initial_design/initial_design.py
@@ -1,5 +1,6 @@
 import logging
 import typing
+from collections import OrderedDict
 
 from ConfigSpace.configuration_space import Configuration, ConfigurationSpace
 from ConfigSpace.hyperparameters import NumericalHyperparameter, \
@@ -140,70 +141,23 @@ class InitialDesign:
                     configs.extend(random_configs)
                 else:
                     configs.append(random_configs)
-            return configs
-        else:
-            return self.configs
+            self.configs = configs
 
-    def _select_configurations(self) -> typing.List[Configuration]:
-        raise NotImplementedError
-
-    def run(self) -> Configuration:
-        """Run the initial design.
-
-        Returns
-        -------
-        incumbent: Configuration
-            Initial incumbent configuration
-        """
-        configs = self.select_configurations()
-        for config in configs:
+        for config in self.configs:
             if config.origin is None:
                 config.origin = 'Initial design'
 
         # add this incumbent right away to have an entry to time point 0
         self.traj_logger.add_entry(train_perf=2**31,
                                    incumbent_id=1,
-                                   incumbent=configs[0])
+                                   incumbent=self.configs[0])
 
-        # run first design
-        inc = None
-        if self.run_first_config:
-            # ensures that first design is part of trajectory file
-            inc = self._run_first_configuration(configs[0], self.scenario)
-            configs.pop(0)
+        # removing duplicates
+        self.configs = list(OrderedDict.fromkeys(self.configs))
+        return self.configs
 
-        if len(set(configs)) >= 1:
-            # intensify will skip all challenger that are identical with the incumbent;
-            # if <configs> has only identical configurations,
-            # intensifiy will not do any configuration runs
-            # (also not on the incumbent)
-            # therefore, at least two different configurations have to be in <configs>
-
-            # running intensification
-            while not self.intensifier.iteration_done:
-                # sample next configuration for intensification
-                config = self.intensifier.next_challenger(
-                    challengers=configs,
-                    chooser=None,
-                    run_history=self.runhistory,
-                    repeat_configs=self.intensifier.repeat_configs
-                )
-
-                # stop if no more configs are generated
-                if not config:
-                    break
-                # remove config from <configs> to not repeat it again
-                configs = [c for c in configs if c != config]
-
-                # evaluate configuration
-                inc, _ = self.intensifier.eval_challenger(
-                    challenger=config,
-                    incumbent=inc,
-                    run_history=self.runhistory,
-                    aggregate_func=self.aggregate_func,
-                )
-
-        return inc
+    def _select_configurations(self) -> typing.List[Configuration]:
+        raise NotImplementedError
 
     def _run_first_configuration(self, initial_incumbent, scenario):
         """Runs the initial design by calling the target algorithm and adding new entries to the trajectory logger.
--- a/smac/initial_design/initial_design.py
+++ b/smac/initial_design/initial_design.py
@@ -8,14 +8,7 @@ from ConfigSpace.hyperparameters import NumericalHyperparameter, \
 from ConfigSpace.util import deactivate_inactive_hyperparameters
 import numpy as np
 
-from smac.intensification.abstract_racer import AbstractRacer
-from smac.tae.execute_ta_run import ExecuteTARun
-from smac.stats.stats import Stats
 from smac.utils.io.traj_logging import TrajLogger
-from smac.scenario.scenario import Scenario
-from smac.runhistory.runhistory import RunHistory
-from smac.tae.execute_ta_run import FirstRunCrashedException
-from smac.utils import constants
 
 __author__ = "Marius Lindauer"
 __copyright__ = "Copyright 2019, AutoML"
@@ -27,22 +20,16 @@ class InitialDesign:
 
     Attributes
     ----------
+    cs : ConfigurationSpace
     configs : typing.List[Configuration]
         List of configurations to be evaluated
-    intensifier
-    runhistory
-    aggregate_func
     """
 
     def __init__(self,
-                 tae_runner: ExecuteTARun,
-                 scenario: Scenario,
-                 stats: Stats,
-                 traj_logger: TrajLogger,
-                 runhistory: RunHistory,
+                 cs: ConfigurationSpace,
                  rng: np.random.RandomState,
-                 intensifier: AbstractRacer,
-                 aggregate_func: typing.Callable,
+                 traj_logger: TrajLogger,
+                 ta_run_limit: float,
                  configs: typing.Optional[typing.List[Configuration]] = None,
                  n_configs_x_params: typing.Optional[int] = 10,
                  max_config_fracs: float = 0.25,
@@ -52,26 +39,15 @@ class InitialDesign:
 
         Parameters
         ---------
-        tae_runner: ExecuteTARun
-            Target algorithm execution object.
-        scenario: Scenario
-            Scenario with all meta information (including configuration space).
-        stats: Stats
-            Statistics of experiments; needed in case initial design already
-            exhausts the budget.
+        cs: ConfigurationSpace
+            configuration space object
+        rng: np.random.RandomState
+            Random state
         traj_logger: TrajLogger
             Trajectory logging to add new incumbents found by the initial
             design.
-        runhistory: RunHistory
-            Runhistory with all target algorithm runs.
-        rng: np.random.RandomState
-            Random state
-        intensifier: Intensifier
-            Intensification object to issue a racing to decide the current
-            incumbent.
-        aggregate_func: typing:Callable
-            Function to aggregate performance of a configuration across
-            instances.
+        ta_run_limit: int
+            Number of iterations allowed for the target algorithm
         configs: typing.Optional[typing.List[Configuration]]
             List of initial configurations. Disables the arguments ``n_configs_x_params`` if given.
             Either this, or ``n_configs_x_params`` or ``init_budget`` must be provided.
@@ -87,19 +63,14 @@ class InitialDesign:
             provided.
         """
 
-        self.tae_runner = tae_runner
-        self.stats = stats
-        self.traj_logger = traj_logger
-        self.scenario = scenario
+        self.cs = cs
         self.rng = rng
+        self.traj_logger = traj_logger
         self.configs = configs
-        self.intensifier = intensifier
-        self.runhistory = runhistory
-        self.aggregate_func = aggregate_func
 
         self.logger = self.logger = logging.getLogger(self.__module__ + "." + self.__class__.__name__)
 
-        n_params = len(self.scenario.cs.get_hyperparameters())
+        n_params = len(self.cs.get_hyperparameters())
         if init_budget is not None:
             self.init_budget = init_budget
             if n_configs_x_params is not None:
@@ -111,14 +82,14 @@ class InitialDesign:
             self.init_budget = len(configs)
         elif n_configs_x_params is not None:
             self.init_budget = int(max(1, min(n_configs_x_params * n_params,
-                                              (max_config_fracs * scenario.ta_run_limit))))
+                                              (max_config_fracs * ta_run_limit))))
         else:
             raise ValueError('Need to provide either argument `init_budget`, `configs` or '
                              '`n_configs_x_params`, but provided none of them.')
-        if self.init_budget > scenario.ta_run_limit:
+        if self.init_budget > ta_run_limit:
             raise ValueError(
                 'Initial budget %d cannot be higher than the run limit %d.'
-                % (init_budget, scenario.ta_run_limit)
+                % (init_budget, ta_run_limit)
             )
         self.logger.info("Running initial design for %d configurations" % self.init_budget)
 
--- a/smac/initial_design/initial_design.py
+++ b/smac/initial_design/initial_design.py
@@ -46,8 +46,6 @@ class InitialDesign:
                  configs: typing.Optional[typing.List[Configuration]] = None,
                  n_configs_x_params: typing.Optional[int] = 10,
                  max_config_fracs: float = 0.25,
-                 run_first_config: bool = True,
-                 fill_random_configs: bool = False,
                  init_budget: typing.Optional[int] = None,
                  ):
         """Constructor
@@ -83,10 +81,6 @@ class InitialDesign:
             ``n_configs_x_params`` if given.
         max_config_fracs: float
             use at most X*budget in the initial design. Not active if a time limit is given.
-        run_first_config: bool
-            specify if target algorithm has to be run once on the first configuration before the intensify call
-        fill_random_configs: bool
-            fill budget with random configurations if initial incumbent sampling returns only 1 configuration
         init_budget : int, optional
             Maximal initial budget (disables the arguments ``n_configs_x_params`` and ``configs``
             if both are given). Either this, or ``n_configs_x_params`` or ``configs`` must be
@@ -143,6 +137,7 @@ class InitialDesign:
                                    incumbent=self.configs[0])
 
         # removing duplicates
+        # (Reference: https://stackoverflow.com/questions/7961363/removing-duplicates-in-lists)
         self.configs = list(OrderedDict.fromkeys(self.configs))
         return self.configs
 
--- a/smac/initial_design/initial_design.py
+++ b/smac/initial_design/initial_design.py
@@ -180,6 +180,7 @@ class InitialDesign:
             # therefore, at least two different configurations have to be in <configs>
             inc, _ = self.intensifier.intensify(
                 challengers=configs,
+                optimizer=None,
                 incumbent=inc,
                 run_history=self.runhistory,
                 aggregate_func=self.aggregate_func,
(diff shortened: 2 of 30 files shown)


smac/runhistory/runhistory.py
```diff
@@ -293,7 +293,7 @@ def get_cost(self, config: Configuration):
         cost: float
             Computed cost for configuration
         """
-        config_id = self.config_ids[config]
+        config_id = self.config_ids.get(config)
```
mfeurer: What's the reason for this change?

smac/intensification/abstract_racer.py
```diff
@@ -95,14 +101,74 @@ def __init__(self, tae_runner: ExecuteTARun,
         else:
             self.instance_specifics = instance_specifics
 
-    def intensify(self, challengers: typing.List[Configuration],
+        # general attributes
```
mfeurer: Could you please add an explanation what they are/will be used for?

smac/intensification/abstract_racer.py
```diff
@@ -95,14 +101,74 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+        self._min_time = 10 ** -5
+        self._num_run = 0
+        self._chall_indx = 0
+        self._ta_time = 0
+
+    def intensify(self, challengers: typing.Optional[typing.List[Configuration]],
```
mfeurer: Could you please put both `self` and `challengers` into its own line each? That'll make it easier to spot what's new.

smac/intensification/abstract_racer.py
```diff
@@ -95,14 +101,74 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                  optimizer: typing.Optional['SMBO'],
```
mfeurer: Hm, is this a good name? In the end, it proposes new points?

smac/intensification/abstract_racer.py
```diff
@@ -95,14 +101,74 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                        n_chall: int,
+                        run_history: RunHistory,
+                        repeat_configs: bool = True) -> Configuration:
+        """ Retuns the next challenger to use in intensification
+        If challenger is None, then optimizer will be used to generate the next challenger
+
+        Parameters
+        ----------
+        challengers : typing.List[Configuration]
+            promising configurations
```
mfeurer: Could you please be more specific about which configurations are returned?

smac/intensification/abstract_racer.py
```diff
@@ -95,14 +101,74 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+        optimizer : SMBO
```
mfeurer: Could you please use the full classpath for objects? Then, they'll be rendered in the documentation.

smac/optimizer/smbo.py
```diff
@@ -196,20 +196,13 @@ def run(self):
...
-            start_time = time.time()
-            X, Y = self.rh2EPM.transform(self.runhistory)
-
-            self.logger.debug("Search for next configuration")
-            # get all found configurations sorted according to acq
-            challengers = self.choose_next(X, Y)
-
-            time_spent = time.time() - start_time
-            time_left = self._get_timebound_for_intensification(time_spent)
+            time_left = self._get_timebound_for_intensification()
```
mfeurer: Hm, I'm not sure if this will still work now that doesn't take the time for `choose_next` into account any more.

smac/runhistory/runhistory.py
ashraaghav: When attempting to get the cost of a config that had not been evaluated, this function raised a key error. So using `get()` returns None if a key is missing. Also, this makes it consistent across other `get_*` methods from RunHistory.

smac/intensification/successive_halving.py
```diff
@@ -258,93 +259,248 @@ def intensify(self, challengers: typing.List[Configuration],
...
+        prev_budget = int(self.all_budgets[self.stage - 1]) if self.stage > 0 else 0
+
+        # selecting instance-seed subset for this budget, depending on the kind of budget
+        if self.instance_as_budget:
+            curr_insts = self.inst_seed_pairs[prev_budget:curr_budget]
+        else:
+            curr_insts = self.inst_seed_pairs
+        n_insts_remaining = len(curr_insts) - self.curr_inst_idx - 1
+
+        try:
```
mfeurer: Can this try/except be moved further down to only be around the function call?

smac/intensification/successive_halving.py
```diff
@@ -258,93 +259,248 @@ def intensify(self, challengers: typing.List[Configuration],
...
+        """
+        Update tracking information for a new stage/iteration and update statistics
+
+        Parameters
+        ----------
+         run_history : RunHistory
+            stores all runs we ran so far
+        """
+
+        if not hasattr(self, 'stage'):
```
mfeurer: Could you please do these in the init method? This would make it easier to understand the method above (if not reading this one first).

smac/optimizer/smbo.py
```diff
@@ -197,19 +197,22 @@ def run(self):
...
+            # sample next configuration for intensification
+            challenger = self.intensifier.next_challenger(
+                challengers=None,
+                chooser=self,
+                run_history=self.runhistory,
+                repeat_configs=self.intensifier.repeat_configs
+            )
 
             time_spent = time.time() - start_time
             time_left = self._get_timebound_for_intensification(time_spent)
```
mfeurer: I believe you only need to change this if SMAC actually spends time in `next_challenger`, right?

smac/optimizer/smbo.py
```diff
@@ -365,7 +367,7 @@ def validate(self, config_mode='inc', instance_mode='train+test',
                                         output_fn=new_rh_path)
         return new_rh
 
-    def _get_timebound_for_intensification(self, time_spent:float):
+    def _get_timebound_for_intensification(self, time_spent: float = 0):
```
mfeurer: Could you please remove this default value? The user should always have to provide it to avoid mistakes.

smac/intensification/successive_halving.py
```diff
@@ -250,98 +259,248 @@ def intensify(self, challengers: typing.List[Configuration],
...
+        return incumbent, inc_perf
+
+    def next_challenger(self, challengers: typing.Optional[typing.List[Configuration]],
+                        chooser: typing.Optional['smac.optimizer.smbo.SMBO'],
+                        run_history: RunHistory,
+                        repeat_configs: bool = True) -> Configuration:
+        """
+        Selects which challenger to use based on the iteration stage and set the iteration parameters.
+        First iteration will choose configurations from the ``chooser`` or input challengers,
+        while the later iterations sample top configurations from the previously selected challengers in that iteration
```
mfeurer: It's not "sample", it's "pick".

smac/intensification/hyperband.py
```diff
@@ -149,46 +193,56 @@ def intensify(self, challengers: typing.List[Configuration],
...
-        else:
-            self.s -= 1
+    def next_challenger(self, challengers: typing.Optional[typing.List[Configuration]],
+                        chooser: typing.Optional['smac.optimizer.smbo.SMBO'],
+                        run_history: RunHistory,
+                        repeat_configs: bool = True):
+        """
+        Selects which challenger to use based on the iteration stage and set the iteration parameters.
+        First iteration will choose configurations from the ``chooser`` or input challengers,
+        while the later iterations sample top configurations from the previously selected challengers in that iteration
```
mfeurer: Same as below.

mfeurer:
This PR could actually revert the changes done to the initial design in https://github.com/automl/SMAC3/commit/748c8da1a99336789aa6e352845ebbedd8e132b0#diff-bff62f2acc174ed56c79157041ddd7dd , right?

smac/intensification/successive_halving.py
```diff
@@ -160,6 +159,12 @@ def __init__(self, tae_runner: ExecuteTARun,
         else:
             self.adaptive_capping = False
 
+        # challengers can be repeated only if optimizing across multiple seeds or changing instance orders every run
+        if not (self.n_seeds > 1 or self.instance_order == 'shuffle'):
+            self.repeat_configs = False
```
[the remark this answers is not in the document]
ashraaghav: This flag is used only when sampling new configurations. Not when there are instances remaining for a configuration or when running higher budgets where it is required to repeat configurations.

smac/intensification/successive_halving.py
```diff
@@ -224,32 +229,28 @@ def _init_sh_params(self, initial_budget: typing.Optional[float],
         # max. no. of SH iterations possible given the budgets
         max_sh_iter = np.floor(np.log(self.max_budget / self.initial_budget) / np.log(self.eta))
         # initial number of challengers to sample
-        if num_initial_challengers is None:
-            self.num_initial_challengers = int(self.eta ** max_sh_iter)
-        else:
-            self.num_initial_challengers = int(num_initial_challengers)
-        # list of budgets that will be used in intensification
+        if not num_initial_challengers:
```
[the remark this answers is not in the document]
ashraaghav: We need this for `Hyperband`. Hyperband has a different way to calculate the initial number of challengers for its successive halving iterations. So we will still need it.

smac/optimizer/smbo.py
ashraaghav: Yes. The EPM and the local search takes a non-negligible time, so this is included after sampling the next configuration.

smac/intensification/abstract_racer.py
```diff
@@ -125,10 +125,10 @@ def eval_challenger(self, challenger: Configuration,
...
         raise NotImplementedError()
 
-    def next_challenger(self, challengers: typing.Optional[typing.List[Configuration]],
-                        chooser: typing.Optional['smac.optimizer.smbo.SMBO'],
-                        run_history: RunHistory,
-                        repeat_configs: bool = True) -> Configuration:
+    def get_next_challenger(self, challengers: typing.Optional[typing.List[Configuration]],
+                            chooser: typing.Optional['smac.optimizer.smbo.SMBO'],
+                            run_history: RunHistory,
+                            repeat_configs: bool = True) -> Configuration:
```
mfeurer: Actually, the return value is `Optional[Configuration]`, isn't it?

smac/intensification/hyperband.py
```diff
@@ -201,22 +208,20 @@ def eval_challenger(self, challenger: Configuration,
...
         return incumbent, inc_perf
 
-    def next_challenger(self, challengers: typing.Optional[typing.List[Configuration]],
-                        chooser: typing.Optional['smac.optimizer.smbo.SMBO'],
-                        run_history: RunHistory,
-                        repeat_configs: bool = True):
+    def get_next_challenger(self, challengers: typing.Optional[typing.List[Configuration]],
+                            chooser: typing.Optional['smac.optimizer.smbo.SMBO'],
+                            run_history: RunHistory,
+                            repeat_configs: bool = True):
```
mfeurer: Could you please add a return type here?

smac/optimizer/smbo.py
```diff
@@ -157,26 +159,7 @@ def start(self):
         if self.stats.ta_runs == 0 and self.incumbent is None:
             self.logger.info('Running initial design')
             # Intensifier initialization
-            self.incumbent = self.initial_design.run()
-
-        elif self.stats.ta_runs > 0 and self.incumbent is None:
```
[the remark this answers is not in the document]
ashraaghav: These were checks to ensure that there is an incumbent selected in the initial design, before the main BO loop. Now initial design only returns configurations, so these checks were unnecessary. It did not seem relevant to push these somewhere inside the BO loop either. So I removed them.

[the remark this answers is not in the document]
mfeurer: No, these checks run instead of the initial design (please re-check the if/elif/elif/else clause). So I think they are quite relevant here and should be kept at the exact same location.

smac/initial_design/initial_design.py
```diff
@@ -118,118 +82,38 @@ def __init__(self,
...
-            Initial incumbent configuration
-        """
-        configs = self.select_configurations()
-        for config in configs:
+        for config in self.configs:
             if config.origin is None:
                 config.origin = 'Initial design'
 
         # add this incumbent right away to have an entry to time point 0
         self.traj_logger.add_entry(train_perf=2**31,
```
mfeurer: Can this be moved to the start() function of the SMBO class? I think it would make sense to have it closer to where runs are conducted, and it just does not fit here.

mfeurer: It would also remove the last argument not directly related to the initial design out of this class.

smac/facade/smac_ac_facade.py
```diff
@@ -412,24 +411,14 @@ def __init__(self,
...
-            'runhistory': runhistory,
             'rng': rng,
+            'ta_run_limit': scenario.ta_run_limit,
             'configs': initial_configurations,
-            'intensifier': intensifier,
-            'aggregate_func': aggregate_func,
             'n_configs_x_params': 0,
             'max_config_fracs': 0.0
             }
-        if isinstance(intensifier, SuccessiveHalving):
```
mfeurer: What's the reason why this is no longer needed?

smac/intensification/abstract_racer.py
```diff
@@ -126,12 +123,28 @@ def eval_challenger(self, challenger: Configuration,
...
+                            repeat_configs: bool = True) -> typing.Optional[Configuration]:
         """
         Abstract method for choosing the next challenger, to allow for different selections across intensifiers
         uses ``_next_challenger()`` by default
+
+        Parameters
+        ----------
+        challengers : typing.List[Configuration]
+            promising configurations
+        chooser : EPMChooser
```
mfeurer: Please give the full class path (necessary for constructing docs).

smac/intensification/abstract_racer.py
```diff
@@ -126,12 +123,28 @@ def eval_challenger(self, challenger: Configuration,
...
+            optimizer that generates next configurations to use for racing
+        run_history : RunHistory
```
mfeurer: same here.

smac/intensification/hyperband.py
```diff
@@ -227,13 +164,23 @@ def get_next_challenger(self, challengers: typing.Optional[typing.List[Configura
         ----------
         challengers : typing.List[Configuration]
             promising configurations
-        chooser : 'smac.optimizer.smbo.SMBO'
+        chooser : EPMChooser
```
mfeurer: same as above.

smac/intensification/hyperband.py
```diff
@@ -227,13 +164,23 @@ def get_next_challenger(self, challengers: typing.Optional[typing.List[Configura
...
             optimizer that generates next configurations to use for racing
         run_history : RunHistory
```
mfeurer: same as above.

smac/intensification/hyperband.py
```diff
@@ -242,12 +189,54 @@ def get_next_challenger(self, challengers: typing.Optional[typing.List[Configura
...
-        int
+        Parameters
+        ----------
+         run_history : RunHistory
+            stores all runs we ran so far
         """
-        return self.hb_iters
+        if not hasattr(self, 's'):
+            # setting initial running budget for future iterations (s & s_max from Algorithm 1)
+            self.s_max = np.floor(np.log(self.max_budget / self.initial_budget) / np.log(self.eta))
```
mfeurer: Would it make sense to also cast this to an integer?

smac/intensification/hyperband.py
```diff
@@ -242,12 +189,54 @@ def get_next_challenger(self, challengers: typing.Optional[typing.List[Configura
...
+        Update tracking information for a new stage/iteration and update statistics.
+        This method is called to initialize stage variables and after all configurations
+        of a successive halving stage are completed.
 
-        Returns
-        -------
-        int
+        Parameters
+        ----------
+         run_history : RunHistory
```
mfeurer: Please use the full class-path here.

test/test_intensify/test_abstract_intensifier.py
```diff
@@ -48,6 +49,33 @@ def setUp(self):
...
+        """
+            test get_next_challenger - pick from list/chooser
+        """
+        intensifier = AbstractRacer(
+            tae_runner=None, stats=self.stats, traj_logger=None,
+            rng=np.random.RandomState(12345), deterministic=True, run_obj_time=False,
+            cutoff=1, instances=[1], initial_budget=1, max_budget=3, eta=2)
+
+        # None when nothing to choose from
+        config = intensifier.get_next_challenger(challengers=None, chooser=None, run_history=self.rh)
```
mfeurer: Hm, shouldn't this result in an exception?

test/test_intensify/test_abstract_intensifier.py
```diff
@@ -48,6 +49,33 @@ def setUp(self):
...
+        self.assertEqual(config, None)
+
+        # next challenger from a list
+        config = intensifier.get_next_challenger(challengers=[self.config1], chooser=None, run_history=self.rh)
```
mfeurer: This test would be more convincing if used more than one configuration, and would run for more than one iteration.

test/test_intensify/test_successive_halving.py
```diff
@@ -62,14 +62,16 @@ def test_init_1(self):
             instances=[1, 2, 3], n_seeds=2, initial_budget=None, max_budget=None, eta=2)
 
         self.assertEqual(len(intensifier.inst_seed_pairs), 6)  # since instance-seed pairs
+        self.assertEqual(len(intensifier.instances), 3)
         self.assertEqual(intensifier.initial_budget, 1)
         self.assertEqual(intensifier.max_budget, 6)
-        self.assertEqual(intensifier.num_initial_challengers, 4)  # 2 iterations
+        self.assertListEqual(intensifier.n_configs_in_stage.tolist(), [4.0, 2.0, 1.0])
```
mfeurer: Could you please change the original attribute in intensifier to be a list (unless there's a reason not to do so).

test/test_intensify/test_successive_halving.py
```diff
@@ -149,33 +147,109 @@ def test_top_k_2(self):
...
+        self.assertEqual(intensifier.running_challenger, config)
+
+        # evaluating configuration
+        _ = intensifier.eval_challenger(challenger=config, incumbent=None, run_history=self.rh,
+                                        aggregate_func=average_cost)
+        config = intensifier.get_next_challenger(challengers=[self.config2], chooser=None, run_history=self.rh)
+        self.assertEqual(config, self.config2)
+        self.assertEqual(len(intensifier.curr_challengers), 1)
+
+    def test_get_next_challenger_2(self):
```
mfeurer: This has the same name as the test above.

test/test_intensify/test_successive_halving.py
```diff
@@ -149,33 +147,109 @@ def test_top_k_2(self):
...
         with self.assertRaisesRegex(AssertionError, 'Cannot compare configs'):
             intensifier._top_k(configs=[self.config2, self.config1, self.config3],
                                k=1, run_history=self.rh)
 
+    def test_get_next_challenger_2(self):
+        """
+            test get_next_challenger for a presently running configuration
+        """
+        intensifier = SuccessiveHalving(
+            tae_runner=None, stats=self.stats, traj_logger=None,
```
mfeurer: Is the `tae_runner` actually still a necessary argument here (probably also true for the `traj_logger`)?

test/test_intensify/test_successive_halving.py
```diff
@@ -228,58 +306,26 @@ def target(x):
...
+                                             aggregate_func=average_cost)
 
         self.assertEqual(inc, self.config1)
-        self.assertEqual(self.stats.ta_runs, 2)
+        self.assertEqual(self.stats.ta_runs, 1)
         self.assertEqual(self.stats.inc_changed, 0)
         self.assertEqual(list(self.rh.data.values())[2][2], StatusType.CAPPED)
-        self.assertEqual(list(self.rh.data.values())[3][2], StatusType.CAPPED)
-
-    @attr('slow')
```
mfeurer: Hm, what happened to this test?

test/test_intensify/test_successive_halving.py
```diff
@@ -228,58 +306,26 @@ def target(x):
...
-
-        self.assertEqual(inc, self.config1)
-        self.assertEqual(len(self.rh.get_runs_for_config(self.config1)), 4)
-        self.assertEqual(len(self.rh.get_runs_for_config(self.config2)), 1)
-
-    @attr('slow')
-    def test_intensify_5(self):
+    def test_intensify_3(self):
         """
-            test intensify with shuffling instance order every run
```
mfeurer: What happened to this test?

test/test_intensify/test_hyperband.py
```diff
@@ -50,87 +51,89 @@ def setUp(self):
...
-                                       aggregate_func=average_cost)
+        self.assertEqual(intensifier.s, 2.0)
+        self.assertEqual(intensifier.hb_iters, 0)
+        self.assertEqual(intensifier.sh_intensifier.initial_budget, 0.25)
+        self.assertEqual(intensifier.sh_intensifier.n_configs_in_stage.tolist(), [4.0, 2.0, 1.0])
 
-        self.assertEqual(intensifier.s_max, 3)
-        self.assertEqual(inc, self.config1)
+        # HB iteration completed
+        intensifier.s = 0.0
```
mfeurer: Could you please call update stage multiple times instead of taking this shortcut?

test/test_intensify/test_hyperband.py
```diff
@@ -50,87 +51,89 @@ def setUp(self):
...
-            instances=[1], initial_budget=0.5, max_budget=1, eta=2)
+            instances=[None], initial_budget=0.5, max_budget=1, eta=2)
 
-        # ensuring correct parameter initialization
+        self.assertFalse(hasattr(intensifier, 's'))
+
+        # Testing get_next_challenger - get next configuration
+        config = intensifier.get_next_challenger(challengers=[self.config2], chooser=None, run_history=self.rh)
         self.assertEqual(intensifier.s, intensifier.s_max)
+        self.assertIsInstance(config, Configuration)
```
mfeurer: You could check that the exact same configuration is returned. Alternatively, you could mock out successive halving and check that it is called correctly.

smac/optimizer/smbo.py
```diff
@@ -131,21 +127,25 @@ def __init__(self,
...
 
-        self._random_search = RandomSearch(
-            acquisition_func, self.config_space, rng
-        )
-
-        self.predict_incumbent = predict_incumbent
+        self.initial_design_configs = []
+
+        # initialize the chooser to get configurations from the EPM
+        self.epm_chooser = EPMChooser(scenario=scenario,
```
mfeurer: Shouldn't this be constructed in the ACFacade?

smac/optimizer/epm_configuration_chooser.py
```diff
@@ -0,0 +1,183 @@
...
+                     ChooserNoCoolDown, ChooserLinearCoolDown]=ChooserNoCoolDown(2.0),
+                 predict_incumbent: bool = True):
+        """
+        Interface to train the EPM and generate next configurations
+
+        Parameters
+        ----------
+        scenario: smac.scenario.scenario.Scenario
+            Scenario object
+        stats: Stats
```
mfeurer: Please use full class paths here.

test/test_smbo/test_epm_configuration_chooser.py
```diff
@@ -0,0 +1,182 @@
...
+
+        x = next(smbo.epm_chooser.choose_next()).get_array()
+        self.assertEqual(x.shape, (2,))
+
+    def test_choose_next_w_empty_rh(self):
+        seed = 42
+        smbo = SMAC4AC(self.scenario, rng=seed).solver
+        smbo.runhistory = RunHistory(aggregate_func=average_cost)
+
+        # should return random search configuration
```
mfeurer: Could you please check the value of `x.origin`? It should be something like 'random search'.

smac/facade/smac_ac_facade.py
ashraaghav: This was done to allow `InitialDesign` to fill the required challengers for a full successive halving run in initial design. Now that we don't have to fill configurations, this part was unnecessary.

test/test_intensify/test_successive_halving.py
ashraaghav: It is a numpy array and it could not be compared directly. So I had to switch it to a list here.

test/test_intensify/test_successive_halving.py
ashraaghav: This functionality being tested in this is already being tested in the other test cases, so this was redundant.

test/test_intensify/test_successive_halving.py
ashraaghav: Yes. We just don't pass any during test cases when we know that will not be used.

test/test_intensify/test_successive_halving.py
ashraaghav: Same reason as the other. But now I think it should be included back since it is a special case.

smac/optimizer/smbo.py
ashraaghav: The other optimizers (like epils) have a different method of selecting the next challenger, so I am not sure if that would be right to pass the chooser as an argument to any general smbo class. Any recommendations?

smac/initial_design/initial_design.py
ashraaghav: `traj_logger` is not a part of SMBO since they are not directly related either. So, it might be better if it remained at `InitialDesign`?

test/test_intensify/test_abstract_intensifier.py
ashraaghav: When you cannot repeat configurations, this gives a chance for exiting gracefully in the loop or try again. But I can also make it raise an error if that would be better.

smac/optimizer/smbo.py
mfeurer: Hm, but this is not accounted for right now anyway, right?

ashraaghav: Not yet. But the main BO loop in that should change too, for the new design.

smac/initial_design/initial_design.py
mfeurer: Hm, good point. However, it would reduce the amount of arguments that need to be passed to the initial design, and I think make the overall design a bit clearer as the initial design will have only a single responsibility.

mfeurer:
Could you please figure out a way to only print the logger output if a test fails? Currently, the whole test fails because there's too much logger output.

smac/optimizer/epils.py
```diff
@@ -149,7 +149,27 @@ def run(self):
         """
         self.stats.start_timing()
         try:
-            self.incumbent = self.initial_design.run()
+            initial_design_configs = self.initial_design.select_configurations()
```
mfeurer: This is now substantially different than smbo.py. Is it possible to inherit from smbo.py here?

smac/intensification/intensification.py
```diff
@@ -112,13 +110,34 @@ def __init__(self, tae_runner: ExecuteTARun,
...
-    def intensify(self, challengers: typing.Optional[typing.List[Configuration]],
-                  optimizer: typing.Optional['SMBO'],
-                  incumbent: Configuration,
-                  run_history: RunHistory,
-                  aggregate_func: typing.Callable,
-                  time_bound: float = float(MAXINT),
-                  log_traj: bool = True) -> typing.Tuple[Configuration, float]:
+        # stage variables
+        # the intensification procedure is divided into 3 'stages':
+        # 1. add incumbent run
```
mfeurer: could you please add line numbers for the Intensify pseudo-code from the SMAC paper to the different variables??

smac/intensification/intensification.py
```diff
@@ -112,13 +110,34 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+        # 2. race challenger
+        # 3. race against configuration for a new incumbent
+        self.run_incumbent = True
+        self.run_challenger = False
+        self.first_run = True
+        self.n_iters = 0
+
+        # challenger related variables
+        self._chall_indx = 0
+        self.running_challenger = None
```
mfeurer: I guess this variable should be named `current_challenger` as it does not specifiy whether something is running, but what is running.

smac/intensification/intensification.py
```diff
@@ -112,13 +110,34 @@ def __init__(self, tae_runner: ExecuteTARun,
...
-                  run_history: RunHistory,
-                  aggregate_func: typing.Callable,
-                  time_bound: float = float(MAXINT),
-                  log_traj: bool = True) -> typing.Tuple[Configuration, float]:
+        # stage variables
+        # the intensification procedure is divided into 3 'stages':
+        # 1. add incumbent run
+        # 2. race challenger
+        # 3. race against configuration for a new incumbent
+        self.run_incumbent = True
```
mfeurer: As there can be three states (run incumbent, run challenger and run default) I think it would be better to use an enum for the different states to make the code more readable.

mfeurer: Potentially, there are four stages, with the fourth one being done. I know this can be represented with the current variables, too, but the code is currently very hard to read because the state is encoded across multiple variables.

smac/initial_design/initial_design.py
mfeurer: Given the EPILS facade I see the advantage of having the logger in the initial design.

test/test_intensify/test_abstract_intensifier.py
mfeurer: Thanks a lot for the change, but you also need to remove the comment above.

smac/optimizer/smbo.py
mfeurer: okay, that makes sense. Thanks for pointing that out.

smac/optimizer/smbo.py
```diff
@@ -213,20 +213,21 @@ def run(self):
...
-                incumbent=self.incumbent,
-                run_history=self.runhistory,
-                aggregate_func=self.aggregate_func,
-                time_bound=max(self.intensifier._min_time, time_left))
-
-            if self.scenario.shared_model:
-                pSMAC.write(run_history=self.runhistory,
-                            output_directory=self.scenario.output_dir_for_this_run,
-                            logger=self.logger)
+            if challenger:
```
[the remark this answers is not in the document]
ashraaghav: `get_next_challenger` can return None for a challenger (e.g, when we do not want to repeat the challengers in successive halving but all configurations generated were used already, they will be filtered out). This check ensures that eval_challenger is called only when there is a challenger.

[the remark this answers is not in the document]
mfeurer: Got it, thank you.

smac/optimizer/smbo.py
mfeurer: Thanks for the change, however, I believe this won't work together with the current implementation in `intensification.py`. Currently, the time against which this is compared is reset to 0 in every call to `eval_challenger` in line 178. The reset to zero should probably happen after the check whether this iteration of successive halving is done. Do we have a unit test for this?

PR #553, opened 2019-11-13
Add argument to directly control the size of the initial design

mfeurer:
This might be advantageous in cases where one does meta-learning and wants to ensure that exactly 2 or 3 configurations are used to start.

### DIFF
--- a/smac/initial_design/initial_design.py
+++ b/smac/initial_design/initial_design.py
@@ -74,9 +74,12 @@ class InitialDesign:
             Function to aggregate performance of a configuration across
             instances.
         configs: typing.Optional[typing.List[Configuration]]
-            List of initial configurations.
+            List of initial configurations. Disables the arguments ``n_configs_x_params`` if given.
+            Either this, or ``n_configs_x_params`` or ``init_budget`` must be provided.
         n_configs_x_params: int
-            how many configurations will be used at most in the initial design (X*D)
+            how many configurations will be used at most in the initial design (X*D). Either
+            this, or ``init_budget`` or ``configs`` must be provided. Disables the argument
+            ``n_configs_x_params`` if given.
         max_config_fracs: float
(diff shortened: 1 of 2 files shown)


smac/initial_design/initial_design.py
```diff
@@ -82,6 +83,8 @@ def __init__(self,
             specify if target algorithm has to be run once on the first configuration before the intensify call
         fill_random_configs: bool
             fill budget with random configurations if initial incumbent sampling returns only 1 configuration
+        init_budget : int, optional
+            Maximal initial budget (alternative to ``n_configs_x_params``)
```
KEggensperger: It is unclear from this documentation what happens if both or None are given

KEggensperger: It is also unclear what happens if, e.g. n_configs_x_params is higher/lower than the initial budget

test/test_initial_design/test_initial_design.py
```diff
@@ -22,8 +22,12 @@ def setUp(self):
...
             name="x1", lower=1, upper=10, default_value=1)
         )
-        self.scenario = Scenario({'cs': self.cs, 'run_obj': 'quality',
-                                  'output_dir': ''})
+        self.scenario = Scenario({
+            'cs': self.cs,
+            'run_obj': 'quality',
+            'output_dir': '',
+            'ta_run_limit': 100,
+        })
```
KEggensperger: Does this change anything?

mfeurer: No, it is only necessary to test that the exception is triggered if we request an initial design larger than the number of target runs.

smac/initial_design/initial_design.py
mfeurer: I just extended the docstrings to clarify the behavior.

PR #554, opened 2019-11-13
Remove default rng from gp prior objects

mfeurer:
To avoid SMAC behaving in a random manner if forgetting to seed these objects.


mfeurer:
> I assume this does not change the current default behaviour?

Correct. It just makes the package easier to use if one constructs the Gaussian process object outside of SMAC.

PR #555, opened 2019-11-13
Issue 488 save incumbent in traj as json

### LINKED ISSUE #488: Saved Incumbent Format
According to the AClib format specifications, we save our trajectory of incumbents such that all parameter values are quoted; so all values are strings. Since we already save our runhistory with the correct types (int, bool, ...) , we should also do it in a further trajectory file such that we can easily read them in.

shukon:
Fix #488 by introducing `alljson` format for trajectory read and write.
Also, add interpretation of non-string categorical values in hyperparameters, such as `CategoricalHyperparameter` with `choices = [1, False, 'random_string]` to address issues such as https://github.com/automl/ParameterImportance/issues/107.

### DIFF
--- a/smac/utils/io/traj_logging.py
+++ b/smac/utils/io/traj_logging.py
@@ -308,8 +308,7 @@ class TrajLogger(object):
                 v = float(v)
             elif isinstance(hp, IntegerHyperparameter):
                 v = int(v)
-            elif (isinstance(hp, CategoricalHyperparameter)
-                  or isinstance(hp, Constant)):
+            elif isinstance(hp, (CategoricalHyperparameter, Constant)):
                 # Checking for the correct type requires jumping some hoops
                 # First, we gather possible interpretations of our string
                 interpretations = [v]
@@ -319,14 +318,16 @@ class TrajLogger(object):
                     interpretations.append(True if v == 'True' else False)
                 else:
                     for t in [int, float]:
-                        try: interpretations.append(t(v))
-                        except ValueError: continue
+                        try:
+                            interpretations.append(t(v))
+                        except ValueError:
+                            continue
 
                 # Second, check if it's in the choices / the correct type.
                 if isinstance(hp, CategoricalHyperparameter):
                     legal = set(interpretations) & set(hp.choices)
                 if isinstance(hp, Constant):
-                    legal = {l for l in legal if type(l) == type(hp.default_value)}
+                    legal = {l for l in legal if type(l) is type(hp.default_value)}
 
                 # Third, issue warnings if the interpretation is not unambigious
                 if len(legal) < 1:
--- a/smac/utils/io/traj_logging.py
+++ b/smac/utils/io/traj_logging.py
@@ -70,6 +70,7 @@ class TrajLogger(object):
                         '"Configuration..."\n')
 
             self.aclib_traj_fn = os.path.join(output_dir, "traj_aclib2.json")
+            self.alljson_traj_fn = os.path.join(output_dir, "traj_alljson.json")
 
         self.trajectory = []
 
@@ -97,6 +98,8 @@ class TrajLogger(object):
                                     ta_time_used, wallclock_time)
             self._add_in_aclib_format(train_perf, incumbent_id, incumbent,
                                       ta_time_used, wallclock_time)
+            self._add_in_alljson_format(train_perf, incumbent_id, incumbent,
+                                        ta_time_used, wallclock_time)
 
     def _add_in_old_format(self, train_perf: float, incumbent_id: int,
                            incumbent: Configuration, ta_time_used: float,
@@ -163,6 +166,8 @@ class TrajLogger(object):
                       "cost": train_perf,
                       "incumbent": conf
                       }
+        # Silent support for passing configuration-dict's directly (since actual Configuration's always have
+        #   origin-attribute since v0.4.11 and minimum required by SMAC is 0.4.6):
         try:
             traj_entry["origin"] = incumbent.origin
         except AttributeError:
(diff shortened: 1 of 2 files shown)


smac/utils/io/traj_logging.py
```diff
@@ -70,6 +69,7 @@ def __init__(self, output_dir, stats):
                         '"Configuration..."\n')
 
             self.aclib_traj_fn = os.path.join(output_dir, "traj_aclib2.json")
+            self.alljson_traj_fn = os.path.join(output_dir, "traj_alljson.json")
```
mlindauer: json.json seems to be a bit redundant

smac/utils/io/traj_logging.py
```diff
@@ -172,6 +176,81 @@ def _add_in_aclib_format(self, train_perf: float, incumbent_id: int,
...
+
+        traj_entry = {"cpu_time": ta_time_used,
+                      "total_cpu_time": None,  # TODO: fix this
+                      "wallclock_time": wallclock_time,
+                      "evaluations": self.stats.ta_runs,
+                      "cost": train_perf,
+                      "incumbent": inc_dict,
+                      }
+        # Silent support for passing configuration-dict's directly (since actual Configuration's always have
+        #   origin-attribute since v0.4.11 and minimum required by SMAC is 0.4.6):
```
mlindauer: maybe it would be better to increase the minimum required version

smac/utils/io/traj_logging.py
```diff
@@ -172,6 +176,81 @@ def _add_in_aclib_format(self, train_perf: float, incumbent_id: int,
...
+        wallclock_time: float
+            Wallclock time used so far
+        """
+        # Silent support for passing configuration-dict's directly:
+        inc_dict = incumbent
+        if isinstance(incumbent, Configuration):
+            inc_dict = incumbent.get_dictionary()
+
+        traj_entry = {"cpu_time": ta_time_used,
+                      "total_cpu_time": None,  # TODO: fix this
```
mlindauer: I don't know whether we will fix/support that anyway in the future. Therefore, we should drop that here.

smac/utils/io/traj_logging.py
```diff
@@ -172,6 +176,81 @@ def _add_in_aclib_format(self, train_perf: float, incumbent_id: int,
...
+        cs: ConfigurationSpace
+            Configuration Space to translate dict object into Confiuration object
+
+        Returns
+        -------
+        trajectory: list
+            Each entry in the list is a dictionary of the form
+            {
+            "cpu_time": float,
+            "total_cpu_time": None, # TODO
```
mlindauer: same as above. please drop

smac/utils/io/traj_logging.py
```diff
@@ -229,6 +307,32 @@ def _convert_dict_to_config(config_list: typing.List[str], cs: ConfigurationSpac
                 v = float(v)
             elif isinstance(hp, IntegerHyperparameter):
                 v = int(v)
+            elif isinstance(hp, (CategoricalHyperparameter, Constant)):
+                # Checking for the correct type requires jumping some hoops
+                # First, we gather possible interpretations of our string
+                interpretations = [v]
+                if v in ["True", "False"]:
```
mfeurer: Wouldn't a boolean be stored as `true` in json representation and then correctly be deserialized?

smac/utils/io/traj_logging.py
```diff
@@ -209,7 +288,6 @@ def read_traj_aclib_format(fn: str, cs: ConfigurationSpace):
 
     @staticmethod
     def _convert_dict_to_config(config_list: typing.List[str], cs: ConfigurationSpace):
```
mfeurer: I know this function already exists, but shouldn't one use `ConfigSpace.util.fix_types` here?

smac/utils/io/traj_logging.py
mfeurer: Or is this only for parsing non-json data?

mlindauer: That function is only called if we read the old aclib format.

smac/utils/io/traj_logging.py
shukon: Sorry for the unclarity: minimum version is not the problem, we do not need this for backwards-compatibility. I simply assumed, there is a reason for this to support passing dict-objects instead of Configurations (as it is also implemented this way in the tests). If there is not, I will remove silent support and rewrite the tests.

smac/utils/io/traj_logging.py
shukon: But isn't "traj_all.json" a bit ambigous? Do you have any proposals for a title of the format?

smac/utils/io/traj_logging.py
shukon: Fair enough. I will remove all TODO-tags in the file.

smac/utils/io/traj_logging.py
shukon: We could, but `ConfigSpace.util.fix_types` assumes all [Categorical's](https://github.com/automl/ConfigSpace/blob/e51b1548ffb0c5df367d5107b1dcc676b3216868/ConfigSpace/util.pyx#L417-L419) and [Constant's](https://github.com/automl/ConfigSpace/blob/e51b1548ffb0c5df367d5107b1dcc676b3216868/ConfigSpace/util.pyx#L423-L425) are strings or am I wrong here?

smac/utils/io/traj_logging.py
mlindauer: I think `traj.json` would suffice and is not ambigous at all.

smac/utils/io/traj_logging.py
mlindauer: At least I cannot remember a reason for why we need to support dicts there. Maybe this is an artifact from the time when we have not used the ConfigSpace.

smac/utils/io/traj_logging.py
```diff
@@ -172,6 +176,81 @@ def _add_in_aclib_format(self, train_perf: float, incumbent_id: int,
...
+        cs: ConfigurationSpace
+            Configuration Space to translate dict object into Confiuration object
+
+        Returns
+        -------
+        trajectory: list
+            Each entry in the list is a dictionary of the form
+            {
+            "cpu_time": float,
+            "total_cpu_time": None,
```
mlindauer: I didn't meant that we drop the comment, but the entire line ;-)

smac/utils/io/traj_logging.py
shukon: Dropping it then.

smac/utils/io/traj_logging.py
shukon: @mfeurer I think i'll leave it here at the least for backwards-compatibility, but I could imagine expanding the `fix_types` would be useful.

shukon:
@mlindauer @mfeurer can I merge?

smac/utils/io/traj_logging.py
mfeurer: Agreed, that's an issue. `fix_types` should inspect the types in the categorical and constants.

PR #557, opened 2019-11-13
Pass more data to the acquisition function

mfeurer:
Changes SMAC to:

* only update the acquisition function with attributes that are already present. This allows to update the acquisition function with costly arguments such as `X`, the whole training data, without the fear of it being stored in an ordinary acquisition function.
* Pass `X`, the incumbent configuration and the incumbent configuration in array format to the acquisition function.
* More unit tests for the acquisition function.

Reason:

This allows to implement acquisition functions which needs to know not only the incumbent value, but also the incumbent configuration, or even all data which was already evaluated.

### DIFF
--- a/smac/optimizer/acquisition.py
+++ b/smac/optimizer/acquisition.py
@@ -40,21 +40,21 @@ class AbstractAcquisitionFunction(object, metaclass=abc.ABCMeta):
         self.logger = PickableLoggerAdapter(self.__module__ + "." + self.__class__.__name__)
 
     def update(self, **kwargs):
-        """Update the acquisition functions values.
+        """Update the acquisition function attributes required for calculation.
 
-        This method will be called if the model is updated. E.g.
-        entropy search uses it to update its approximation of P(x=x_min),
-        EI uses it to update the current fmin.
+        This method will be called after fitting the model, but before maximizing the acquisition
+        function. As an examples, EI uses it to update the current fmin.
 
-        The default implementation takes all keyword arguments and sets the
-        respective attributes for the acquisition function object.
+        The default implementation only updates the attributes of the acqusition function which
+        are already present.
 
         Parameters
         ----------
         kwargs
         """
         for key in kwargs:
-            setattr(self, key, kwargs[key])
+            if hasattr(self, key):
+                setattr(self, key, kwargs[key])
 
     def __call__(self, configurations: List[Configuration]):
         """Computes the acquisition value for a given X
@@ -484,7 +484,6 @@ class LCB(AbstractAcquisitionFunction):
         super(LCB, self).__init__(model)
         self.long_name = 'Lower Confidence Bound'
         self.par = par
-        self.eta = None  # to be compatible with the existing update calls in SMBO
         self.num_data = None
 
     def _compute(self, X: np.ndarray):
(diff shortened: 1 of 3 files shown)


smac/optimizer/acquisition.py
```diff
@@ -40,21 +40,21 @@ def __init__(self, model: AbstractEPM):
...
+        The default implementation only updates the attributes of the acqusition function which
+        are already present.
 
         Parameters
         ----------
         kwargs
         """
         for key in kwargs:
-            setattr(self, key, kwargs[key])
+            if hasattr(self, key):
```
KEggensperger: Why shouldn't you throw an error if one tries to update an attribute that doesn't exist?

smac/optimizer/smbo.py
```diff
@@ -281,37 +291,47 @@ def choose_next(self, X: np.ndarray, Y: np.ndarray,
...
+            incumbent_array = convert_configurations_to_array([all_configs])
+            incumbent_value = self.runhistory.get_cost(incumbent)
             incumbent_value_as_array = np.array(incumbent_value).reshape((1, 1))
+            # It's unclear how to do this for inv scaling and potential future scaling.
+            # This line should be changed if necessary
             incumbent_value = self.rh2EPM.transform_response_values(incumbent_value_as_array)
             incumbent_value = incumbent_value[0][0]
 
-        return incumbent_value
+        return incumbent, incumbent_array, incumbent_value
```
KEggensperger: Shouldn't this function then be renamed to "get_incumbent()"?

test/test_smbo/test_acquisition.py
```diff
@@ -3,7 +3,15 @@
 
 import numpy as np
 
-from smac.optimizer.acquisition import EI, LogEI, EIPS, PI, LCB, IntegratedAcquisitionFunction
+from smac.optimizer.acquisition import (
```
KEggensperger: Are the parentheses necessary?
I don't like this ...

smac/optimizer/acquisition.py
mfeurer: The idea is that the SMBO class passes a large amount of values to the acquisition function, and only the ones which are needed by the acquisition function are stored. This way, the acquisition function does not need to know what can possibly be passed, and the smbo class does not need to know what arguments are required by the acquisition function.

test/test_smbo/test_acquisition.py
mfeurer: Not strictly, but there are two reasons for this coding style:

1. Simpler diffs for future changes: https://stackoverflow.com/a/14377271
2. Suggested by PEP328: https://www.python.org/dev/peps/pep-0328/

smac/optimizer/smbo.py
mfeurer: done

test/test_smbo/test_acquisition.py
KEggensperger: hmpf

smac/optimizer/acquisition.py
KEggensperger: Yes, I just would like to avoid that users thinks they have set an attribute which is not set

mfeurer: Okay, I just added such a check.

PR #563, opened 2019-11-25
Add acquisition function optimizer args to ROAR facade

mfeurer:
This allows me to pass the recent fixed set acquisition function optimizer (#558) to the ROAR facade.

### DIFF
--- a/smac/facade/roar_facade.py
+++ b/smac/facade/roar_facade.py
@@ -8,7 +8,7 @@ from smac.epm.random_epm import RandomEPM
 from smac.facade.smac_ac_facade import SMAC4AC
 from smac.initial_design.initial_design import InitialDesign
 from smac.intensification.abstract_racer import AbstractRacer
-from smac.optimizer.ei_optimization import RandomSearch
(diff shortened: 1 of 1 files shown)


mlindauer:
The code seems to be fine. But I don't get why we should offer an API to pass a acquisition function optimizer to ROAR, since ROAR anyways only uses a random model. So, I would propose to simply use a hard-coded acq optimizer here.

mfeurer:
This allows to run ROAR over a fixed set of candidates. The acquisition function optimizer suggests new points, and the current random search optimizer will suggest random points. If the target function is a grid of pre-computed function evaluations, the random suggestions will not match. Passing an acquisition function optimizer allows to pass a callable which returns only legal configurations (in most cases the full grid of allowed values), like the one added in #558.

mlindauer:
So, a fixed set which is a subspace of the configspace?

mfeurer:
> So, a fixed set which is a subspace of the configspace?

Yes, for example to reproduce the random search results in [Wistuba et al., 2018](https://link.springer.com/article/10.1007/s10994-017-5684-y).

PR #564, opened 2019-12-03
Fix#513 - improve handling of sobol sequence for large search spaces

mfeurer:
The Sobol Sequence implementation we're using fails with more than 40 dimensions. This PR adds

* a unit test for this
* a change to the SMAC4HPO facade to replace the sobol grid if there are too many hyperparameters.

### DIFF
--- a/smac/facade/smac_bo_facade.py
+++ b/smac/facade/smac_bo_facade.py
@@ -53,20 +53,15 @@ class SMAC4BO(SMAC4AC):
         see ~smac.facade.smac_facade for documentation
         """
 
-        # The logger is only initialized on the call to super AFTER constructing all objects. In order to still log
-        # messages, we keep them in an array and fire them off later
-        log_messages = []
-
         scenario = kwargs['scenario']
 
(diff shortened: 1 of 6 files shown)


mlindauer:
Hi,

Actually, I'm skeptical that SMAC should simply change its behavior because it does not support X. 

From a HPO perspective: It would also not be nice if a target algorithm overwrites its configuration and does not what it should do according to its given configuration.

I think throwing an appropriate error is the way to go.

Best,
Marius

mfeurer:
> I think throwing an appropriate error is the way to go.

That makes sense. I just changed the error message.

smac/facade/smac_bo_facade.py
```diff
@@ -5,6 +5,7 @@
 from smac.epm.gp_base_prior import HorseshoePrior, LognormalPrior
 from smac.epm.util_funcs import get_types, get_rng
 from smac.initial_design.sobol_design import SobolDesign
+from smac.initial_design.latin_hypercube_design import LHDesign
```
mlindauer: Can we drop the LHD import again since it's not used?

PR #569, opened 2019-12-03
Incumbent selection

KEggensperger:
Changes the EPMChooser to only choose configurations based on the highest budget available.


smac/facade/smac_bohb_facade.py
```diff
@@ -0,0 +1,96 @@
...
+from smac.epm.rf_with_instances import RandomForestWithInstances
+from smac.initial_design.random_configuration_design import RandomConfigurations
+from smac.intensification.hyperband import Hyperband
+
+__author__ = "Marius Lindauer"
+__copyright__ = "Copyright 2018, ML4AAD"
+__license__ = "3-clause BSD"
+
+
+class SMAC4BOHB(SMAC4AC):
```
mlindauer: The class name is misleading, since BOHB is not the application domains (such as AC or HPO). Looking through the code, I have the impression that I does a lot of stuff the SMAC4HPO facade is also doing. 
First of all, I would like to avoid duplicated code and secondly the class name should reflect the HPO application domain. For example, BOHB4HPO?

KEggensperger: I agree to rename the class, but I would prefer to reduce code duplication after we have seen some results

Issue #449, opened 2018-08-13
`python3` instead of `python`


mlindauer:
We should change the docu to use `python3` instead of `python` since `python` is associated with `python2` by default: https://www.python.org/dev/peps/pep-0394/

PhMueller:
This should be fixed with the pull request #573.

mlindauer:
It looks like we use `python` now instead of `python3`. And not the other way as requested. I'm confused.

Issue #439, opened 2018-05-11
Do not round up cutoff 


mlindauer:
We round up the cutoff to the next integer so far.
That's maybe necessary for the pynisher but in all other use cases (where we don't limit the runtime within SMAC) we should not touch the cutoff. Therefore, I would propose to drop the rounding in execute_ta_run.

mfeurer:
Fine by me, but we should then add it to the execution path via the pynisher.

PhMueller:
Should be fixed with #573.

mlindauer:
@PhMueller Was this fixed in any PR? You closed #573 before mergening and if I'm not wrong, it is still not fixed in the dev branch.

PR #573, opened 2019-12-03
Encoding Error, Removing cutoff ceiling #449, #439: 

PhMueller:
- fixes encoding errors with the authors' names
- converts docs from python3 to python
- ceils the cutoff time only if it is passed to the pynisher


PhMueller:
Oh my bad. I'll fix it and start a new PR.

PR #576, opened 2019-12-05
Remove code

mfeurer:
Removes

* rf_with_instances_hpo.py
* EPILS
* some commented out code in ei_optimization.py (which would replace local search with differential evolution in the interleaved random/local search)
* _component_builder() and _get_acm_cs() in smbo.py
* objective.py - move the cost aggregation function into the runhistory class, remove the runtime aggregation function (as the runtime is in the cost field for runtime minimization), remove the concept of aggregate function


mfeurer:
Thanks for the review Marius, I just added things to the changelog.

PR #583, opened 2020-01-03
MYPY annotations part2: Runhistory

mfeurer:


smac/runhistory/runhistory.py
```diff
@@ -21,7 +21,13 @@
 class RunKey(collections.namedtuple('RunKey', ['config_id', 'instance_id', 'seed', 'budget'])):
     __slots__ = ()
 
-    def __new__(cls, config_id, instance_id, seed, budget=0.0):
+    def __new__(
+        cls,
```
KEggensperger: If there is no type for `cls` it would at least be great to have a docstring explaiing what this is

smac/runhistory/runhistory2epm.py
```diff
@@ -161,7 +163,7 @@ def _build_matrix(self, run_dict: typing.Mapping[RunKey, RunValue],
         """
         raise NotImplementedError()
 
-    def transform(self, runhistory: RunHistory, budget_subset: list = None):
+    def transform(self, runhistory: RunHistory, budget_subset: list = None) -> typing.Tuple[np.ndarray, np.ndarray]:
```
KEggensperger: Shouldn't budget_subset be also a `typing.list`?

smac/runhistory/runhistory2epm.py
```diff
@@ -129,16 +131,16 @@ def __init__(
                              type(self.imputor))
 
         # Learned statistics
-        self.min_y = None
-        self.max_y = None
-        self.perc = None
+        self.min_y = np.NaN
```
KEggensperger: Doesn't this impact line 206 in this file?

smac/runhistory/runhistory.py
mfeurer: I'll add that.

smac/runhistory/runhistory2epm.py
mfeurer: Yes, it should.

smac/runhistory/runhistory2epm.py
mfeurer: Good catch!

PR #589, opened 2020-01-13
Annotate TAE and initial design

mfeurer:


smac/tae/execute_func.py
```diff
@@ -74,6 +83,9 @@ def __init__(self, ta, stats=None, runhistory=None,
         self._accepts_seed = 'seed' in signature.keys()
         self._accepts_instance = 'instance' in signature.keys()
         self._accepts_budget = 'budget' in signature.keys()
+        if not callable(ta):
+            raise TypeError('Argument `ta` must be a callalbe, but is %s' % type(ta))
```
KEggensperger: typo

smac/tae/execute_ta_run.py
```diff
@@ -198,6 +214,7 @@ def start(self, config: Configuration,
                                     "in the trajectory-file.")
 
         if self.run_obj == "runtime":
+            assert cutoff is not None  # please mypy
```
KEggensperger: Could you catch this with if and raise an exception

mfeurer: I added a raise further up in the code prior to calling `super().run()`

PR #590, opened 2020-01-20
Annotate the intensification module

mfeurer:


smac/intensification/successive_halving.py
```diff
@@ -19,6 +19,10 @@
 
 
 class SuccessiveHalving(AbstractRacer):
+    curr_inst_idx = 0
+    running_challenger = None
+    curr_challengers = set()  # type: typing.Set[Configuration]
```
ashraaghav: Why define only 3 of the stage variables outside init?

mfeurer: `mypy` is able to infer the types of the other variables. For a `set()` it's unclear what will be in there so I had to define it.

ashraaghav: Yes, but I was wondering why the three variables here (`curr_inst_idx`, `running_challenger`, `curr_challengers`) are initialized outside of `__init__` as global class variables.

smac/intensification/intensification.py
```diff
@@ -461,6 +466,7 @@ def _get_instances_to_run(self,
 
         # Line 11
         self.rs.shuffle(missing_runs)
+        assert N > 0
```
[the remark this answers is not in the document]
mfeurer: Yes, will change.

smac/intensification/successive_halving.py
mfeurer: It seems that this is not necessary and the variables can be defined in the init, too. Will change.

PR #594, opened 2020-02-20
Type annotation for util files

mfeurer:


smac/utils/validate.py
```diff
@@ -285,7 +298,7 @@ def validate_epm(self,
         elif reuse_epm is False or self.epm is None:
```
ashraaghav: Might be nitpicky, but won't it be better to use `not reuse_epm` ?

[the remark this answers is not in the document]
mfeurer: Yep, changed that.

smac/utils/validate.py
```diff
@@ -295,7 +295,7 @@ def validate_epm(self,
         """
         if not isinstance(runhistory, RunHistory) and (self.epm is None or reuse_epm is False):
```
ashraaghav: same here too

smac/utils/validate.py
```diff
@@ -282,10 +295,10 @@ def validate_epm(self,
...
```
ashraaghav: same here too

PR #598, opened 2020-02-28
Earlier incumbent selection for successive halving

mfeurer:


smac/intensification/successive_halving.py
```diff
@@ -291,7 +294,7 @@ def eval_challenger(self,
         #   - this check is required because there is no incumbent performance
         #     for the first ever 'intensify' run (from initial design)
         if incumbent:
-            inc_runs = run_history.get_runs_for_config(incumbent)
+            inc_runs = run_history.get_runs_for_config(incumbent, only_max_observed_budget=True)
```
KEggensperger: Check why line 304 is different from line 421

smac/intensification/successive_halving.py
```diff
@@ -550,11 +555,61 @@ def _get_incumbent(self,
...
+                        else:
+                            self.logger.debug("  %s remains unchanged: %r" %
+                                              (param[0], param[1]))
+                    new_incumbent = challenger
+                else:
+                    self.logger.debug("Incumbent (%.4f) is at least as good as the challenger (%.4f) on budget %.4f.",
+                                      inc_cost, chall_cost, inc_run.budget)
+                    new_incumbent = incumbent
+
+                print(inc_cost, chall_cost)
```
KEggensperger: Do we need this print?

smac/intensification/successive_halving.py
```diff
@@ -550,11 +555,61 @@ def _get_incumbent(self,
...
+        else:
```
KEggensperger: This is unreachable code?

smac/intensification/successive_halving.py
mfeurer: No, but I already removed it.

smac/intensification/successive_halving.py
mfeurer: I will adapt the code.

smac/intensification/successive_halving.py
mfeurer: I simplified the code to remove the confusion here.

PR #605, opened 2020-03-02
FIX #603 update cost for first evaluated config

### LINKED ISSUE #603: Update costs of first evaluated configuration
Right now (since 0.12.0), we add the first configuration to the trajectory with inf-cost, but we never update its cost, hence the stats of inc-changes are wrong and the trajectory could differ from old runs.

Furthermore, the trajectory logger and the stats object should not be updated independently.

KEggensperger:


smac/intensification/intensification.py
```diff
@@ -410,15 +412,23 @@ def _race_challenger(self,
...
                 # move on to the next iteration
                 self.stage = IntensifierStage.RUN_INCUMBENT
                 self.continue_challenger = False
 
             elif new_incumbent == challenger:
+                # New incumbent found
                 incumbent = challenger
-                # New incumbent found. Go to the next stage
-                self.stage = IntensifierStage.RUN_DEFAULT
+                # compare against default configuration if provided, else go to next iteration
```
KEggensperger: Calling this the default configuration might be confusing. Could you rename this to "basis". Please also change this everywhere if you agree

smac/intensification/intensification.py
```diff
@@ -410,15 +412,23 @@ def _race_challenger(self,
...
+                if self.always_race_against and \
+                        self.always_race_against != challenger:
+                    self.stage = IntensifierStage.RUN_DEFAULT
```
KEggensperger: And then this stage "RUN_CONSTANT" should be called "RUN_BASIS_CONFIGURATION"

smac/intensification/successive_halving.py
```diff
@@ -351,23 +357,33 @@ def eval_challenger(self,
                                   "Interrupting current challenger and moving on to the next one")
                 # ignore all pending instances
                 self.curr_inst_idx = np.inf
+                status = StatusType.CAPPED
 
             # adding challengers to the list of evaluated challengers
-            self.curr_challengers.add(challenger)
+            #  - should not capped
+            #  - should be successful in at least 1 run
+            # curr_challengers is a set, so "at least 1" success can be counted by set addition
```
mfeurer: Okay, so if something worked once, it's added to `curr_challengers`, and if it fails afterwards, it's in both? If yes, could you please clarify this in the comment?

smac/intensification/successive_halving.py
```diff
@@ -351,23 +357,33 @@ def eval_challenger(self,
                                   "Interrupting current challenger and moving on to the next one")
                 # ignore all pending instances
                 self.curr_inst_idx = np.inf
+                status = StatusType.CAPPED
```
KEggensperger: Isn't this already set by the tae?

test/test_intensify/test_intensify.py
```diff
@@ -398,40 +398,145 @@ def target(x):
...
         self.assertFalse(intensifier.continue_challenger)
 
-        # run `always_race_against` now since the incumbent has changed
-        config, _ = intensifier.get_next_challenger(challengers=[self.config2, self.config1, self.config3],
+        # run `always_race_against`
+        config, _ = intensifier.get_next_challenger(challengers=None,  # don't need a new list here as old one is cont'd
                                                     chooser=None)
-        inc, _ = intensifier.eval_challenger(challenger=config, incumbent=inc,
-                                             run_history=self.rh,)
+        self.assertEqual(config, self.config1)
```
KEggensperger: Would it be possible to return config3 here as this is the config that will be evaluated by eval_challenger?

smac/intensification/successive_halving.py
ashraaghav: Yes. I'll add it

test/test_intensify/test_intensify.py
```diff
@@ -398,40 +398,145 @@ def target(x):
...
+        with self.assertRaises(StopIteration):
+            next(intensifier.configs_to_run)
+
+        # intensification iteration #2
+        # run incumbent
+        config, _ = intensifier.get_next_challenger(challengers=None,  # since incumbent is run, no configs required
+                                                    chooser=None)
+        self.assertEqual(config, self.config1)
+        self.assertEqual(intensifier.stage, IntensifierStage.RUN_INCUMBENT)
+        inc, _ = intensifier.eval_challenger(challenger=config, incumbent=inc, run_history=self.rh, )
```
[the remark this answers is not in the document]
ashraaghav: Yes, since it is the RUN_INCUMBENT stage.

test/test_intensify/test_intensify.py
```diff
@@ -398,40 +398,145 @@ def target(x):
...
+                                                    chooser=None)
+        self.assertEqual(config, self.config3)
+        self.assertEqual(intensifier.stage, IntensifierStage.RUN_CHALLENGER)
+        inc, _ = intensifier.eval_challenger(challenger=config, incumbent=inc, run_history=self.rh, )
+
+        self.assertEqual(inc, self.config3)
+        self.assertEqual(self.stats.inc_changed, 3)
+        self.assertEqual(intensifier.stage, IntensifierStage.RUN_INCUMBENT)
+
+        self.assertEqual(len(self.rh.get_runs_for_config(self.config1, only_max_observed_budget=True)), 1)
```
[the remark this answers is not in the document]
ashraaghav: There were 2 calls to `eval_challenger` for `config1`, but we only had 1 instance in a deterministic setting, so there is only 1 TA run for this config.

smac/intensification/successive_halving.py
ashraaghav: Yes, but only in run history. When we get a `CappedRunException`, we don't get outputs from the `ExecuteTARun` in the try block. So we need to set it explicitly here.

PR #610, opened 2020-03-04
Fix #575 bohb facade and examples

### LINKED ISSUE #575: Fix/Extend BOHB
* BOHB facade should inherit from the existing SMAC4HPO facade
* unit tests for the new functions in rh2epm and runhistory to test the selection with budgets (https://github.com/automl/SMAC3/pull/569)
* Change the SH and HB examples to use the new facade.

ashraaghav:
* Fixed the issues addressed in #575 
* Added python version of spear qcp to ci examples script


smac/facade/smac_bohb_facade.py
```diff
@@ -42,47 +40,9 @@ def __init__(self, **kwargs: typing.Any):
...
-        random_config_chooser_kwargs['prob'] = random_config_chooser_kwargs.get('prob', 0.2)
-        kwargs['random_configuration_chooser_kwargs'] = random_config_chooser_kwargs
-
-        # better improve acquisition function optimization
-        # 1. increase number of sls iterations
-        acquisition_function_optimizer_kwargs = kwargs.get('acquisition_function_optimizer_kwargs', dict())
-        acquisition_function_optimizer_kwargs['n_sls_iterations'] = 10
-        kwargs['acquisition_function_optimizer_kwargs'] = acquisition_function_optimizer_kwargs
 
         print(scenario.__dict__.get('always_race_against'))
```
KEggensperger: Do we need this print for something?

test/test_runhistory/test_runhistory2epm.py
```diff
@@ -338,6 +338,27 @@ def test_get_X_y(self):
         c_sol = np.array([False, False, True, True])
         self.assertTrue(np.all(c == c_sol))
 
+    def test_budget_selection(self):
+        '''
+            adding some rundata and check budget selection
+        '''
+
+        rh2epm = runhistory2epm.RunHistory2EPM4LogCost(num_params=2,
```
KEggensperger: Why would you use the LogCost transformer here?

ashraaghav: Any transformer would work, right? But yes, I'll switch it to regular cost

examples/spear_qcp/SMAC4AC_spear_qcp.py
```diff
@@ -23,6 +23,7 @@
                           # non-deterministic target algorithm
                           'initial_budget': 1,
                           'eta': 3,
+                          'min_chall': 1  # because successive halving cannot handle min_chall > 1
                           }
 
     smac = SMAC4AC(scenario=scenario,  # scenario object
```
mfeurer: Wouldn't it be good if you used the BOHB facade here?

ashraaghav: I wanted to show some example that can use SMAC4AC facade and use that to pass a different intensifier. So I combined them. Also, isn't the BOHB facade only for HPO?

mfeurer: > python -c 'import autosklearn.data.competition_c_functions; print("SO imported from: %s" % autosklearn.data.competition_c_functions.__file__)'

That's a good question, maybe we should leave SMAC4AC then.

smac/facade/smac_bohb_facade.py
```diff
@@ -12,7 +10,7 @@
 __license__ = "3-clause BSD"
 
 
-class BOHB4HPO(SMAC4AC):
+class BOHB4HPO(SMAC4HPO):
```
KEggensperger: Shouldn't we also for the sake of clarity explicitly set the Hyperband parameters, e.g. eta, here as well?

PR #611, opened 2020-03-04
Two new y-transformation approaches

mlindauer:
* bilog (log transformations above 0 and below 0)
* Gaussian Copula (ECDF -> quantiles -> Inverse Gaussian CDF)

### DIFF
--- a/smac/runhistory/runhistory2epm.py
+++ b/smac/runhistory/runhistory2epm.py
@@ -13,12 +13,12 @@ from smac.utils import constants
 from smac.scenario.scenario import Scenario
 from smac.utils.constants import VERY_SMALL_NUMBER
 
-__author__ = "Katharina Eggensperger"
-__copyright__ = "Copyright 2015, ML4AAD"
+__author__ = "Katharina Eggensperger, Marius Lindauer"
+__copyright__ = "Copyright 2015-2020, AutoML.org"
 __license__ = "3-clause BSD"
 __maintainer__ = "Katharina Eggensperger"
 __email__ = "eggenspk@cs.uni-freiburg.de"
-__version__ = "0.0.1"
+__version__ = "0.0.3"
 
 
 class AbstractRunHistory2EPM(object):
--- a/smac/runhistory/runhistory2epm.py
+++ b/smac/runhistory/runhistory2epm.py
@@ -3,6 +3,7 @@ import logging
 import typing
 
 import numpy as np
+import scipy as sp
 
 from smac.tae.execute_ta_run import StatusType
 from smac.runhistory.runhistory import RunHistory, RunKey, RunValue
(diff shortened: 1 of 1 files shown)


dengdifan:
If everyone is happy with the implementation, I will merge this branch

smac/runhistory/runhistory2epm.py
```diff
@@ -545,6 +547,57 @@ def transform_response_values(self, values: np.ndarray) -> np.ndarray:
...
+        ----------
+        values : np.ndarray
+            Response values to be transformed.
+
+        Returns
+        -------
+        np.ndarray
+        """
+        # ECDF
+        quants = [sp.stats.percentileofscore(values, v)/100 - VERY_SMALL_NUMBER for v in values]
```
mfeurer, line 596: I believe this is incorrect. I reimplemented this according to [Salinas et al.](http://proceedings.mlr.press/v119/salinas20a.html), which appears to give better, and most importantly, symmetric outputs:
```
import numpy as np
import scipy.stats

values = np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
VERY_SMALL_NUMBER = 1e-10

# This PR
quants = [scipy.stats.percentileofscore(values, v)/100 - VERY_SMALL_NUMBER for v in values]
output = np.array([scipy.stats.norm.ppf(q) for q in quants]).reshape((-1, 1))
print(output)

# Correct
quants = (scipy.stats.rankdata(values.flatten()) - 1) / (len(values) - 1)
cutoff = 1 / (4 * np.power(len(values), 0.25) * np.sqrt(np.pi * np.log(len(values))))
quants = np.clip(quants, a_min=cutoff, a_max=1 - cutoff)
# Inverse Gaussian CDF
rval = np.array([scipy.stats.norm.ppf(q) for q in quants]).reshape((-1, 1))
print(rval)
```
output:
```
[-1.28155157e+00 -8.41621234e-01 -5.24400513e-01 -2.53347103e-01
 -2.50662848e-10  2.53347103e-01  5.24400512e-01  8.41621233e-01
  1.28155156e+00  6.36134089e+00]
[-1.62322583 -1.22064035 -0.76470967 -0.4307273  -0.1397103   0.1397103
  0.4307273   0.76470967  1.22064035  1.62322583]
```

mfeurer (changes requested):
I'm not sure if we want to merge this PR at the moment:

1. We don't have a method that uses quantile transformations
2. I think the quantile transformation should be improved
3. We don't have a method that uses bilog transformations at the moment

alexandertornede:
We will have a look at how these methods perform once we have the new benchmarking fully in place.

mfeurer:
The recent [HEBO](https://www.jair.org/index.php/jair/article/view/13643/26827) suggests using a PowerTransform from scikit-learn. If you plan to benchmark these two, could you also throw this one in the mix?

alexandertornede:
Thanks for the pointer. Sure!

mfeurer:
Was this, by any chance, implemented in #1276 ?

PR #614, opened 2020-03-06
Fix capping issue in successive halving

ashraaghav:
There was an issue with running adaptive capping with successive halving because capped configurations were being considered when selecting for the next stage. Now they are ignored.

### DIFF
--- a/smac/intensification/successive_halving.py
+++ b/smac/intensification/successive_halving.py
@@ -362,11 +362,13 @@ class SuccessiveHalving(AbstractRacer):
             # adding challengers to the list of evaluated challengers
             #  - should not capped
             #  - should be successful in at least 1 run
+            # curr_challengers is a set, so "at least 1" success can be counted by set addition
+            # for each successful instance run for a challenger (no duplicates created)
             if np.isfinite(self.curr_inst_idx) and \
                     status == StatusType.SUCCESS:
                 self.curr_challengers.add(challenger)  # successful configs
             else:
-                self.fail_challengers.add(challenger)   # capped configs
+                self.fail_challengers.add(challenger)   # capped/crashed configs
 
             # get incumbent in the last stage if all instances have been evaluated
             if n_insts_remaining <= 0:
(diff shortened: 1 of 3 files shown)


smac/intensification/successive_halving.py
```diff
@@ -482,12 +488,14 @@ def _update_stage(self, run_history: RunHistory) -> None:
...
-            self.curr_challengers = set()
+            self.curr_challengers = set()  # successful configs
+            self.cap_challengers = set()   # capped configs
 
         else:
             self.stage += 1
 
-            if self.stage < len(self.all_budgets):
+            if self.stage < len(self.all_budgets) and \
+                    len(self.curr_challengers) > 0:
```
mfeurer: Hey, I think it would be great to actually have a unit test to check the case where all runs are capped. What do you think?

ashraaghav: Good idea. There is already a test case for capping. I'll extend that.

test/test_intensify/test_successive_halving.py
```diff
@@ -369,6 +367,61 @@ def target(x: Configuration, instance: str):
...
+        self.assertEqual(inc, self.config1)
+        self.assertEqual(self.stats.ta_runs, 2)
+        self.assertEqual(self.stats.inc_changed, 0)
+        self.assertEqual(list(self.rh.data.values())[3][2], StatusType.CAPPED)
+
+        # now the SH iteration should begin a new iteration since all configs were capped!
+        self.assertEqual(intensifier.sh_iters, 1)
+        self.assertEqual(intensifier.stage, 0)
+
+        # should raise an error
```
mfeurer: Could you please extend the comment here? I don't immediately see why you added this.

test/test_intensify/test_successive_halving.py
```diff
@@ -369,6 +367,61 @@ def target(x: Configuration, instance: str):
...
+                                             run_history=self.rh, )
+        self.assertEqual(inc, self.config1)
+        self.assertEqual(self.stats.ta_runs, 1)
+        self.assertEqual(list(self.rh.data.values())[2][2], StatusType.CAPPED)
+
+        # provide configurations
+        config, _ = intensifier.get_next_challenger(challengers=[self.config3],
+                                                    chooser=None, run_history=self.rh)
+        # config3 should also be capped and config1 should be the incumbent
+        inc, _ = intensifier.eval_challenger(challenger=config,
```
mfeurer: Would it be possible that you mock out the np.argsort in top_k and check that it isn't called? That would make this test even greater.

PR #616, opened 2020-03-12
Stricter checks for things added to runhistory

mfeurer:
I recently faced a weird bug where the configurations in the runhistory are `None`, but which I can't easily debug. This PR will prevent adding `None` to the runhistory and will help me debugging this issue if the bug pops up again.


mfeurer:
I cannot recall such a discussion. I'd prefer to keep it this way as it makes the code much easier.

PR #617, opened 2020-03-13
revert intensification behaviour for first configuration

ashraaghav:
Intensification now evaluates the 1st incumbent twice, just like previous versions of SMAC

### DIFF
--- a/test/test_intensify/test_intensify.py
+++ b/test/test_intensify/test_intensify.py
@@ -489,7 +489,6 @@ class TestIntensify(unittest.TestCase):
         """
             test eval_challenger() - a complete intensification run without a `always_race_against` configuration
         """
-
         def target(x):
             return 2 * x['a'] + x['b']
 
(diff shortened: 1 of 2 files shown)


test/test_intensify/test_intensify.py
```diff
@@ -389,29 +389,40 @@ def test_eval_challenger_1(self):
...
+        self.assertEqual(intensifier.stage, IntensifierStage.RUN_FIRST_CONFIG)
         # eval config 2 (=first run)
         inc, _ = intensifier.eval_challenger(challenger=config, incumbent=None, run_history=self.rh, )
         self.assertEqual(inc, self.config2)
+        self.assertEqual(intensifier.stage, IntensifierStage.RUN_INCUMBENT)
+        self.assertEqual(self.stats.inc_changed, 1)
+
+        # regular intensification begins - run incumbent first
+        config, _ = intensifier.get_next_challenger(challengers=None,  # don't need a new list here as old one is cont'd
+                                                    chooser=None)
```
mfeurer: Could you please add a check here that the config is the same as the variable `inc`?

test/test_intensify/test_intensify.py
```diff
@@ -449,28 +473,25 @@ def target(x):
...
-        self.assertEqual(self.stats.inc_changed, 3)
+        self.assertEqual(inc, self.config1)
         self.assertEqual(intensifier.stage, IntensifierStage.RUN_CHALLENGER)
 
-        self.assertEqual(len(self.rh.get_runs_for_config(self.config1, only_max_observed_budget=True)), 1)
-        self.assertEqual(len(self.rh.get_runs_for_config(self.config2, only_max_observed_budget=True)), 1)
-        self.assertEqual(len(self.rh.get_runs_for_config(self.config3, only_max_observed_budget=True)), 2)
+        self.assertEqual(len(self.rh.get_runs_for_config(self.config1, only_max_observed_budget=True)), 3)
+        self.assertEqual(len(self.rh.get_runs_for_config(self.config2, only_max_observed_budget=True)), 2)
+        self.assertEqual(len(self.rh.get_runs_for_config(self.config3, only_max_observed_budget=True)), 0)  # capped
```
mfeurer: I'm wondering if a capped run should be excluded from this function. It might be surprising to not get such runs when asking for them. @mlindauer @KEggensperger do you have any suggestions?

test/test_intensify/test_intensify.py
```diff
@@ -481,28 +502,42 @@ def target(x):
...
-        self.assertEqual(inc, self.config2)
-        self.assertEqual(intensifier.stage, IntensifierStage.RUN_CHALLENGER)
+        self.assertEqual(inc, self.config3)
+        self.assertEqual(intensifier.stage, IntensifierStage.RUN_INCUMBENT)
         self.assertEqual(self.stats.inc_changed, 1)
 
+        # regular intensification begins - run incumbent
+        config, _ = intensifier.get_next_challenger(challengers=None,  # since incumbent is run, no configs required
+                                                    chooser=None)
+        self.assertEqual(intensifier.stage, IntensifierStage.RUN_INCUMBENT)
```
mfeurer: Isn't this one run too much?

test/test_intensify/test_intensify.py
```diff
@@ -515,25 +550,18 @@ def target(x):
         # run incumbent
         config, _ = intensifier.get_next_challenger(challengers=None,  # since incumbent is run, no configs required
                                                     chooser=None)
-        self.assertEqual(config, self.config1)
+        self.assertEqual(config, self.config2)
         self.assertEqual(intensifier.stage, IntensifierStage.RUN_INCUMBENT)
         inc, _ = intensifier.eval_challenger(challenger=config, incumbent=inc, run_history=self.rh, )
 
-        self.assertEqual(inc, self.config1)
```
mfeurer: Is there a reason why you removed these checks?

ashraaghav: We already test the same thing in the same test case, so I thought it was redundant.

test/test_intensify/test_intensify.py
ashraaghav: Do you mean that this incumbent configuration is evaluated again? It won't be run because it doesn't have any runs left (1 instance & deterministic). So the TA runs remain the same.

mfeurer: Okay, and these are tested below. So I guess it's alright then.

PR #632, opened 2020-03-26
Add converged state

KEggensperger:
1. Add a new StatusType, DONOTADVANCE, which indicates that this configuration might or might not have a result, but should not be considered further, e.g. advance to the next budget for SH.

1. When using budgets, we will always use all configuration with this new type to build a model regardless of the budget they were run on.

Furthermore this PR:

- Clarifies which states in runhistory2EPM are considered successful or used for imputation by removing default values
- Correctly considers budgets in runhistory2EPM when constructing the matrices
- By default also uses MEMOUTs to build the model when running AC and optimizing for quality


smac/facade/smac_ac_facade.py
```diff
@@ -492,6 +494,15 @@ def __init__(self,
                 'impute_censored_data': False,
                 'impute_state': None,
             })
+
+        if isinstance(intensifier, SuccessiveHalving) or isinstance(intensifier, Hyperband) \
```
mfeurer: Hey, you need to only check for SuccessiveHalving, Hyperband is a subclass of that.

smac/runhistory/runhistory2epm.py
```diff
@@ -65,6 +66,8 @@ def __init__(
             If None, set to [StatusType.SUCCESS, ]
         impute_censored_data: bool, optional
             Should we impute data?
+        consider_for_higher_budgets_state: list, optional
+            Additionally consider all runs with these states for budget < current budget
```
mfeurer: Shouldn't it be `budget > current budget`?

test/test_facade/test_smac_facade.py
```diff
@@ -81,6 +91,36 @@ def test_pass_tae_runner_objective(self):
...
+        self.assertSetEqual(set(smbo.solver.epm_chooser.rh2EPM.impute_state), set())
+        self.assertSetEqual(set(smbo.solver.epm_chooser.rh2EPM.consider_for_higher_budgets_state),
+                            set())
+
+        for intensifier in (SuccessiveHalving, Hyperband):
+            smbo = SMAC4AC(self.scenario, intensifier=intensifier,
+                           intensifier_kwargs=self.sh_intensifier_kwargs)
+            self.assertTrue(type(smbo.solver.epm_chooser.rh2EPM) == RunHistory2EPM4Cost)
+            self.assertSetEqual(set(smbo.solver.epm_chooser.rh2EPM.success_states),
+                                {StatusType.SUCCESS, StatusType.CRASHED, StatusType.MEMOUT})
```
mfeurer: Maybe you also want to check for the states to advance?

smac/facade/smac_ac_facade.py
mfeurer: Also, I just realized that you need to check for `issubclass` as intensifier is actually a class, not an object.

KEggensperger: I should check whether `intensifier_instance` is a subclass of `SuccessiveHalving`

smac/runhistory/runhistory2epm.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         self,
         scenario: Scenario,
         num_params: int,
-        success_states: typing.Optional[typing.List[StatusType]] = None,
+        success_states: typing.List[StatusType] = None,
```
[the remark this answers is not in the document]
KEggensperger: agree

test/test_facade/test_smac_facade.py
KEggensperger: What do you mean? Isn't line 111 doing this?

PR #638, opened 2020-04-06
Make GP predict faster

ashraaghav:


smac/epm/base_epm.py
```diff
@@ -202,23 +211,27 @@ def predict(self, X: np.ndarray) -> typing.Tuple[np.ndarray, np.ndarray]:
...
-        if len(var.shape) == 1:
+        if var is not None and len(var.shape) == 1:
             var = var.reshape((-1, 1))
 
         return mean, var
 
-    def _predict(self, X: np.ndarray) -> typing.Tuple[np.ndarray, np.ndarray]:
+    def _predict(self, X: np.ndarray,
+                 cov_return_type: typing.Optional[str] = 'diagonal_cov') \
+            -> typing.Tuple[np.ndarray, np.ndarray]:
```
mfeurer, line 225: And here, too.

smac/epm/gaussian_process.py
```diff
@@ -237,7 +237,9 @@ def _optimize(self) -> np.ndarray:
                 theta_star = theta
         return theta_star
 
-    def _predict(self, X_test: np.ndarray, full_cov: bool = False) -> typing.Tuple[np.ndarray, np.ndarray]:
+    def _predict(self, X_test: np.ndarray,
+                 cov_return_type: typing.Optional[str] = 'diagonal_cov') \
+            -> typing.Tuple[np.ndarray, np.ndarray]:
```
mfeurer, line 242: And here, too.

smac/epm/gaussian_process.py
```diff
@@ -262,15 +264,30 @@ def _predict(self, X_test: np.ndarray, full_cov: bool = False) -> typing.Tuple[n
             raise Exception('Model has to be trained first!')
 
         X_test = self._impute_inactive(X_test)
-        mu, var = self.gp.predict(X_test, return_cov=True)
-        var = np.diag(var)
 
-        # Clip negative variances and set them to the smallest
-        # positive float value
-        var = np.clip(var, VERY_SMALL_NUMBER, np.inf)
+        predict_kwargs = {'return_cov': False, 'return_std': True}
```
mfeurer, line 268: Could this be moved into the else block? It looks like it belongs rather there than here.

smac/epm/gaussian_process.py
```diff
@@ -262,15 +264,30 @@ def _predict(self, X_test: np.ndarray, full_cov: bool = False) -> typing.Tuple[n
...
+
+            if cov_return_type == 'diagonal_cov':
+                var = var ** 2  # since we get standard deviation for faster computation
+
+            # Clip negative variances and set them to the smallest
+            # positive float value
+            var = np.clip(var, VERY_SMALL_NUMBER, np.inf)
+
+            if self.normalize_y:
+                mu, var = self._untransform_y(mu, var)
```
mfeurer, line 290: Hm, I don't think this is correct if we want to return the standard deviation, as it assumes that this is the variance that needs un-normalization. Could you please have a look here?

smac/epm/base_epm.py
```diff
@@ -169,14 +169,23 @@ def _train(self, X: np.ndarray, Y: np.ndarray) -> 'AbstractEPM':
         """
         raise NotImplementedError
 
-    def predict(self, X: np.ndarray) -> typing.Tuple[np.ndarray, np.ndarray]:
+    def predict(self, X: np.ndarray,
+                cov_return_type: typing.Optional[str] = 'diagonal_cov') \
+            -> typing.Union[np.ndarray, typing.Tuple[np.ndarray, np.ndarray]]:
```
mfeurer, line 174: Is this correct? According to the GP unit test, the return value should be `typing.Union[np.ndarray, typing.Optional[np.ndarray]]`, right?

smac/epm/base_epm.py
```diff
@@ -169,14 +169,23 @@ def _train(self, X: np.ndarray, Y: np.ndarray) -> 'AbstractEPM':
...
+        cov_return_type: typing.Optional[str]
+            Specifies what to return along with the mean. (Applies to only Gaussian Process for now)
+            Can take 4 values: [None, diagonal_std, diagonal_cov, full_cov]
+            * None - only mean is returned
+            * diagonal_std - standard deviation at test points is returned
+            * diagonal_cov - diagonal of the covariance matrix is returned
+            * full_cov - whole covariance matrix between the test points is returned
 
         Returns
         -------
```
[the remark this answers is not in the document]
mfeurer: Could you please update the docstring of the return value, too?

smac/epm/gaussian_process.py
```diff
@@ -246,31 +248,47 @@ def _predict(self, X_test: np.ndarray, full_cov: bool = False) -> typing.Tuple[n
...
+            if self.normalize_y:
+                mu = self._untransform_y(mu)
+
+        else:
+            predict_kwargs = {'return_cov': False, 'return_std': True}
+            if cov_return_type == 'full_cov':
+                predict_kwargs = {'return_cov': True, 'return_std': False}
+
+            mu, var = self.gp.predict(X_test, **predict_kwargs)
+            var = var ** 2  # since we get standard deviation for faster computation
```
mfeurer, line 281: This shouldn't be done for the full covariance matrix case, right?

test/test_epm/test_gp.py
```diff
@@ -294,6 +294,12 @@ def test_predict_with_actual_values(self):
         # There's a slight difference between my local installation and travis
         self.assertLess(abs(var_hat[0][0] - 1121.8409184001594), 2)
 
+        # test other covariance results
+        _, var_hat = model.predict(X, cov_return_type='full_cov')
```
mfeurer, line 298: Could you please check that the returned variances/stds for all three cases which are not none are numerically similar? I would like to make sure that we don't introduce stupid bugs here in the future.

smac/epm/base_epm.py
ashraaghav: Then it should be `typing.Tuple[np.ndarray, typing.Optional[np.ndarray]]`, right?

mfeurer: Yes, of course.

Issue #637, opened 2020-04-06
TAE and TAE kwargs are not passed down in hydra and psmac


kstrauch94:
#### Description
When using hydra and a custom TAE it will fail because it doesn't pass the TAE and its kwargs correctly. The exact spot in the [hydra facade](https://github.com/automl/SMAC3/blob/master/smac/facade/experimental/hydra_facade.py) is in line 191. Here we pass down the TAE kwargs but not the tae. I think that simply adding
`tae_kwargs["tae"] = self._tae` 
below line 191 can fix this. 

The second place where this happens is in the [psmac facade](https://github.com/automl/SMAC3/blob/master/smac/facade/experimental/psmac_facade.py) in line 255. Here, we pass neither the TAE or its kwargs.

#### Steps/Code to Reproduce
Run hydra with a custom TAE with more than the regular TAE arguments

#### Expected Results
smac runs smoothly

#### Actual Results
smac crashes

mlindauer:
@AndreBiedenkapp do you have time to look into it?

AndreBiedenkapp:
I'll look into it but it might be some time before I can get to it.

kstrauch94:
I opened a pull request that should fix these issues. Hopefully, this makes them clearer.

alexandertornede:
We are very sorry that the stalebot closed the PR and that no one had the time to take care of this for over 2 years now.

However, psmac is not supported in 2.0 anymore and instead we support parallelism natively now. Furthermore, the hydra interface is redesigned for 2.0. Thus, we will close this issue.

PR #641, opened 2020-04-10
Fix issue 637 and problems that arise after that

kstrauch94:
This pull request fixes the problems mentioned in issue [637](https://github.com/automl/SMAC3/issues/637) along with other issues that arise after fixing those.

- Fix problems in  issue [637](https://github.com/automl/SMAC3/issues/637)
- Retrieve selected incumbents with correct syntax
- "Bandaid" fix to runhistory RunKey sometimes having a str type budget leading to problems in the Runhistory._add function
- ExecuteTARunHydra should now support TAE that have more arguments than the default TAE


dengdifan:
Overall it looks good to me @AndreBiedenkapp @mfeurer @KEggensperger  any comments?

PR #654, opened 2020-05-22
Fix infinite loop

mfeurer:
Fixes an infinite loop in intensification which could occur if the acquisition function would always return the same configuration and no random configurations were used. The fix is to only end intensify if there are either no challengers left or one challenger was executed (and therefore the model would change for the next iteration).

### DIFF
--- a/smac/intensification/intensification.py
+++ b/smac/intensification/intensification.py
@@ -126,6 +126,7 @@ class Intensifier(AbstractRacer):
 
         # challenger related variables
         self._chall_indx = 0
+        self.num_chall_run = 0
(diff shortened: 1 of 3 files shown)


smac/intensification/intensification.py
```diff
@@ -294,6 +301,7 @@ def _add_inc_run(self,
                 next_seed = self.rs.randint(low=0, high=MAXINT, size=1)[0]
 
             if available_insts:
+                self.num_run += 1
```
KEggensperger, line 304: Is there a reason to increase num_run before running the next config?

PR #663, opened 2020-06-28
Separated next challenger form eval

franchuterivera:
Separated evaluation from get next challengers in intensification.

Created an object called runinfo that contains the necessary information for an isolated run.

The intensifier is now a state machine with the main stages given by (also in old code):
    RUN_FIRST_CONFIG = 0  # to replicate the old initial design
     RUN_INCUMBENT = 1  # Lines 3-7
     RUN_CHALLENGER = 2  # Lines 8-17
     RUN_BASIS = 3
and substrates given by (new):
         self.challenger_same_as_incumbent = False
         self.no_available_insts = False
        self.to_run_empty = False

Readjusted the test to have the same functionality. Verified also via examples executing the same line of code using a profiler. Fundamentally, this change can be seen as a reordering such that:

- Eval challenger code was mostly moved to get challengers to be able to obtain the seed/instance
- Added a new function called process results to update the state machine before next iteration

### DIFF
--- a/smac/facade/smac_ac_facade.py
+++ b/smac/facade/smac_ac_facade.py
@@ -386,7 +386,6 @@ class SMAC4AC(object):
 
         # initialize intensification
         intensifier_def_kwargs = {
-            'tae_runner': tae_runner_instance,
             'stats': self.stats,
             'traj_logger': traj_logger,
             'rng': rng,
@@ -544,7 +543,8 @@ class SMAC4AC(object):
             'acquisition_func': acquisition_function_instance,
             'rng': rng,
             'restore_incumbent': restore_incumbent,
-            'random_configuration_chooser': random_configuration_chooser_instance
+            'random_configuration_chooser': random_configuration_chooser_instance,
+            'tae_runner': tae_runner_instance,
         }  # type: Dict[str, Any]
 
         if smbo_class is None:
@@ -631,7 +631,7 @@ class SMAC4AC(object):
         TAE: smac.tae.execute_ta_run.ExecuteTARun
 
         """
-        return self.solver.intensifier.tae_runner
+        return self.solver.tae_runner
 
     def get_runhistory(self) -> RunHistory:
         """
--- a/smac/facade/smac_ac_facade.py
+++ b/smac/facade/smac_ac_facade.py
@@ -349,7 +349,6 @@ class SMAC4AC(object):
         tae_def_kwargs = {
             'stats': self.stats,
             'run_obj': scenario.run_obj,
-            'runhistory': runhistory,
             'par_factor': scenario.par_factor,  # type: ignore[attr-defined] # noqa F821
             'cost_for_crash': scenario.cost_for_crash,  # type: ignore[attr-defined] # noqa F821
             'abort_on_first_run_crash': scenario.abort_on_first_run_crash  # type: ignore[attr-defined] # noqa F821
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -2,6 +2,7 @@ import logging
 import typing
 import time
 from collections import OrderedDict
+from enum import Enum
 
 import numpy as np
 
@@ -18,6 +19,19 @@ _config_to_run_type = typing.Iterator[typing.Optional[Configuration]]
 __author__ = "Ashwin Raaghav Narayanan"
 __copyright__ = "Copyright 2019, ML4AAD"
 __license__ = "3-clause BSD"
+
+
+class IntensifierBehest(Enum):
+    """Class to define different request on how to process
+    the runinfo
+
+    Gives the flexibility to indicate whether no more configs
+    are available (TERMINATE in this case), no need to rerun
+    previous configurations (SKIP) or we simple should run
+    """
+    RUN = 0  # Normal run execution of a run info
+    SKIP = 1  # Skip running the run info
+    TERMINATE = 2  # No more configurations to try
 
 
 class AbstractRacer(object):
@@ -115,7 +129,8 @@ class AbstractRacer(object):
                      incumbent: Configuration,
                      chooser: typing.Optional[EPMChooser],
                      run_history: RunHistory,
-                     repeat_configs: bool = True) -> RunInfo:
+                     repeat_configs: bool = True
+                     ) -> typing.Tuple[IntensifierBehest, RunInfo]:
         """
         Abstract method for choosing the next challenger, to allow for different selections across intensifiers
         uses ``_next_challenger()`` by default
@@ -139,6 +154,8 @@ class AbstractRacer(object):
         -------
         run_info: RunInfo
             An object that encapsulates necessary information for a config run
+        behest: IntensifierBehest
+            Indicator of how to consume the RunInfo object
         """
         raise NotImplementedError()
 
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -25,23 +25,25 @@ class RunInfoIntent(Enum):
     """Class to define different requests on how to process
     the runinfo
 
-    Gives the flexibility to indicate whether no more configs
-    are available (STOP_ITERATION in this case), no need to rerun
-    previous configurations (SKIP) or if the SMBO should simple
+    Gives the flexibility to indicate whether a configuration
+    should be skipped (SKIP) or if the SMBO should simple
     run a generated run_info.
     """
     RUN = 0  # Normal run execution of a run info
     SKIP = 1  # Skip running the run_info
-    STOP_ITERATION = 2  # No more configurations to try
 
 
 class AbstractRacer(object):
     """
     Base class for all racing methods
 
-    The "intensification" is designed to be spread across multiple ``eval_challenger()`` runs.
-    This is to facilitate on-demand configuration sampling if multiple configurations are required,
-    like Successive Halving or Hyperband.
+    The "intensification" is designed to output a RunInfo object with enough information
+    to run a given configuration (for example, the run info contains the instance/seed
+    pair, as well as the associated resources).
+
+    A worker can execute this RunInfo object and produce a RunValue object with the
+    execution results. Each intensifier process the RunValue object and updates it's
+    internal state in preparation for the next iteration.
 
     **Note: Do not use directly**
 
@@ -136,7 +138,8 @@ class AbstractRacer(object):
         Abstract method for choosing the next challenger, to allow for different selections across intensifiers
         uses ``_next_challenger()`` by default
 
-        If no more challengers are available, the method should return None as configuration
+        If no more challengers are available, the method should issue a SKIP via
+        RunInfoIntent.SKIP, so that a new iteration can sample new configurations.
 
         Parameters
         ----------
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -110,15 +110,17 @@ class AbstractRacer(object):
         # to mark the end of an iteration
         self.iteration_done = False
 
-    def get_next_challenger(self,
-                            challengers: typing.Optional[typing.List[Configuration]],
-                            incumbent: Configuration,
-                            chooser: typing.Optional[EPMChooser],
-                            run_history: RunHistory,
-                            repeat_configs: bool = True) -> RunInfo:
+    def get_next_run(self,
+                     challengers: typing.Optional[typing.List[Configuration]],
+                     incumbent: Configuration,
+                     chooser: typing.Optional[EPMChooser],
+                     run_history: RunHistory,
+                     repeat_configs: bool = True) -> RunInfo:
         """
         Abstract method for choosing the next challenger, to allow for different selections across intensifiers
         uses ``_next_challenger()`` by default
+
+        If no more challengers are available, the method should return None as configuration
 
         Parameters
         ----------
@@ -170,7 +172,7 @@ class AbstractRacer(object):
             The tracked time of a configuration execution
         time_bound : float, optional (default=2 ** 31 - 1)
             time in [sec] available to perform intensify
-        status:  typing.Optional[StatusType],
+        status:  typing.Optional[StatusType]
             The status of the execution of a given config
         runtime:
             The elapsed time according to the ta runner
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -10,8 +10,8 @@ from smac.optimizer.epm_configuration_chooser import EPMChooser
 from smac.stats.stats import Stats
 from smac.utils.constants import MAXINT
 from smac.configspace import Configuration
-from smac.runhistory.runhistory import RunHistory
-from smac.tae.execute_ta_run import ExecuteTARun
+from smac.runhistory.runhistory import RunHistory, RunInfo, StatusType
+from smac.tae.execute_ta_run import ExecuteTARun, CappedRunException, BudgetExhaustedException
 from smac.utils.io.traj_logging import TrajLogger
 
 _config_to_run_type = typing.Iterator[typing.Optional[Configuration]]
@@ -117,43 +117,75 @@ class AbstractRacer(object):
         # to mark the end of an iteration
         self.iteration_done = False
 
-    def eval_challenger(self,
-                        challenger: Configuration,
-                        incumbent: typing.Optional[Configuration],
-                        run_history: RunHistory,
-                        time_bound: float = float(MAXINT),
-                        log_traj: bool = True) -> typing.Tuple[Configuration, float]:
-        """
-        Runs intensification by evaluating one configuration-instance at a time
+    def eval_challenger(
+        self,
+        run_info: RunInfo,
+        time_bound: float = float(MAXINT),
+    ) -> typing.Tuple[StatusType, typing.Any, float, dict]:
+        """Runs a configuration with the provided input.
         *Side effect:* adds runs to run_history
 
         Parameters
         ----------
-        challenger : Configuration
-            promising configuration
-        incumbent : typing.Optional[Configuration]
-            best configuration so far, None in 1st run
-        run_history : smac.runhistory.runhistory.RunHistory
-            stores all runs we ran so far
+        run_info : RunInfo
+            An object that specifies the inputs to a tae runner
         time_bound : float, optional (default=2 ** 31 - 1)
             time in [sec] available to perform intensify
-        log_traj : bool
-            whether to log changes of incumbents in trajectory
 
         Returns
         -------
-        typing.Tuple[Configuration, float]
-            incumbent and incumbent cost
+        status: StatusType
+            The status of the execution of a given config
+            {SUCCESS, TIMEOUT/CAPPED, CRASHED, ABORT}
+        cost: typing.Any
+            cost/regret/quality (typing.Any)
+        duration: float
+            runtime (None if not returned by TA)
+        res: Dict
+            additional info
         """
 
-        raise NotImplementedError()
+        if time_bound < self._min_time:
+            raise ValueError("time_bound must be >= %f" % self._min_time)
+
+        if run_info.config is None or run_info.config is None:
+            raise ValueError("Call to eval configurations without any valid config")
+
+        try:
+            # Cap makes sense only in challenger. For incumbent runs, capped will be
+            # false. For a challenger, the run_info.cutoff will be smaller than self.cutoff
+            capped = False
+            if (self.cutoff is not None) and (run_info.cutoff < self.cutoff):
+                capped = True
+
+            status, cost, dur, res = self.tae_runner.start(
+                config=run_info.config,
+                instance=run_info.instance,
+                seed=run_info.seed,
+                cutoff=run_info.cutoff,
+                budget=run_info.budget,
+                instance_specific=self.instance_specifics.get(run_info.instance, "0"),
+                capped=capped
+            )
+
+        except CappedRunException:
+            self.logger.debug("Challenger itensification timed out due "
+                              "to adaptive capping.")
+            status, cost, dur, res = StatusType.CAPPED, float(MAXINT), 0.0, {}
+
+        except BudgetExhaustedException:
+            # SMBO stops due to its own budget checks
+            self.logger.debug("Budget exhausted; Return incumbent")
+            status, cost, dur, res = StatusType.CRASHED, float(MAXINT), 0.0, {}
+
+        return status, cost, dur, res
 
     def get_next_challenger(self,
                             challengers: typing.Optional[typing.List[Configuration]],
+                            incumbent: Configuration,
                             chooser: typing.Optional[EPMChooser],
                             run_history: RunHistory,
-                            repeat_configs: bool = True) -> \
-            typing.Tuple[typing.Optional[Configuration], bool]:
+                            repeat_configs: bool = True) -> RunInfo:
         """
         Abstract method for choosing the next challenger, to allow for different selections across intensifiers
         uses ``_next_challenger()`` by default
@@ -162,6 +194,8 @@ class AbstractRacer(object):
         ----------
         challengers : typing.List[Configuration]
             promising configurations
+        incumbent: Configuration
+             incumbent configuration
         chooser : smac.optimizer.epm_configuration_chooser.EPMChooser
             optimizer that generates next configurations to use for racing
         run_history : smac.runhistory.runhistory.RunHistory
@@ -171,16 +205,56 @@ class AbstractRacer(object):
 
         Returns
         -------
-        typing.Optional[Configuration]
-            next configuration to evaluate
-        bool
-            flag telling if the configuration is newly sampled or one currently being tracked
+        run_info: RunInfo
+            An object that encapsulates necessary information for a config run
         """
-        challenger = self._next_challenger(challengers=challengers,
-                                           chooser=chooser,
-                                           run_history=run_history,
-                                           repeat_configs=repeat_configs)
-        return challenger, True
+        raise NotImplementedError()
+
+    def process_results(self,
+                        challenger: Configuration,
+                        incumbent: typing.Optional[Configuration],
+                        run_history: RunHistory,
+                        elapsed_time: float,
+                        time_bound: float,
+                        status: StatusType,
+                        runtime: float,
+                        log_traj: bool = True,
+                        ) -> \
+            typing.Tuple[Configuration, float]:
+        """
+        The intensifier stage will be updated based on the results/status
+        of a configuration execution.
+        Also, a incumbent will be determined.
+
+        Parameters
+        ----------
+        challenger : Configuration
+            A configuration to challenge the incumbent. Can even be the incumbent
+            to gain more confidence on it.
+        incumbet : Configuration
+            Best configuration seen so far
+        run_history : typing.Optional[smac.runhistory.runhistory.RunHistory]
+            stores all runs we ran so far
+            if False, an evaluated configuration will not be generated again
+        elapsed_time:
+            The tracked time of a configuration execution
+        time_bound : float, optional (default=2 ** 31 - 1)
+            time in [sec] available to perform intensify
+        status: StatusType
+            The status of the execution of a given config
+        runtime:
+            The elapsed time according to the ta runner
+        log_traj: bool
+            Whether to log changes of incumbents in trajectory
+
+        Returns
+        -------
+        incumbent: Configuration()
+            current (maybe new) incumbent configuration
+        inc_perf: float
+            empirical performance of incumbent configuration
+        """
+        raise NotImplementedError()
 
     def _next_challenger(self,
                          challengers: typing.Optional[typing.List[Configuration]],
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -21,17 +21,18 @@ __copyright__ = "Copyright 2019, ML4AAD"
 __license__ = "3-clause BSD"
 
 
-class IntensifierBehest(Enum):
-    """Class to define different request on how to process
+class RunInfoIntent(Enum):
+    """Class to define different requests on how to process
     the runinfo
 
     Gives the flexibility to indicate whether no more configs
-    are available (TERMINATE in this case), no need to rerun
-    previous configurations (SKIP) or we simple should run
+    are available (STOP_ITERATION in this case), no need to rerun
+    previous configurations (SKIP) or if the SMBO should simple
+    run a generated run_info.
     """
     RUN = 0  # Normal run execution of a run info
-    SKIP = 1  # Skip running the run info
-    TERMINATE = 2  # No more configurations to try
+    SKIP = 1  # Skip running the run_info
+    STOP_ITERATION = 2  # No more configurations to try
 
 
 class AbstractRacer(object):
@@ -130,7 +131,7 @@ class AbstractRacer(object):
                      chooser: typing.Optional[EPMChooser],
                      run_history: RunHistory,
                      repeat_configs: bool = True
-                     ) -> typing.Tuple[IntensifierBehest, RunInfo]:
+                     ) -> typing.Tuple[RunInfoIntent, RunInfo]:
         """
         Abstract method for choosing the next challenger, to allow for different selections across intensifiers
         uses ``_next_challenger()`` by default
@@ -154,7 +155,7 @@ class AbstractRacer(object):
         -------
         run_info: RunInfo
             An object that encapsulates necessary information for a config run
-        behest: IntensifierBehest
+        intent: RunInfoIntent
             Indicator of how to consume the RunInfo object
         """
         raise NotImplementedError()
@@ -178,12 +179,12 @@ class AbstractRacer(object):
         challenger : Configuration
             A configuration that was previously executed, and whose status
             will be used to define the next stage.
-        incumbent : Configuration
+        incumbent : typing.Optional[Configuration]
             Best configuration seen so far
-        run_history : typing.Optional[smac.runhistory.runhistory.RunHistory]
+        run_history : RunHistory
             stores all runs we ran so far
             if False, an evaluated configuration will not be generated again
-        time_bound : float, optional (default=2 ** 31 - 1)
+        time_bound : float
             time in [sec] available to perform intensify
         result: RunValue
             Contain the result (status and other methadata) of exercising
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -9,7 +9,7 @@ from smac.optimizer.epm_configuration_chooser import EPMChooser
 
 from smac.stats.stats import Stats
 from smac.configspace import Configuration
-from smac.runhistory.runhistory import RunHistory, RunInfo, StatusType
+from smac.runhistory.runhistory import RunHistory, RunInfo, RunValue
 from smac.utils.io.traj_logging import TrajLogger
 
 _config_to_run_type = typing.Iterator[typing.Optional[Configuration]]
@@ -146,10 +146,8 @@ class AbstractRacer(object):
                         challenger: Configuration,
                         incumbent: typing.Optional[Configuration],
                         run_history: RunHistory,
-                        elapsed_time: float,
                         time_bound: float,
-                        status: StatusType,
-                        runtime: float,
+                        result: RunValue,
                         log_traj: bool = True,
                         ) -> \
             typing.Tuple[Configuration, float]:
@@ -168,14 +166,11 @@ class AbstractRacer(object):
         run_history : typing.Optional[smac.runhistory.runhistory.RunHistory]
             stores all runs we ran so far
             if False, an evaluated configuration will not be generated again
-        elapsed_time:
-            The tracked time of a configuration execution
         time_bound : float, optional (default=2 ** 31 - 1)
             time in [sec] available to perform intensify
-        status:  typing.Optional[StatusType]
-            The status of the execution of a given config
-        runtime:
-            The elapsed time according to the ta runner
+        result: RunValue
+            Contain the result (status and other methadata) of exercising
+            a challenger/incumbent.
         log_traj: bool
             Whether to log changes of incumbents in trajectory
 
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -8,10 +8,8 @@ import numpy as np
 from smac.optimizer.epm_configuration_chooser import EPMChooser
 
 from smac.stats.stats import Stats
-from smac.utils.constants import MAXINT
 from smac.configspace import Configuration
 from smac.runhistory.runhistory import RunHistory, RunInfo, StatusType
-from smac.tae.execute_ta_run import ExecuteTARun, CappedRunException, BudgetExhaustedException
 from smac.utils.io.traj_logging import TrajLogger
 
 _config_to_run_type = typing.Iterator[typing.Optional[Configuration]]
@@ -34,8 +32,6 @@ class AbstractRacer(object):
 
     Parameters
     ----------
-    tae_runner : tae.executre_ta_run_*.ExecuteTARun* Object
-        target algorithm run executor
     stats: Stats
         stats object
     traj_logger: smac.utils.io.traj_logging.TrajLogger
@@ -64,7 +60,6 @@ class AbstractRacer(object):
     """
 
     def __init__(self,
-                 tae_runner: ExecuteTARun,
                  stats: Stats,
                  traj_logger: TrajLogger,
                  rng: np.random.RandomState,
@@ -89,7 +84,6 @@ class AbstractRacer(object):
         self.cutoff = cutoff
         self.deterministic = deterministic
         self.run_obj_time = run_obj_time
-        self.tae_runner = tae_runner
 
         self.minR = minR
         self.maxR = maxR
@@ -107,7 +101,6 @@ class AbstractRacer(object):
             self.instance_specifics = instance_specifics
 
         # general attributes
-        self._min_time = 10 ** -5
         self.num_run = 0  # Number of runs done in an iteration so far
         self._chall_indx = 0
         self._ta_time = 0.
@@ -116,69 +109,6 @@ class AbstractRacer(object):
         self.repeat_configs = True
         # to mark the end of an iteration
         self.iteration_done = False
-
-    def eval_challenger(
-        self,
-        run_info: RunInfo,
-        time_bound: float = float(MAXINT),
-    ) -> typing.Tuple[StatusType, typing.Any, float, dict]:
-        """Runs a configuration with the provided input.
-        *Side effect:* adds runs to run_history
-
-        Parameters
-        ----------
-        run_info : RunInfo
-            An object that specifies the inputs to a tae runner
-        time_bound : float, optional (default=2 ** 31 - 1)
-            time in [sec] available to perform intensify
-
-        Returns
-        -------
-        status: StatusType
-            The status of the execution of a given config
-            {SUCCESS, TIMEOUT/CAPPED, CRASHED, ABORT}
-        cost: typing.Any
-            cost/regret/quality (typing.Any)
-        duration: float
-            runtime (None if not returned by TA)
-        res: Dict
-            additional info
-        """
-
-        if time_bound < self._min_time:
-            raise ValueError("time_bound must be >= %f" % self._min_time)
-
-        if run_info.config is None or run_info.config is None:
-            raise ValueError("Call to eval configurations without any valid config")
-
-        try:
-            # Cap makes sense only in challenger. For incumbent runs, capped will be
-            # false. For a challenger, the run_info.cutoff will be smaller than self.cutoff
-            capped = False
-            if (self.cutoff is not None) and (run_info.cutoff < self.cutoff):
-                capped = True
-
-            status, cost, dur, res = self.tae_runner.start(
-                config=run_info.config,
-                instance=run_info.instance,
-                seed=run_info.seed,
-                cutoff=run_info.cutoff,
-                budget=run_info.budget,
-                instance_specific=self.instance_specifics.get(run_info.instance, "0"),
-                capped=capped
-            )
-
-        except CappedRunException:
-            self.logger.debug("Challenger itensification timed out due "
-                              "to adaptive capping.")
-            status, cost, dur, res = StatusType.CAPPED, float(MAXINT), 0.0, {}
-
-        except BudgetExhaustedException:
-            # SMBO stops due to its own budget checks
-            self.logger.debug("Budget exhausted; Return incumbent")
-            status, cost, dur, res = StatusType.CRASHED, float(MAXINT), 0.0, {}
-
-        return status, cost, dur, res
 
     def get_next_challenger(self,
                             challengers: typing.Optional[typing.List[Configuration]],
@@ -229,9 +159,9 @@ class AbstractRacer(object):
         Parameters
         ----------
         challenger : Configuration
-            A configuration to challenge the incumbent. Can even be the incumbent
-            to gain more confidence on it.
-        incumbet : Configuration
+            A configuration that was previously executed, and whose status
+            will be used to define the next stage.
+        incumbent : Configuration
             Best configuration seen so far
         run_history : typing.Optional[smac.runhistory.runhistory.RunHistory]
             stores all runs we ran so far
@@ -240,7 +170,7 @@ class AbstractRacer(object):
             The tracked time of a configuration execution
         time_bound : float, optional (default=2 ** 31 - 1)
             time in [sec] available to perform intensify
-        status: StatusType
+        status:  typing.Optional[StatusType],
             The status of the execution of a given config
         runtime:
             The elapsed time according to the ta runner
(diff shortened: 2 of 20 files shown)


smac/intensification/abstract_racer.py
```diff
@@ -117,43 +117,75 @@ def __init__(self,
...
+                cutoff=run_info.cutoff,
+                budget=run_info.budget,
+                instance_specific=self.instance_specifics.get(run_info.instance, "0"),
+                capped=capped
+            )
+
+        except CappedRunException:
+            self.logger.debug("Challenger itensification timed out due "
+                              "to adaptive capping.")
+            status, cost, dur, res = StatusType.CAPPED, float(MAXINT), 0.0, {}
```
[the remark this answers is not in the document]
franchuterivera: isn't it MAXINT? because that is defined to be the cost for a crash:
smac/tae/execute_func.py:        cost_for_crash: float = float(MAXINT)

Thanks, let me change the 0.0 to be the run_info.cutoff, as it is more informative

smac/intensification/abstract_racer.py
```diff
@@ -117,43 +117,75 @@ def __init__(self,
...
+
+        except BudgetExhaustedException:
+            # SMBO stops due to its own budget checks
+            self.logger.debug("Budget exhausted; Return incumbent")
+            status, cost, dur, res = StatusType.CRASHED, float(MAXINT), 0.0, {}
```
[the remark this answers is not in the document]
franchuterivera: I see what you mean. I cannot find an existing status for this. I will treat this scenario as a None, because it is not a crash/abort/fail. 

The if the status is None, the pertinent checks will be the same as before

smac/intensification/successive_halving.py
```diff
@@ -273,143 +273,112 @@ def _init_sh_params(self,
...
-            cutoff = self._adapt_cutoff(challenger=challenger,
-                                        run_history=run_history,
-                                        inc_sum_cost=inc_sum_cost)
-            if cutoff is not None and cutoff <= 0:
-                # ran out of time to validate challenger
-                self.logger.debug("Stop challenger intensification due to adaptive capping.")
-                self.curr_inst_idx = np.inf
-
-        self.logger.debug('Cutoff for challenger: %s' % str(cutoff))
+        if challenger:
```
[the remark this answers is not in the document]
franchuterivera: When we try sampling configuration from previous sample configurations:
https://github.com/automl/SMAC3/blob/d5b93811aa74e0513e580576ebb0fbf857e18f1a/smac/intensification/successive_halving.py#L483
challenger = self.configs_to_run.pop(0)  

We can run into an scenario where there is no challenger. So a next iteration will be triggered in the next round.

smac/intensification/successive_halving.py
```diff
@@ -273,143 +273,112 @@ def _init_sh_params(self,
...
-                              "Interrupting optimization run and returning current incumbent")
-
-        # if all configurations for the current stage have been evaluated, reset stage
-        num_chal_evaluated = (
-            len(self.success_challengers | self.fail_challengers | self.do_not_advance_challengers)
-            + self.fail_chal_offset
-        )
-        if num_chal_evaluated == self.n_configs_in_stage[self.stage] and n_insts_remaining <= 0:
+            # Make sure that there is no Budget exhausted
+            if status:
```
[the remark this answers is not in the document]
franchuterivera: Here is where I plan to use the budget exception :)

Originally it was intended for this purpose. But, maybe we should create a new status to make it more clear?

smac/intensification/abstract_racer.py
```diff
@@ -117,43 +117,75 @@ def __init__(self,
...
 
-    def eval_challenger(self,
-                        challenger: Configuration,
-                        incumbent: typing.Optional[Configuration],
-                        run_history: RunHistory,
-                        time_bound: float = float(MAXINT),
-                        log_traj: bool = True) -> typing.Tuple[Configuration, float]:
-        """
-        Runs intensification by evaluating one configuration-instance at a time
+    def eval_challenger(
```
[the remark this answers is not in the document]
franchuterivera: Yes, I will move it. I left it here because I assume we eventually will use a worker scheme for it and delete it, but let me move it to the main loop.

test/test_intensify/test_abstract_racer.py
```diff
@@ -48,67 +47,6 @@ def setUp(self):
 
         self.logger = logging.getLogger(self.__module__ + "." + self.__class__.__name__)
 
-    def test_get_next_challenger(self):
```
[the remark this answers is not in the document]
franchuterivera: Yes, so there was a code before for the abstract racer, that implemented get next challengers which is no longer valid:

Situation before:
get_next_challenger -> code available in abstract calls
eval_next_challenger-> whoever inherited from this class, needs to implement this

Now, the get_next_challenger code (which is not used) requires eval_next_challenger to work. So I removed this unused code, and requested to be a raise not implemented. In this scenario, the test for the previous code is no longer valid.

smac/intensification/abstract_racer.py
```diff
@@ -117,43 +117,75 @@ def __init__(self,
...
+        self,
+        run_info: RunInfo,
+        time_bound: float = float(MAXINT),
+    ) -> typing.Tuple[StatusType, typing.Any, float, dict]:
```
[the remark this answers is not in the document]
franchuterivera: understood, I will change it in the next commit

smac/intensification/abstract_racer.py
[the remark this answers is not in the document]
franchuterivera: I will create a new status to be more informative, BUDGETEXHAUSTED instead of None

smac/intensification/abstract_racer.py
```diff
@@ -171,16 +205,56 @@ def get_next_challenger(self,
...
+        The intensifier stage will be updated based on the results/status
+        of a configuration execution.
+        Also, a incumbent will be determined.
+
+        Parameters
+        ----------
+        challenger : Configuration
+            A configuration to challenge the incumbent. Can even be the incumbent
+            to gain more confidence on it.
+        incumbet : Configuration
```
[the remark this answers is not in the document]
franchuterivera: Thanks, will fix in next commit

smac/intensification/successive_halving.py
```diff
@@ -273,143 +273,112 @@ def _init_sh_params(self,
...
         Parameters
         ----------
         challenger : Configuration
-            promising configuration
-        incumbent : typing.Optional[Configuration]
-            best configuration so far, None in 1st run
-        run_history : smac.runhistory.runhistory.RunHistory
+            A configuration to challenge the incumbent. Can even be the incumbent
+            to gain more confidence on it.
+        incumbet : Configuration
```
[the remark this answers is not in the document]
franchuterivera: Thanks, will fix in next commit

smac/intensification/successive_halving.py
```diff
@@ -273,143 +273,112 @@ def _init_sh_params(self,
...
+        Also, a incumbent will be determined.
 
         Parameters
         ----------
         challenger : Configuration
-            promising configuration
-        incumbent : typing.Optional[Configuration]
-            best configuration so far, None in 1st run
-        run_history : smac.runhistory.runhistory.RunHistory
+            A configuration to challenge the incumbent. Can even be the incumbent
```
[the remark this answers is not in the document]
franchuterivera: Thanks, will fix in next commit

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
franchuterivera: I created a new status condition to make the code more understandable. Will be in next commit

smac/intensification/hyperband.py
```diff
@@ -116,43 +116,59 @@ def __init__(self,
...
+        Also, a incumbent will be determined.
 
         Parameters
         ----------
         challenger : Configuration
-            promising configuration
-        incumbent : typing.Optional[Configuration]
-            best configuration so far, None in 1st run
-        run_history : smac.runhistory.runhistory.RunHistory
+            A configuration to challenge the incumbent. Can even be the incumbent
```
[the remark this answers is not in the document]
franchuterivera: Thanks, will fix in next commit

franchuterivera:
For all the remaining comments, the original code used challenger=None as a mechanism to move to the next iteration.
Following your comment:
**How can this statement become False? And is there something we could do to prevent this? Or change the code so that it is clearer what's happening when it's False?**

Let me see if we can change the code to directly trigger a stage transition. If that is possible I can add a test for config None, and remove the aforementioned lines

smac/runhistory/runhistory.py
```diff
@@ -32,6 +32,10 @@ def __new__(
         return super().__new__(cls, config_id, instance_id, seed, budget)
 
 
+RunInfo = collections.namedtuple(
+    'RunInfo', ['config', 'new', 'instance', 'seed', 'cutoff', 'budget'])
```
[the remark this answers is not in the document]
franchuterivera: Understood, would push this in the next commit.

smac/intensification/abstract_racer.py
```diff
@@ -240,7 +170,7 @@ def process_results(self,
             The tracked time of a configuration execution
         time_bound : float, optional (default=2 ** 31 - 1)
             time in [sec] available to perform intensify
-        status: StatusType
+        status:  typing.Optional[StatusType],
```
mfeurer, line 173: No need for a comma here

smac/intensification/hyperband.py
```diff
@@ -222,6 +219,10 @@ def get_next_challenger(self,
             run_history=run_history,
             repeat_configs=self.sh_intensifier.repeat_configs
         )
+
+        # For debug purposes, set this flag
```
mfeurer, line 223: For debugging what exactly?

smac/intensification/abstract_racer.py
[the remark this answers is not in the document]
mfeurer: No, I think it should be like this https://github.com/automl/SMAC3/blob/d5b93811aa74e0513e580576ebb0fbf857e18f1a/smac/tae/execute_ta_run.py#L239

smac/optimizer/smbo.py
```diff
@@ -258,6 +278,70 @@ def run(self) -> Configuration:
 
         return self.incumbent
 
+    @staticmethod
+    def eval_challenger(
+        run_info: RunInfo,
+        tae_runner: ExecuteTARun,
+        time_bound: float = float(MAXINT),
+        logger: typing.Optional[Logger] = None,
```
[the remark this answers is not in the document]
franchuterivera: As discussed offline, I will move eval_challenger as a separate function inside smbo in preparation to add parallel support. 
That is the reason why logger/runner are attributes.

smac/optimizer/smbo.py
```diff
@@ -258,6 +278,70 @@ def run(self) -> Configuration:
 
         return self.incumbent
 
+    @staticmethod
+    def eval_challenger(
+        run_info: RunInfo,
+        tae_runner: ExecuteTARun,
```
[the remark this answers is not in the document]
franchuterivera: As discussed offline, I will move eval_challenger as a separate function inside smbo in preparation to add parallel support. 
That is the reason why logger/runner are attributes.

smac/intensification/successive_halving.py
```diff
@@ -338,7 +336,7 @@ def process_results(self,
         if challenger:
 
             # Make sure that there is no Budget exhausted
-            if status:
+            if not status == StatusType.BUDGETEXHAUSTED:
```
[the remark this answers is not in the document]
franchuterivera: Yes it should be !=, thanks for the catch.

smac/intensification/hyperband.py
franchuterivera: Updated the comment to be:
           # For testing purposes, this attribute highlights whether a
           # new challenger is proposed or not. Not required from a functional
           # perspective

smac/intensification/abstract_racer.py
franchuterivera: Removed the comma!

smac/intensification/successive_halving.py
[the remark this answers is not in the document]
mfeurer: Can this now be dropped as `process_results` is only called if there is a challenger?

smac/optimizer/smbo.py
```diff
@@ -13,19 +13,100 @@
...
 
+from logging import Logger
+
+
 __author__ = "Aaron Klein, Marius Lindauer, Matthias Feurer"
 __copyright__ = "Copyright 2015, ML4AAD"
 __license__ = "3-clause BSD"
 
 
+def eval_challenger(
```
mlindauer, line 39: I don't see why this function lives in `smbo.py`. Seems more like a tae function.

smac/runhistory/runhistory.py
```diff
@@ -32,6 +32,10 @@ def __new__(
         return super().__new__(cls, config_id, instance_id, seed, budget)
 
 
+RunInfo = collections.namedtuple(
```
mlindauer, line 35: seems to be repetitive compared to `RunKey`. Why do we need both?

mfeurer: The difference is that this has the configuration object itself, while the `RunKey` has only the configuration ID. The idea is to pass all relevant information for execution to the target function, so that the target function can be executed in a different worker that does not have access to the runhistory.

smac/optimizer/smbo.py
franchuterivera: Definitely, moving it out.

franchuterivera:
The overall idea of this push is to isolate the necessary information to run a config, via this new object called RunInfo.

Also, to be able to submit isolated runs of the configs with this RunInfo object (to a scheduler), we need to disentangle execution and post-processing, particularly with regards to the run history.

For that, I am adding in this push 2 changes:

-  Removing the run history from the ta runner, and centralizing the processing of finished runs to the smbo.

- Defining a RUNNING status, that indicates that a run is launched and in progress. The smbo, process finished runs and update the run history accordingly.


When this change is processed, we want (in a separate PR), to put in a proof of concept of a scheduler with a parallel random search via dask.

smac/intensification/intensification.py
```diff
@@ -171,112 +202,133 @@ def eval_challenger(self,
...
+                self.added_challenger_run = False
+                incumbent = self._process_racer_results(
+                    challenger=challenger,
+                    incumbent=incumbent,
+                    run_history=run_history,
+                    log_traj=log_traj,
+                )
+
+        elapsed_time = result.endtime - result.starttime
+        if not result.status == StatusType.BUDGETEXHAUSTED:
```
mfeurer, line 245: Could you use != here? Also, would it be possible to not make the budget exhausted statustype go into the process results function? If the budget is exhausted, nothing should be changed any more, right?

smac/intensification/intensification.py
```diff
@@ -588,6 +699,202 @@ def get_next_challenger(self,
...
+                     chooser: typing.Optional[EPMChooser],
+                     run_history: RunHistory,
+                     repeat_configs: bool = True,
+                     ) -> RunInfo:
+        """
+        Wrapper around query challengers, to get the next configuration
+        to run, as well as updating the internal state of the intensifier.
+
+        If query challenger suggest a new iteration through a None challenger,
+        this function calls itself recursively until a valid configuration is
```
mfeurer, line 714: Would it be possible to make this function non-iterative? It's already quite complicated and I have the impression that the recursiveness adds another layer of complexity on top.

franchuterivera: Yes, I am gonna push a code that will clearly align more clearly to the intensification procedure. 

Please let me know if it required further clarification.

smac/intensification/intensification.py
franchuterivera: Yes, this is here because of the legacy of the old code, but this is something that the smbo can catch and proceed accordingly. I will move this to smbo.

smac/intensification/intensification.py
```diff
@@ -132,35 +133,65 @@ def __init__(self, tae_runner: ExecuteTARun,
         self.configs_to_run = iter([])  # type: _config_to_run_type
         self.update_configs_to_run = True
 
+        # Status Tracking variables
```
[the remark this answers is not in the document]
franchuterivera: Yes, new status are:
`   class IntensifierStage(Enum):
       """Class to define different stages of intensifier"""
       RUN_FIRST_CONFIG = 0  # to replicate the old initial design
       RUN_INCUMBENT = 1  # Lines 3-7
       RUN_CHALLENGER = 2  # Lines 8-17
       RUN_BASIS = 3
   
       # helpers to determine what type of run to process
       # A challenger is assumed to be processed if the stage
       # is not from first_config or incumbent
       PROCESS_FIRST_CONFIG_RUN = 4
       PROCESS_INCUMBENT_RUN = 5
   
`

test/test_intensify/test_successive_halving.py
```diff
@@ -52,12 +53,40 @@ def setUp(self):
 
         self.logger = logging.getLogger(self.__module__ + "." + self.__class__.__name__)
 
+    def _eval_challenger(self, run_info, taf):
```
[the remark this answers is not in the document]
franchuterivera: yes, I will make this happen in the next push.

smac/utils/validate.py
```diff
@@ -37,15 +39,34 @@ def _unbound_tae_starter(
...
     -------
-    tae_results: tuple
+    tae_results: RunValue
         return from tae.start
     """
-    return tae.start(*args, **kwargs)
+    result = execute_ta_run_wrapper(tae, run_info, *args, **kwargs)
+    tae.stats.ta_runs += 1
+    tae.stats.ta_time_used += float(result.time)
+    if runhistory:
```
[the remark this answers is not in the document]
franchuterivera: It can be optional, please look at https://github.com/automl/SMAC3/blob/d5b93811aa74e0513e580576ebb0fbf857e18f1a/test/test_utils/test_validate.py#L80

I will agree however that static checking should have highlighted, that it should be typing.Optional and I will add this. 


I am not at all familiar with the validation, but looking at the unit testing, the owner expect it to be None in certain cases.

smac/runhistory/runhistory.py
```diff
@@ -247,8 +252,10 @@ def _add(self, k: RunKey, v: RunValue, status: StatusType,
         self.data[k] = v
         self.external[k] = origin
 
+        # Capped data is added above
+        # Do not register the cost until the run has completed
         if origin in (DataOrigin.INTERNAL, DataOrigin.EXTERNAL_SAME_INSTANCES) \
-                and status != StatusType.CAPPED:  # Capped data is added above
+                and (status != StatusType.CAPPED and status != StatusType.RUNNING):
```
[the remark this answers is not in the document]
franchuterivera: I will add this to the next push!

smac/optimizer/smbo.py
```diff
@@ -205,18 +220,47 @@ def run(self) -> Configuration:
...
                 try:
-                    self.incumbent, inc_perf = self.intensifier.eval_challenger(
-                        challenger=challenger,
+                    result = execute_ta_run_wrapper(
+                        run_info=run_info,
+                        tae_runner=self.tae_runner,
+                        logger=self.logger,
+                    )
+
+                    # Mimic adding this run as Running
```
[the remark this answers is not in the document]
franchuterivera: This makes a lot of sense, let me do that

smac/tae/execute_ta_run_wrapper.py
```diff
@@ -0,0 +1,157 @@
...
+            status = StatusType.CAPPED
+            # In case of a capped run, we expect the
+            # status to be:
+            # status, cost, dur, res = StatusType.CAPPED, float(MAXINT), run_info.cutoff, {}
+            # This is set by the TA callable
+    else:
+        if status == StatusType.CRASHED:
+            cost = tae_runner.cost_for_crash
+
+    if tae_runner.run_obj == "runtime":
```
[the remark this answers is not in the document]
franchuterivera: Sorry, copy paste error

smac/facade/smac_ac_facade.py
```diff
@@ -544,7 +542,8 @@ def __init__(self,
             'acquisition_func': acquisition_function_instance,
             'rng': rng,
             'restore_incumbent': restore_incumbent,
-            'random_configuration_chooser': random_configuration_chooser_instance
+            'random_configuration_chooser': random_configuration_chooser_instance,
+            'tae_runner': tae_runner_instance,
```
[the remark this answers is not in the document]
franchuterivera: The tae runner is currently an object, whose run method is what executes the config.

We need to somehow specify to the execute_ta_run_wrapper/future worker, what to execute, and the mechanism used right now is providing this object.

This object reaches smbo as an argument. To remove this from here, we have to then deprecate the ta runners in favor of the user providing a callable to execute the runs? It will still have to reach the smbo.

smac/optimizer/smbo.py
```diff
@@ -326,3 +371,60 @@ def _get_timebound_for_intensification(self, time_spent: float, update: bool) ->
...
+        run_info: RunInfo
+            Describes the run (config) from which to process the results
+        result: RunValue
+            Contains relevant information regarding the execution of a config
+        """
+
+        # Change the status from RUNNING to budget exhausted,
+        # for tracking purposes. As there is no more budget, the
+        # SMBO optimize will finish
+        if result.status == StatusType.BUDGETEXHAUSTED:
```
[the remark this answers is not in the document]
franchuterivera: thanks, i will make this more clear

smac/facade/smac_ac_facade.py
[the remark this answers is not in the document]
mfeurer: My bad, I read wrongly and didn't see in the side-by-side diff that this was a new dict. Sorry for bothering here.

smac/intensification/intensification.py
```diff
@@ -9,33 +8,89 @@
...
 class Intensifier(AbstractRacer):
-    """Races challengers against an incumbent (a.k.a. SMAC's intensification
-    procedure).
+    """Races challengers against an incumbent
+
+    SMAC's intensification procedure is denoted below.
+    This class honors bellow procedure and implements it's sequential
+    nature via stages (IntensifierStage), which are also highlighted below
+
+    Procedure 2: Intensify(Θ~new, θinc, M, R, tintensify, Π, cˆ)
```
mfeurer, line 58: The first theta should be uppercase, is that possible? Also, for variables which are subscript, could you please use a more LaTeX-Stype notation and do t_intensify?

smac/intensification/intensification.py
```diff
@@ -9,33 +8,89 @@
...
+    cˆ(θ, Π0) denotes the empirical cost of θ on the subset of instances
```
mfeurer, line 59: Should `cˆ(θ, Π0)` be `cˆ(θ, Π')`?

smac/intensification/intensification.py
```diff
@@ -9,33 +8,89 @@
...
+    Π0 ⊆ Π, based on the runs in R; maxR is a parameter
```
mfeurer, line 60: Should `Π0 ⊆ Π` be `Π' ⊆ Π`?

smac/intensification/intensification.py
```diff
@@ -9,33 +8,89 @@
...
+    ____________________________________________________________
+    1 for i := 1, . . . , length(Θ~ new) do
+    2 θnew ← Θ~ new[i]
```
mfeurer, line 63: Would it be possible to use indents here as in SMAC's procedure 2?

smac/intensification/intensification.py
```diff
@@ -137,30 +192,303 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+        if self.stage == IntensifierStage.PROCESS_FIRST_CONFIG_RUN:
+            self.stage = IntensifierStage.RUN_FIRST_CONFIG
+        elif self.stage == IntensifierStage.PROCESS_INCUMBENT_RUN:
+            self.stage = IntensifierStage.RUN_INCUMBENT
+
+        try:
+
+            # Line 2:
+            # Understand who is the challenger.
+            # Not necessarily a challenger is always a new configuration,
```
mfeurer, line 254: Could you please write 'A challenger is not necessarily a new configuration'?

smac/intensification/intensification.py
```diff
@@ -137,30 +192,303 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+            # challenger on multiple N (line 17)
+            challenger = self.query_challenger(
+                challengers=challengers,
+                chooser=chooser,
+                run_history=run_history,
+                repeat_configs=repeat_configs,
+            )
+
+            # if first ever run, then assume current challenger to be the incumbent
+            # In this case, we need a branch new challenger
```
mfeurer, line 266: What's a 'branch new challenger'?

smac/intensification/intensification.py
```diff
@@ -137,30 +192,303 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                    incumbent = challenger
+                else:
+                    inc_runs = run_history.get_runs_for_config(
+                        incumbent,
+                        only_max_observed_budget=True
+                    )
+                    if len(inc_runs) > 0:
+                        self.logger.debug(
+                            "Skipping RUN_FIRST_CONFIG stage since "
+                            "incumbent is already run before"
```
mfeurer, line 282: is -> was

smac/intensification/intensification.py
```diff
@@ -137,30 +192,303 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                        log_traj: bool = True,
+                        ) -> \
+            typing.Tuple[Configuration, float]:
+        """
+        The intensifier stage will be updated based on the results/status
+        of a configuration execution.
 
-        Implementation of Procedure 2 in Hutter et al. (2011).
+        During intensification, a the following can happen:
+        -> Challenger raced against incumbent. self.added_challenger_run
```
mfeurer, line 468: That's no longer correct, is it?

smac/intensification/intensification.py
```diff
@@ -137,30 +192,303 @@ def __init__(self, tae_runner: ExecuteTARun,
...
         ----------
         challenger : Configuration
-            promising configurations
-        incumbent : typing.Optional[Configuration]
+            a configuration that was previously executed, and whose status
+            will be used to define the next stage.
+        incumbet : typing.Optional[Configuration]
             best configuration so far, None in 1st run
-        run_history : smac.runhistory.runhistory.RunHistory
+        run_history : typing.Optional[smac.runhistory.runhistory.RunHistory]
```
mfeurer, line 484: It's not optional above, right?

smac/intensification/intensification.py
```diff
@@ -515,16 +920,96 @@ def _get_instances_to_run(self,
 
         return to_run, inc_sum_cost
 
-    def get_next_challenger(self,
-                            challengers: typing.Optional[typing.List[Configuration]],
-                            chooser: typing.Optional[EPMChooser],
-                            run_history: typing.Optional[RunHistory] = None,
-                            repeat_configs: bool = True) -> \
+    def query_challenger(self,
```
mfeurer, line 923: I feel that this function currently does more than one thing which I find really confusing. Suggestion: at the beginning of a new iteration we call the function to create new challengers from the challenger list or the chooser, this would reduce the amount of different things this function does and hopefully make it a bit easier to comprehend.

smac/intensification/intensification.py
```diff
@@ -137,30 +192,303 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                    # we might need to comply with line 17 and keep running the
+                    # same challenger
+                    self.logger.debug("No further runs for challenger possible")
+                    self._process_racer_results(
+                        challenger=challenger,
+                        incumbent=incumbent,
+                        run_history=run_history,
+                    )
+
+                    # Remove the challenger
```
mfeurer, line 326: Why is the challenger removal actually necessary here? Isn't this done in smbo.py?

smac/intensification/intensification.py
```diff
@@ -137,30 +192,303 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                        self.logger.debug(
+                            "Skipping RUN_FIRST_CONFIG stage since "
+                            "incumbent is already run before"
+                        )
+                        self.stage = IntensifierStage.RUN_INCUMBENT
+
+            # Before we proceed with the intensification loop,
+            # we have to check if it makes sense to run the challenger
+            # That is, if running lines 8-18 but there is no more time due
+            # to adaptive caping or there are no more instances to run
```
mfeurer, line 289: caping -> capping

smac/intensification/intensification.py
```diff
@@ -137,30 +192,303 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                        only_max_observed_budget=True
+                    )
+                    if len(inc_runs) > 0:
+                        self.logger.debug(
+                            "Skipping RUN_FIRST_CONFIG stage since "
+                            "incumbent is already run before"
+                        )
+                        self.stage = IntensifierStage.RUN_INCUMBENT
+
+            # Before we proceed with the intensification loop,
```
mfeurer, line 286: Is this here only for optimization? If yes, I would like to leave it out of this PR to keep the intensification code as simple as possible for now.

franchuterivera: Wait, I think the biggest part of the complexity overhead is to remove the None scenario.

What I mean by this, is that the previous code outputs None as a criteria to skip an iteration/move to a new iteration and so on. 

I thought we wanted to remove this behavior and always output a valid configuration for a runner to launch. Hence we look ahead of time if there are not gonna be challengers, instances and so on.

If that is not a requirement, then the code can be greatly simplified by removing not only this if but many other look-ahead places in the code. Please let me know.

mfeurer: > If that is not a requirement, then the code can be greatly simplified by removing not only this if but many other look-ahead places in the code. Please let me know.

In my current understanding we don't require intensify to return a valid runinfo as we return the `IntensificationBehest` to allow skipping a configuration/having the function called again. I guess this will make it a bit more inefficient, but it will make the code much more readable and simpler for now. Does this make sense?

franchuterivera: a lot, thanks, this will allow further simplification. Let me remove some code here and there and do another push.

smac/intensification/abstract_racer.py
```diff
@@ -20,6 +21,19 @@
...
+    """Class to define different request on how to process
+    the runinfo
+
+    Gives the flexibility to indicate whether no more configs
+    are available (TERMINATE in this case), no need to rerun
+    previous configurations (SKIP) or we simple should run
+    """
+    RUN = 0  # Normal run execution of a run info
+    SKIP = 1  # Skip running the run info
+    TERMINATE = 2  # No more configurations to try
```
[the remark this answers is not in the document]
franchuterivera: Moved to:

```
 class RunInfoIntent(Enum):
     """Class to define different requests on how to process
     the runinfo
 
     Gives the flexibility to indicate whether no more configs
     are available (STOP_ITERATION in this case), no need to rerun
     previous configurations (SKIP) or if the SMBO should simple
     run a generated run_info.
     """
     RUN = 0  # Normal run execution of a run info
     SKIP = 1  # Skip running the run_info
     STOP_ITERATION = 2  # No more configurations to try
```

smac/intensification/abstract_racer.py
```diff
@@ -20,6 +21,19 @@
 __license__ = "3-clause BSD"
 
 
+class IntensifierBehest(Enum):
```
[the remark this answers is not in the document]
franchuterivera: Moved to:

```
 class RunInfoIntent(Enum):
     """Class to define different requests on how to process
     the runinfo
 
     Gives the flexibility to indicate whether no more configs
     are available (STOP_ITERATION in this case), no need to rerun
     previous configurations (SKIP) or if the SMBO should simple
     run a generated run_info.
     """
     RUN = 0  # Normal run execution of a run info
     SKIP = 1  # Skip running the run_info
     STOP_ITERATION = 2  # No more configurations to try
```

smac/optimizer/smbo.py
```diff
@@ -198,14 +198,19 @@ def run(self) -> Configuration:
...
                 incumbent=self.incumbent,
                 chooser=self.epm_chooser,
                 run_history=self.runhistory,
                 repeat_configs=self.intensifier.repeat_configs,
             )
 
+            # Check if no more configs,
+            if behest == IntensifierBehest.TERMINATE:
+                self.logger.warn("No more valid configurations to try during intensification")
+                break
```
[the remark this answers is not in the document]
franchuterivera: It should be a break, because there are no more configurations to run. 

That is, in the unlikely event (but possible in unit testing due to the limited configs to try) that there is nothing more to try, this intent is triggered

smac/optimizer/smbo.py
```diff
@@ -220,43 +225,45 @@ def run(self) -> Configuration:
...
-                            instance_id=run_info.instance,
-                            seed=run_info.seed,
-                            budget=run_info.budget,
+                    if result.status != StatusType.BUDGETEXHAUSTED:
+
+                        # Add the results of the run to the run history
+                        self._incorporate_run_results(run_info, result)
+
+                        # Update the intensifier with the result of the runs
+                        # Has to be executed regardless the config to run is None,
```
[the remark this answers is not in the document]
franchuterivera: Yes, thanks it has been updated

smac/intensification/intensification.py
franchuterivera: It was supposed to be an arrow, which clearly I will remove now. I also added a more latex-like visualization of this code

smac/intensification/intensification.py
franchuterivera: thanks, yes it is missing the apostrophe

smac/intensification/intensification.py
franchuterivera: Yes, I added proper indentation now

smac/intensification/intensification.py
franchuterivera: Done!

smac/intensification/intensification.py
franchuterivera: Updated the comment, it was supposed to be brand, but moved away from this.

smac/intensification/intensification.py
franchuterivera: updated the comment accordingly

smac/intensification/intensification.py
franchuterivera: updated the comment accordingly

smac/intensification/intensification.py
franchuterivera: True, I fixed the docstrings accordingly.

smac/intensification/intensification.py
```diff
@@ -515,16 +920,96 @@ def _get_instances_to_run(self,
...
+
+        That is, not necessarily a new challenger will be produce, as following scenarios
+        are relevant (in order of priority):
+
+        -> RUN_BASIS:
+            Prioritize running against a configuration
+        -> Line 17:
+            Not enough N runs available to decide if the challenger is better/worst than
+            the incumbent, so continue challenger is honored.
+        -> Line 2/11:
```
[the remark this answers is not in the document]
franchuterivera: Yes, this comment is misleading and I have updated it

smac/intensification/intensification.py
franchuterivera: As per discussed previously, will move to use SKP in favor of more simple code. This will be removed

smac/intensification/intensification.py
franchuterivera: thanks, updated the comment

smac/intensification/abstract_racer.py
```diff
@@ -22,6 +21,19 @@
...
+    """Class to define different request on how to process
```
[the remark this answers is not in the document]
franchuterivera: thanks, updated the comment

smac/intensification/abstract_racer.py
```diff
@@ -22,6 +21,19 @@
...
+    the runinfo
+
+    Gives the flexibility to indicate whether no more configs
+    are available (TERMINATE in this case), no need to rerun
+    previous configurations (SKIP) or we simple should run
+    """
+    RUN = 0  # Normal run execution of a run info
+    SKIP = 1  # Skip running the run info
```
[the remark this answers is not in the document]
franchuterivera: thanks, updated the comment

smac/intensification/abstract_racer.py
```diff
@@ -22,6 +21,19 @@
 __license__ = "3-clause BSD"
 
 
+class IntensifierBehest(Enum):
+    """Class to define different request on how to process
+    the runinfo
+
+    Gives the flexibility to indicate whether no more configs
+    are available (TERMINATE in this case), no need to rerun
+    previous configurations (SKIP) or we simple should run
```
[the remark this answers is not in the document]
franchuterivera: thanks, updated the comment

smac/intensification/abstract_racer.py
```diff
@@ -117,70 +124,81 @@ def __init__(self,
...
+             incumbent configuration
+        chooser : smac.optimizer.epm_configuration_chooser.EPMChooser
+            optimizer that generates next configurations to use for racing
         run_history : smac.runhistory.runhistory.RunHistory
             stores all runs we ran so far
-        time_bound : float, optional (default=2 ** 31 - 1)
-            time in [sec] available to perform intensify
-        log_traj : bool
-            whether to log changes of incumbents in trajectory
+        repeat_configs : bool
```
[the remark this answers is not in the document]
franchuterivera: I need a little bit of guidance here. 

This is the same as in master https://github.com/automl/SMAC3/blob/d5b93811aa74e0513e580576ebb0fbf857e18f1a/smac/intensification/abstract_racer.py#L155

Should this be changed?

smac/intensification/abstract_racer.py
```diff
@@ -117,70 +124,81 @@ def __init__(self,
...
+            if False, an evaluated configuration will not be generated again
 
         Returns
         -------
-        typing.Tuple[Configuration, float]
-            incumbent and incumbent cost
+        run_info: RunInfo
```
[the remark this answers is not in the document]
franchuterivera: thanks for catching this, updated accordingly

smac/optimizer/smbo.py
```diff
@@ -326,3 +378,53 @@ def _get_timebound_for_intensification(self, time_spent: float, update: bool) ->
                           "intensification: %.4f (%.2f)" %
                           (total_time, time_spent, (1 - frac_intensify), time_left, frac_intensify))
         return time_left
+
+    def _incorporate_run_results(self, run_info: RunInfo, result: RunValue) -> None:
+        """
+        The SMBO submit a config run request via a RunInfo object.
+        When that config run is completed, a RunValue, which contains
+        all the relevant information for a job, is returned.
```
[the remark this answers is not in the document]
franchuterivera: updated the comment to be more verbose

smac/runhistory/runhistory.py
```diff
@@ -173,6 +177,7 @@ def add(
         endtime: float = 0.0,
         additional_info: typing.Optional[typing.Dict] = None,
         origin: DataOrigin = DataOrigin.INTERNAL,
+        force_update: bool = False,
```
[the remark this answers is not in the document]
franchuterivera: Added this new bool argument

smac/intensification/intensification.py
franchuterivera: I tried simplifying the function so that now it is needed when you desired a new challenger, not in all possible situations as it was before. Please let me know if you consider it should be simplified more.

Conceptually I think this function was meant to be used to get the next challenger for racing (it is taken from the master branch), but it could be called at any time by anyone, so ti was a bit flexible, but un-readable.

I removed this flexibility and updated the doc accordingly.

smac/intensification/intensification.py
```diff
@@ -9,33 +8,89 @@
...
+    PROCESS_INCUMBENT_RUN = 5
+
 
 class Intensifier(AbstractRacer):
-    """Races challengers against an incumbent (a.k.a. SMAC's intensification
-    procedure).
+    """Races challengers against an incumbent
+
+    SMAC's intensification procedure is denoted below.
+    This class honors bellow procedure and implements it's sequential
```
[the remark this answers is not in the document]
franchuterivera: updated the comment accordingly

smac/intensification/intensification.py
franchuterivera: yes thanks, updated the comment

franchuterivera:
A did a new push, but likely depending on the outcome of the above pending comments, a new push might be needed.

smac/intensification/abstract_racer.py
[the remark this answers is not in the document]
KEggensperger: I just had a brief look at it, this will require more work (and should be done in another PR). For now this is fine and there is nothing to do here.

smac/optimizer/smbo.py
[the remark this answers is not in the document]
mfeurer: This still puzzles me. Having a break here completely ends the optimization and does not move to a new iteration. However, we did not have such a feature before. Maybe we need to discuss this offline.

test/test_intensify/test_intensify.py
```diff
@@ -1090,16 +1096,21 @@ def target(x):
...
         # In this particular unit testing, no further new configurations
         # nor a chooser is provided, so we run into a value error
-        with self.assertRaisesRegex(ValueError, "No configurations/chooser provided"):
-            behest, run_info = intensifier.get_next_run(
-                challengers=[self.config1],
-                incumbent=inc,
-                run_history=self.rh,
-                chooser=None
-            )
+        intent, run_info = intensifier.get_next_run(
```
mfeurer, line 1099: The comment is no longer aligned with the code

smac/intensification/intensification.py
```diff
@@ -237,192 +558,283 @@ def eval_challenger(self,
...
         inc_perf = run_history.get_cost(incumbent)
 
         return incumbent, inc_perf
 
-    def _add_inc_run(self,
-                     incumbent: Configuration,
-                     run_history: RunHistory,
-                     log_traj: bool = True) -> None:
-        """Add new run for incumbent
+    def _get_next_inc_config(self,
```
mfeurer, line 583: Could you rename this to `_get_next_inc_run` as it does not return a configuration?

smac/intensification/intensification.py
```diff
@@ -137,94 +190,362 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                # This point marks the transitions from lines 3-7
+                # to 8-18.
+
+                self.logger.debug(
+                    "No further instance-seed pairs for incumbent available."
+                )
+
+                self.stage = IntensifierStage.RUN_CHALLENGER
+
+                self._process_inc_run(
```
mfeurer, line 321: Shouldn't this have happened in the previous iteration in `process_run`?

smac/intensification/intensification.py
```diff
@@ -137,94 +190,362 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                challengers=challengers,
+                chooser=chooser,
+            )
+
+        # No new challengers are available for this iteration,
+        # Move to the next iteration. This can only happen
+        # when all configurations for this iteration are exhausted
+        # and have been run in all proposed instance/pairs.
+        if challenger is None:
+            return RunInfoIntent.SKIP, RunInfo(
```
mfeurer, line 350: Don't you also have to call `_next_iteration`?

smac/intensification/intensification.py
```diff
@@ -137,94 +190,362 @@ def __init__(self, tae_runner: ExecuteTARun,
...
+                    self.logger.debug(
+                        "Stop challenger itensification due "
+                        "to adaptive capping."
+                    )
+
+                # Nevertheless, if there are no more instances to run,
+                # we might need to comply with line 17 and keep running the
+                # same challenger
+                self.logger.debug("No further runs for challenger possible")
+                self._process_racer_results(
```
mfeurer, line 413: I have the same questions here as with the function process incumbent runs. Is this really necessary and shouldn't such handling be done in the process_results?

smac/intensification/intensification.py
franchuterivera: As in the previous code, this is something that get_next_challenger does.

I think it makes sense to keep it there because it highlights when the iteration is done?

If you think otherwise, I can move it here.

smac/intensification/intensification.py
franchuterivera: I think the justification for this to be like that (as in the previous code) is that if there is no new challenger, you don't call process results, so this piece of code makes sure that on the next iteration we move to run the incumbent or we keep running the challenger, based on the run_history.

smac/intensification/intensification.py
mfeurer: Okay, no I get it - could you please add a not to the docstring of `get_next_challenger` that it can trigger a new iteration?

smac/intensification/intensification.py
franchuterivera: I completely agree and self._process_inc_run can be removed in favor of simplicity. 

It was only here to print the incumbent performance for debugging purposes.

smac/intensification/intensification.py
mfeurer: Okay, got it now. Could you please add more info that line 17 actually doubles the number of instances to be compared here?

smac/intensification/intensification.py
franchuterivera: understood, I will rename it.

smac/intensification/intensification.py
franchuterivera: thanks, I updated the full comment to be:
                   # Nevertheless, if there are no more instances to run,
                   # we might need to comply with line 17 and keep running the
                   # same challenger. In this case, if there is not enough information
                   # to decide if the challenger is better/worst than the incumbent,
                   # line 17 doubles the number of instances to run.

smac/intensification/hyperband.py
```diff
@@ -116,43 +112,52 @@ def __init__(self,
...
         Parameters
         ----------
         challenger : Configuration
-            promising configuration
-        incumbent : typing.Optional[Configuration]
-            best configuration so far, None in 1st run
-        run_history : smac.runhistory.runhistory.RunHistory
+            A configuration that was previously executed, and whose status
+            will be used to define the next stage.
+        incumbet : Configuration
```
mfeurer: There's a typo here, should be incumbe**n**t

mfeurer: Sorry, looked at the wrong commits.

smac/intensification/intensification.py
franchuterivera: thanks, I will update the comment to be:
This method triggers a call to _next_iteration if there are no more configurations
           to run, for the current intensification loop. This marks the transition to Line 2,
           where a new configuration to intensify will be drawn from epm/initial challengers.

test/test_intensify/test_intensify.py
franchuterivera: updated the comment to be:

 # In the upcoming get next run, the stage is RUN_CHALLENGER
 --        # so the intensifier tries to run config1. Nevertheless, 
 --        # there are no further instances for this configuration available. 
           # In this scenario, the intensifier produces a SKIP intent as an indication
 --        # that a new iteration must be initiated, and for code simplicity, 
           # relies on a new call to get_next_run to yield more configurations

smac/intensification/hyperband.py
franchuterivera: fixed!

franchuterivera:
I removed the stop_iteration in favor of simplicity and moved to the approach available in the master branch.

smac/intensification/intensification.py
```diff
@@ -171,60 +502,45 @@ def eval_challenger(self,
...
+
+                self._ta_time += result.time
+                incumbent = self._process_racer_results(
+                    challenger=challenger,
+                    incumbent=incumbent,
+                    run_history=run_history,
+                    log_traj=log_traj,
+                )
 
+        elapsed_time = result.endtime - result.starttime
```
KEggensperger, line 543: I am wondering two things here: 
1) What happened to self.elapsed time? It is still an attribute of this class but is is never updated, however, I don't see where we use it?
2) What is the difference between result.endtime - result.starttime and result.time

franchuterivera: Thanks, this clearly has to be cleaned. Let me do a push for this.

In the old code, there are 2 mechanisms to see if there is time left for this iteration:

- Trusting the evaluator implemented by the user (use_ta_time_bound is set to true), so the time this iteration has gone is constructed by adding the duration stated in the runhistory time attribute
- Measuring the duration via time.time() at the start (result.starttime) and at the end (result.endtime ) regardless of what the ta function does, and use this to measure the accumulative duration of this iteration.

So to answer your point 2, the difference is whether you trust the ta run method implemented by the user of smac, or you just track it yourself using the time module.


Elapsed time, in the old code measure the time it took to evaluate the challenger configuration

franchuterivera:
I also notice the following opportunities to improve unit testing:

1- Time bound/elapsed time tracking -- I completely agree a unit testing is needed for this
2- Interaction between smbo and intensifier. I think a test that optimizes a dummy function like rosenbrock should be added to smbo to make sure the handshaking between smbo/intensifier is honored. Currently, smbo tests are not optimization driven.
3- The intensifier does not have a stage transition testing. It tests really good the individual stages, like racing a challenger or gaining more confidence on the incumbent, but not making sure a transition should happen on known scenarios.
4- Adding more examples for successive halving, like for instance optimizing Rosebrock on successive halving and hyperband was really useful for me to clean up the code.

PR #671, opened 2020-08-09
Run info intensification dask

franchuterivera:
This push is an increment to #663 

Relevant to check:

- A new base class for runners
- An implementation of that class for serial runs, called SerialRunners
- An implementation of that class for parallel runs, called DaskParallelRunner

Initially, I started with a bridge design pattern, but the thing is that we would have to recode every single serial runner to a parallel. I believe it is better to just have DaskParallelRunner wrapper a single runner base on the number of n_workers defined in the scenario.

This approach tries to make smbo agnostic of the strategy of parallelization, and even now we can easily plug in ray/dask/pyro as desired.


franchuterivera:
Matthias, before I continue coding (I am missing the testing code for this new classes), we can sync up to see if the direction make sense.

PR #672, opened 2020-08-12
Only validate limit in case of Sobol design.

### LINKED ISSUE #664: SMAC4HPO fails on Sobol initial design limitation even when initial design is replaced.
#### Description
`SMAC4HPO` only supports dimensionality up to 40, due to Sobol initial design, even if the initial design is replaced with another (eg. `RandomConfigurations`). If the initial design is replaced with something else, this limitation should no longer apply.

#### Steps/Code to Reproduce
https://gist.github.com/8uurg/42cc7e1c9999d6c38185ac0b5c910ee7

#### Expected Results
The optimizer runs.

#### Actual Results
```
<snip>
  File "/<snip>/lib/python3.8/site-packages/smac/facade/smac_hpo_facade.py", line 44, in __init__
    raise ValueError(
ValueError: The default initial design "Sobol sequence" can only handle up to 40 dimensions. Please use a different initial design, such as "the Latin Hypercube design".
```

#### Versions
0.12.2

8uurg:
#### Reference Issues/PRs

Fixes #664. SMAC4HPO fails on Sobol initial design limitation even when initial design is replaced.

#### What does this implement/fix? Explain your changes.
An assertion in the HPO facade is overly strict, and limits the number of hyperparameters, even when the initial design has been replaced with a design other than `SobolDesign`.
This PR makes it so that a ValueError is only raised when the design used is SobolDesign.


mfeurer:
Looks good to me, thanks for contributing!

PR #677, opened 2020-09-06
Parallel sh

franchuterivera:
examples/parallel_sh_mlp.py is a good start point for the changes. Important notes:

- Updated successive halving code to work in parallel. That is, even when using multiple cores, the original code can work, but it will issue a WAIT when it needs for running conflgs to complete. That is, when a SH stage is over, the new code waits for all configs of that stage to wait before moving to a new stage.

- Created a parallel successive halving code, which basically spawns successive halvers on a need basis.


smac/intensification/successive_halving.py
```diff
@@ -820,3 +872,70 @@ def _top_k(self,
...
+
+        # Check which of the current configs is running
+        my_configs = [c for c, i, s in self.run_tracker]
+        running_configs = set()
+        for k, v in run_history.data.items():
+            if v.status == StatusType.RUNNING:
+                if run_history.ids_config[k.config_id] in my_configs:
+                    running_configs.add(k.config_id)
+
+        # if all configurations for the current stage have been evaluated, reset stage
```
mfeurer, line 933: There's no reset happening here, right? Could you please drop this from the comment?

test/test_intensify/test_successive_halving.py
```diff
@@ -971,3 +972,148 @@ def test_incumbent_selection_designs(self):
...
+
+        # Now, when the intensifier is requested, to provide a new run,
+        # it should provide a new instance, not the same one
+        new_intent, new_run_info = intensifier.get_next_run(
+            challengers=[self.config2, self.config3, self.config4],
+            chooser=None,
+            run_history=self.rh,
+            incumbent=None,
+        )
+        self.assertNotEqual(new_run_info, run_info)
```
mfeurer, line 1037: If it should be a new instance, could you please check for exactly that?

test/test_intensify/test_successive_halving.py
```diff
@@ -971,3 +972,148 @@ def test_incumbent_selection_designs(self):
...
+        taf = ExecuteTAFuncDict(ta=target, stats=self.stats, run_obj='quality')
+
+        taf.runhistory = self.rh
+        # select best on any budget
+        intensifier = SuccessiveHalving(
+            stats=self.stats, traj_logger=None,
+            rng=np.random.RandomState(12345), run_obj_time=False,
+            instances=list(range(20)), initial_budget=3, max_budget=10, eta=2)
+
+        # Get and process the first instance pair of config 1
```
mfeurer, line 991: Could you please add a check for an empty run history?

test/test_intensify/test_successive_halving.py
```diff
@@ -971,3 +972,148 @@ def test_incumbent_selection_designs(self):
...
+            challengers=[self.config2, self.config3, self.config4],
+            chooser=None,
+            run_history=self.rh,
+            incumbent=None,
+        )
+        self.assertNotIn(newer_run_info, [new_run_info, run_info])
+
+    def test_launched_all_configs_for_current_stage(self):
+        """
+        Makes sure we return the proper number of running configs
```
mfeurer, line 1061: Is this a copy-paste error?

test/test_intensify/test_successive_halving.py
```diff
@@ -971,3 +972,148 @@ def test_incumbent_selection_designs(self):
...
+            run_tracker.append((run_info.config, run_info.instance, run_info.seed))
+            self.rh.add(config=run_info.config,
+                        instance_id=run_info.instance,
+                        seed=run_info.seed,
+                        budget=run_info.budget,
+                        cost=10,
+                        time=1,
+                        status=StatusType.RUNNING,
+                        additional_info=None)
+        print(intensifier._launched_all_configs_for_current_stage(self.rh))
```
mfeurer, line 1106: Please drop this print statement.

smac/intensification/successive_halving.py
```diff
@@ -195,6 +200,11 @@ def __init__(self,
         self.fail_challengers = set()  # type: typing.Set[Configuration]
         self.fail_chal_offset = 0
 
+        # Track which configs were launched. This will allow to have an extra check to make sure
+        # that a successive halver deals only with the configs it launched, but also allows querying the
+        # status of the configs via the RH -- knowing what to query
```
mfeurer, line 205: could you please add a brief explanation why this would have to query the run history in the first place and also write out RH?

test/test_intensify/test_parallel_successive_halving.py
```diff
@@ -0,0 +1,391 @@
...
+        as a nice side effect runhistory/stats are properly filled
+        """
+        challengers = [self.config1, self.config2, self.config3, self.config4]
+        incumbent = None
+        for i in range(100):
+            try:
+                intent, run_info = sh.get_next_run(
+                    challengers=challengers,
+                    incumbent=None, chooser=None, run_history=rh)
+            except ValueError as e:
```
mfeurer, line 300: Running out of configs shouldn't raise a ValueError as this is not about calling something with an illegal value, isn't it? Is there a more appropriate exception we could raise?

test/test_intensify/test_parallel_successive_halving.py
```diff
@@ -0,0 +1,391 @@
...
+
+        # This makes sure there is a single incumbent in PSH
+        self.assertEqual(inc_perf, inc_perf_psh)
+
+        # We don't want to loose any configuration, and particularly
+        # we want to make sure the values of SH to PSH match
+        self.assertEqual(len(self.rh.data), len(rh.data))
+
+        configs_sh_rh = {}
+        for k, v in rh.data.items():
```
mfeurer, line 373: Shouldn't one of these two for loops go over `self.rh` instead of `rh`?

examples/mlp_from_cfg_func.py
```diff
@@ -0,0 +1,53 @@
...
+from sklearn.neural_network import MLPClassifier
+
+
+# A common function to be optimized by a Real valued Intensifier
+digits = load_digits()
+
+
+def mlp_from_cfg(cfg, seed, instance, budget, **kwargs):
+    """
+        Creates a MLP classifier from sklearn and fits the given data on it.
```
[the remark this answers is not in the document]
franchuterivera: Done, added:
```
+# Target Algorithm
+# The signature of the function determines what arguments are passed to it
+# i.e., budget is passed to the target algorithm if it is present in the signature

```

examples/parallel_sh_mlp.py
```diff
@@ -0,0 +1,98 @@
+"""
+================================================
+Optimizing an MLP with Parallel SuccesiveHalving
+================================================
+An example for the usage of a model-free SuccessiveHalving intensifier in SMAC,
+with support for parallel execution. The configurations are randomly sampled.
```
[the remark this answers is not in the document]
franchuterivera: Done:
```
-with support for parallel execution. The configurations are randomly sampled.
+for parallel execution. The configurations are randomly sampled.

```

examples/parallel_sh_mlp.py
```diff
@@ -0,0 +1,98 @@
...
+from smac.configspace import ConfigurationSpace
+from smac.facade.roar_facade import ROAR
+from smac.scenario.scenario import Scenario
+from smac.intensification.parallel_successive_halving import ParallelSuccessiveHalving
+from smac.initial_design.random_configuration_design import RandomConfigurations
+
+# --------------------------------------------------------------
+# We need to provide a pickable function and use __main__
+# to be compliant with multiprocessing API
+# Below is a WA to have a packaged function called rosenbrock_2d
```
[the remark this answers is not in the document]
franchuterivera: Fixed this for fmix rosenbrock and the new code:
```
-# Below is a WA to have a packaged function called rosenbrock_2d
+# Below is a work around to have a packaged function called
+# mlp_from_cfg_func

```

examples/parallel_sh_mlp.py
```diff
@@ -0,0 +1,98 @@
+"""
+================================================
+Optimizing an MLP with Parallel SuccesiveHalving
+================================================
+An example for the usage of a model-free SuccessiveHalving intensifier in SMAC,
+with support for parallel execution. The configurations are randomly sampled.
+
+This examples uses a real-valued SuccessiveHalving through epochs.
+
+4 workers are allocated for this run. As soon as the any worker is iddle,
```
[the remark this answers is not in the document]
franchuterivera: Sorry, while debug I changed this. Fixed

examples/parallel_sh_mlp.py
```diff
@@ -0,0 +1,98 @@
...
+                         # Then you should handle runtime and memory yourself in the TA
+                         "cutoff": 20,  # runtime limit for target algorithm
+                         "memory_limit": 3072,  # adapt this to reasonable value for your hardware
+                         })
+
+    # Intensification parameters
+    # Intensifier will allocate from 5 to a maximum of 50 epochs to each configuration
+    # max_active_SH controls how many Succesive Halving childs are created to prevent idle
+    # workers. If not passed via intensifier_kwargs it is set to the max number of workers
+    intensifier_kwargs = {'initial_budget': 5, 'max_budget': 25, 'eta': 3, 'max_active_SH': 8,
```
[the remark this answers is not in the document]
franchuterivera: same here

smac/intensification/parallel_successive_halving.py
```diff
@@ -0,0 +1,337 @@
...
+                        ) -> \
+            typing.Tuple[Configuration, float]:
+        """
+        The intensifier stage will be updated based on the results/status
+        of a configuration execution.
+
+        To do so, this procedures redirects the result argument, to the
+        respective SuccessiveHalving object that generated the original config.
+
+        Also, a incumbent will be determined. This determination is done
```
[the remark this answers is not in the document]
franchuterivera: fixed this:
```
-        Also, a incumbent will be determined. This determination is done
+        Also, an incumbent will be determined. This determination is done

```

smac/intensification/parallel_successive_halving.py
```diff
@@ -0,0 +1,337 @@
...
+            time in [sec] available to perform intensify
+        result: RunValue
+            Contain the result (status and other methadata) of exercising
+            a challenger/incumbent.
+        log_traj: bool
+            Whether to log changes of incumbents in trajectory
+
+        Returns
+        -------
+        incumbent: Configuration()
```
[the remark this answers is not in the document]
franchuterivera: Fixed this for both successive halving and the new code:
```
-        incumbent: Configuration()
+        incumbent: Configuration

```

smac/intensification/parallel_successive_halving.py
```diff
@@ -0,0 +1,337 @@
...
+        Returns
+        -------
+            List:
+                The order in which to query for new jobs
+
+        """
+        expected = self.last_active_instance + 1
+        if strategy == 'ordinal' or expected == len(self.sh_instances):
+            # This is hpbanster strategy, just pick on an ordered fashion
+            return list(range(len(self.sh_instances)))
```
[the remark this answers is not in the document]
franchuterivera: Added a new strategy called more_advanced_iteration that prioritized more advanced runs

smac/intensification/parallel_successive_halving.py
```diff
@@ -0,0 +1,337 @@
...
+        elif strategy == 'serial':
+            # All SH instances are copies launching the same type of
```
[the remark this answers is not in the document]
franchuterivera: same here

smac/intensification/parallel_successive_halving.py
```diff
@@ -0,0 +1,337 @@
...
+    min_chall: int
+        minimal number of challengers to be considered (even if time_bound is exhausted earlier). This class will
+        raise an exception if a value larger than 1 is passed.
+    incumbent_selection: str
+        How to select incumbent in successive halving. Only active for real-valued budgets.
+        Can be set to: [highest_executed_budget, highest_budget, any_budget]
+        * highest_executed_budget - incumbent is the best in the highest budget run so far (default)
+        * highest_budget - incumbent is selected only based on the highest budget
+        * any_budget - incumbent is the best on any budget i.e., best performance regardless of budget
+    max_active_SH: int
```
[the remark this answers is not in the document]
franchuterivera: Added the runner object to dynamically query the number of workers

smac/intensification/successive_halving.py
```diff
@@ -410,18 +419,41 @@ def get_next_run(self,
         if not hasattr(self, 'stage'):
             self._update_stage(run_history=run_history)
 
+        # In the case of multiprocessing, we have runs in Running stage, which have not
+        # been processed via process_results(). get_next_run() is called agnostically by
+        # smbo. To prevent launching more configs, that the ones needed, we query if
```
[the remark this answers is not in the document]
franchuterivera: fixed this!

test/test_intensify/test_successive_halving.py
franchuterivera: Improved check to check for same config and a new instance

test/test_intensify/test_successive_halving.py
franchuterivera: I added a new check, but maybe I misunderstood this. Please let me know if the new check is what you were looking for

test/test_intensify/test_successive_halving.py
franchuterivera: Yes, I fixed this.

test/test_intensify/test_successive_halving.py
franchuterivera: Removed

smac/intensification/successive_halving.py
franchuterivera: Improved the comment for this

test/test_intensify/test_parallel_successive_halving.py
franchuterivera: thanks very much for catching this. The check was useless because of this. I fixed it.

test/test_intensify/test_parallel_successive_halving.py
franchuterivera: This is a testing only problem, because we really do not provide an EPM:

https://github.com/automl/SMAC3/blob/9890e0cdc4ce898de25fa00d794cbe0af83a09c2/smac/intensification/abstract_racer.py#L222

It makes sense as a valueerror because there needs to be epm or enough configs from where to sample.

Do you think this should be changed? In real execution,smac creates the EPM so this is unlikely to happen

smac/intensification/successive_halving.py
franchuterivera: Sorry, copy paste issue that is now fixed

smac/intensification/parallel_successive_halving.py
```diff
@@ -333,5 +342,29 @@ def _sh_scheduling(self, strategy: str = 'serial') -> typing.List[int]:
...
             # scheduling
             return list(range(expected, len(self.sh_instances))) + list(range(expected))
+        elif strategy == 'more_advanced_iteration':
+            # We want to prioritize runs that are close to finishing an iteration.
+            # In the context of successive halving, an iteration have stages that are
+            # composed of # of configs and each configs has # of instance-seed pairs
+            preference = []
+            for i, sh in self.sh_instances.items():
+                # Each row of this matrix is id, stage, configs+instances for stage
+                # We use sh.run_tracker as a cheap way to now how advanced the run is
```
mfeurer, line 352: now how -> know how

smac/intensification/parallel_successive_halving.py
```diff
@@ -333,5 +342,29 @@ def _sh_scheduling(self, strategy: str = 'serial') -> typing.List[int]:
             # in above example, if last scheduler was 5, it is like ordinal
             # scheduling
             return list(range(expected, len(self.sh_instances))) + list(range(expected))
+        elif strategy == 'more_advanced_iteration':
+            # We want to prioritize runs that are close to finishing an iteration.
+            # In the context of successive halving, an iteration have stages that are
```
mfeurer, line 347: have stages -> has stages

smac/intensification/parallel_successive_halving.py
```diff
@@ -67,15 +69,16 @@ class ParallelSuccessiveHalving(AbstractRacer):
         * highest_executed_budget - incumbent is the best in the highest budget run so far (default)
         * highest_budget - incumbent is selected only based on the highest budget
         * any_budget - incumbent is the best on any budget i.e., best performance regardless of budget
-    max_active_SH: int
-        Number of max active successive halvers to support
+    runner: DaskParallelRunner
```
mfeurer, line 72: I'm still wondering if limiting the number of SH instances is really necessary. Is this to prevent the whole thing running havoc or does this provide functionality which I don't see right now?

test/test_intensify/test_parallel_successive_halving.py
mfeurer: Ah, okay, that makes sense then.

test/test_intensify/test_parallel_successive_halving.py
```diff
@@ -285,6 +289,65 @@ def test_sh_scheduling(self):
...
+        )
+
+        # Add a new run that's empty
+        self.PSH.sh_instances[4] = add_sh_mock(stage=0, config_inst_pairs=0)
+        self.assertCountEqual(
+            list(self.PSH._sh_scheduling(strategy='more_advanced_iteration')),
+            [1, 3, 2, 0, 4]
+        )
+
+        # Lastly make 0 stage 4 but with not as many instances as 1
```
mfeurer, line 344: I believe there's something wrong here - the comment says that we change sh_instance 0, but the code changes sh_instance 4, yet the assert checks that sh_instance 0 was changed.

smac/intensification/parallel_successive_halving.py
franchuterivera: Yeah exactly, to prevent it to run havoc. I can think of certain troublesome scenarios... One example is, The tae takes 20 seconds to complete a job. So each time we query for sh instances, they all will be busy. So we will keep on creating new SH instances.

I believe the goal is to prevent having idle workers, then having more instances than number of workers doesn't make sense? Again this number is dynamically updated to the number of workers. But it is a mechanism to ensure we don't create new SH without a real purpose.

test/test_intensify/test_parallel_successive_halving.py
franchuterivera: Sorry, I used assert count equal but according to the documentation:
```
a and b have the same elements in the same number, regardless of their order.
```

So I am gonna move to assert equal

smac/intensification/parallel_successive_halving.py
franchuterivera: thanks, fixed

smac/intensification/parallel_successive_halving.py
franchuterivera: thanks, fixed

smac/intensification/parallel_successive_halving.py
mfeurer: Okay, got it. What do you think of passing the number of workers as part of `get_next_algorithm`? Then it's clearer that this function depends on the number of workers, the PSH doesn't need to know about the dask worker, and the facade doesn't need to know either.

mfeurer: And other intensifiers which don't support parallelism can raise an exception if there is more than one worker.

franchuterivera: Thanks, this makes a lot of sense. I have coded and pushed this change

smac/intensification/parallel_successive_halving.py
```diff
@@ -0,0 +1,382 @@
...
+
+        """
+
+        # This function might be called when no SuccessiveHalving instances
+        # exist (first iteration), so we return an empty list in that case
+        if len(self.sh_instances) == 0:
+            return []
+
+        expected = self.last_active_instance + 1
+        if strategy == 'ordinal' or expected == len(self.sh_instances):
```
mfeurer, line 346: I'm actually wondering if we still need multiple strategies here? What would be the reason to not use 'more_advanced_iteration'?

franchuterivera: thanks, I have removed this.

mfeurer (approved):

PR #684, opened 2020-09-16
Private classes for HB and SH

franchuterivera:
Changes in class names:

- ParallelSuccessiveHalving -> SuccessiveHalving
- SuccessiveHalving -> _SuccessiveHalving
- ParallelHyperband -> _Hyperband
- Hyperband -> _Hyperband

Please do notice that I updated the documentation to reflect the new user interfaces, that Is the new successive halving details the method as the final user will see this. Nevertheless, there are clarifications noted on what the private class does


mfeurer (approved):

mfeurer:
This looks good, but it appears that one of the unit tests fails after rearranging files.

PR #686, opened 2020-09-17
Parallel bookkeeping

mfeurer:
This PR splits the stats attribute `ta_runs` into `ta_runs_submitted` and `ta_runs_finished` to correctly keep track of both.

### DIFF
--- a/smac/stats/stats.py
+++ b/smac/stats/stats.py
@@ -23,7 +23,8 @@ class Stats(object):
 
     Attributes
     ----------
-    ta_runs
+    submitted_ta_runs
+    finished_ta_runs
     wallclock_time_used
     ta_time_used
     inc_changed
@@ -40,7 +41,8 @@ class Stats(object):
         """
         self.__scenario = scenario
 
-        self.ta_runs = 0
+        self.submitted_ta_runs = 0
+        self.finished_ta_runs = 0
         self.n_configs = 0
         self.wallclock_time_used = 0.0
         self.ta_time_used = 0.0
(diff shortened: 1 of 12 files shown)


KEggensperger (approved):

smac/stats/stats.py
```diff
@@ -137,7 +139,7 @@ def get_remaining_ta_runs(self) -> int:
         """Subtract the target algorithm runs in the scenario with the used ta
         runs"""
         if self.__scenario:
-            return self.__scenario.ta_run_limit - self.ta_runs  # type: ignore[attr-defined] # noqa F821
+            return self.__scenario.ta_run_limit - self.submitted_ta_runs  # type: ignore[attr-defined] # noqa F821
```
mlindauer, line 142: Is that working? I don't know the new code base fully, but the old version would throw an exception as soon as the "ta_runs" budget would be empty and we would terminate SMAC. Now, we would need to wait that the last running run terminates before we can terminate SMAC.

mfeurer: Yes, that's accounted for [here](https://github.com/automl/SMAC3/blob/e3ea18ecb4d3eb2e3e1352b0b38e4c580f0d9e8e/smac/optimizer/smbo.py#L284):

```python
if self.stats.is_budget_exhausted():
    self.logger.warn("Exhausted configuration budget")

    # The budget can be exhausted  for 2 reasons: number of ta runs or
    # time. If the number of ta runs is reached, but there is still budget,
    # wait for the runs to finish
    while self.tae_runner.pending_runs():

        self.tae_runner.wait()

        for run_info, result in self.tae_runner.get_finished_runs():
            # Add the results of the run to the run history
            # Additionally check for new incumbent
            self._incorporate_run_results(run_info, result, time_left)

    # Break from the intensification loop,
    # as there are no more resources
    break
```

mlindauer (approved):

PR #687, opened 2020-09-21
Additional info logs in crashes 

franchuterivera:
Add a check to make sure we log in crashes in additional info

### DIFF
--- a/smac/tae/base.py
+++ b/smac/tae/base.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 import math
 import time
+import traceback
 import typing
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


smac/tae/base.py
```diff
@@ -227,11 +227,11 @@ def run_wrapper(
                 budget=run_info.budget,
                 instance_specific=run_info.instance_specific
             )
-        except Exception:
+        except Exception as e:
             status = StatusType.CRASHED
             cost = self.cost_for_crash
             runtime = time.time() - start
-            additional_info = {}
+            additional_info = {'exception_msg': str(e)}
```
mfeurer, line 234: What do you think about adding the actual traceback as [here](https://github.com/automl/auto-sklearn/blob/70d1a301fc0f5e7524c4d37ce1a2ff6ce45d8d73/autosklearn/evaluation/__init__.py#L35) to increase the amount of information for debugging?

franchuterivera: Thank you very much, this makes total sense to make debugging easier. I added these changes.

PR #694, opened 2020-10-07
Add better error messages

mfeurer:
* in case of first run crashed
* in case of the pynisher crashing


franchuterivera:
Hi Matthias, the error you are seeing:

```
  AssertionError: 'in run_wrapper' not found in 'Traceback (most recent
```

I was making sure that the error that was raised come directly from the run_wrapper function. Because of the new try block, that is no longer the case. So this string search was to make sure we didn't capture any other expression. Really there is a much better way which is to enhance the test to search for an AttributeError for pickable objects. The string changes between python versions, so AttributeError and pickle are 2 good suggestions

franchuterivera (approved):

PR #698, opened 2020-10-09
Logging enhancements and running configs fix

franchuterivera:
**Problem statement:**
Jobs for a configuration called config are sent to dask and tagged with config_id=1.  It status is then set to RUNNING. It properly completes, but when added back to runhistory through the following question:

```
self.runhistory.config_ids.get(config)
```
Such config no longer is EXACTLY the same, as a worker recreated the object so the ID is different and the above question returns NONE. So that is, when adding the config back to the runhistory, it is added as a whole new configuration.

**My proposed solution:**
Add a better handshaking mechanism that relies on config_id. That is, when runhistory adds the config with config_id=1, we keep adding/tracking/updated based on the config_id, not so much as the object that can change.

BTW, the config is the same, it is just that python re-creates the object with a new id in memory.

### DIFF
--- a/smac/optimizer/smbo.py
+++ b/smac/optimizer/smbo.py
@@ -240,7 +240,7 @@ class SMBO(object):
                         self.runhistory.config_ids.get(run_info.config),
                     )
                 )
-                self.runhistory.add(
+                config_id = self.runhistory.add(
                     config=run_info.config,
                     cost=float(MAXINT),
                     time=0.0,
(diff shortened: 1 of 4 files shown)


mfeurer:
Thanks a lot for checking this.

However, I don't understand why this doesn't already work correctly right now. The configuration object implements [`__hash__`](https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/configuration_space.pyx#L1525), [`__eq__`](https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/configuration_space.pyx#L1510) and [`__ne__`](https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/configuration_space.pyx#L1519) and none of them depends on the config_id. From my understanding this should have made everything work already.

franchuterivera:
We fixed this problem through Cconfigspace: https://github.com/automl/ConfigSpace/pull/167

We will close this PR as it is no longer needed.

PR #709, opened 2020-11-19
Moved to internal registration of instances per config

franchuterivera:
Problem Statement:
When running in parallel, the current development code has problem tracking the number of instances per configuration. That is, there is no real per-config tracking of what instances have been launched and which one finished -- per configuration.

This causes an error of the form:
```
ValueError: Cannot compare configs that were run on different instances-seeds-budgets: [InstSeedBudgetKey(instance='75193_10MIN_logloss', seed=0, budget=0.0), InstSeedBudgetKey(instance='189909_10MIN_logloss', seed=0, budget=0.0)] vs [InstSeedBudgetKey(instance='189909_10MIN_logloss', seed=0, budget=0.0)]
```

Where an instance 75193_10MIN_logloss was never launched, because the code failed to notice this run was not launched for this config. 

Solution:
Track per configuration what instances have been launched.

### DIFF
--- a/smac/intensification/successive_halving.py
+++ b/smac/intensification/successive_halving.py
@@ -203,7 +203,12 @@ class _SuccessiveHalving(AbstractRacer):
         self.incumbent_selection = incumbent_selection
 
         # Define state variables to please mypy
-        self.curr_inst_idx = 0
+
+        # current instance index tracks two things. Configurations that are to be launched,
+        # That is config A needs to run in 3 instances/seed pairs, then curr_inst_idx should
+        # track this. But then, if a new configuration is added in the case of parallelism
+        # a new separate curr_inst_idx needs to be started
+        self.curr_inst_idx = {}  # type: typing.Dict[Configuration, int]
         self.running_challenger = None
         self.success_challengers = set()  # type: typing.Set[Configuration]
         self.do_not_advance_challengers = set()  # type: typing.Set[Configuration]
@@ -367,12 +372,11 @@ class _SuccessiveHalving(AbstractRacer):
 
         # Make sure that there is no Budget exhausted
         if result.status == StatusType.CAPPED:
-            self.curr_inst_idx = np.inf
+            self.curr_inst_idx[run_info.config] = np.inf
             n_insts_remaining = 0
         else:
             self._ta_time += result.time
             self.num_run += 1
-            self.curr_inst_idx += 1
 
         # adding challengers to the list of evaluated challengers
         #  - Stop: CAPPED/CRASHED/TIMEOUT/MEMOUT/DONOTADVANCE (!= SUCCESS)
@@ -380,9 +384,9 @@ class _SuccessiveHalving(AbstractRacer):
         # curr_challengers is a set, so "at least 1" success can be counted by set addition (no duplicates)
         # If a configuration is successful, it is added to curr_challengers.
         # if it fails it is added to fail_challengers.
-        if np.isfinite(self.curr_inst_idx) and result.status == StatusType.SUCCESS:
+        if np.isfinite(self.curr_inst_idx[run_info.config]) and result.status == StatusType.SUCCESS:
             self.success_challengers.add(run_info.config)  # successful configs
-        elif np.isfinite(self.curr_inst_idx) and result.status == StatusType.DONOTADVANCE:
+        elif np.isfinite(self.curr_inst_idx[run_info.config]) and result.status == StatusType.DONOTADVANCE:
             self.do_not_advance_challengers.add(run_info.config)
         else:
             self.fail_challengers.add(run_info.config)  # capped/crashed/do not advance configs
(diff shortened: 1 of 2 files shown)


smac/intensification/successive_halving.py
```diff
@@ -972,11 +959,13 @@ def _get_pending_instances_for_stage(self, run_history: RunHistory) -> int:
...
-        # But when all configurations have been launched and are running in run history
-        # n_insts_remaining becomes -1, which is confusing. Cap to zero
-        n_insts_remaining = max(len(curr_insts) - self.curr_inst_idx - 1, 0)
+        # self.curr_inst_idx - 1 is the last proposed instance/seed pair from get_next_run
+        # But it is zero indexed, so (self.curr_inst_idx - 1) + 1 is the number of configurations
+        # that we have proposed to run in total for the running configuration via get_next_run
+        n_insts_remaining = len(curr_insts)
+        if self.running_challenger is not None:
+            n_insts_remaining = max(
+                len(curr_insts) - self.curr_inst_idx[self.running_challenger], 0)
```
mfeurer, line 968: Why is this only calculated for `self.running_challenger` and not for all entries in `self.curr_inst_idx`?

mfeurer: Or maybe the current challenger should be an argument to this function to make it easier to understand it's purpose?

smac/intensification/successive_halving.py
```diff
@@ -1003,12 +992,9 @@ def _launched_all_configs_for_current_stage(self, run_history: RunHistory) -> bo
...
 
-        # _count_running_instances_for_challenger will count the running instances
-        # of the last challenger. It makes sense here, because we assume that if we
-        # moved to a new challenger, all instances have been launched for a previous
-        # challenger
-        running_instances = self._count_running_instances_for_challenger(run_history)
-        n_insts_remaining = len(curr_insts) - (self.curr_inst_idx + running_instances)
+        n_insts_remaining = len(curr_insts)
+        if self.running_challenger is not None:
+            n_insts_remaining = len(curr_insts) - self.curr_inst_idx[self.running_challenger]
```
mfeurer, line 997: shouldn't this check be moved in one of the for-loops below? Or can this maybe be simplified somehow? It appears to be working, but I find this function to be very complicated now.

mfeurer: Looking at this function again, doesn't it do the same thing as the function above except that the return value is either `True` or `False` and not an integer?

smac/intensification/successive_halving.py
```diff
@@ -366,21 +366,43 @@ def process_results(self,
...
             self._ta_time += result.time
             self.num_run += 1
 
+        # 0: Before moving to a new stage, we have to complete M x N tasks, where M is the
+        # total number of configurations evaluated in N instance/seed pairs.
+        # The last active configuration is stored in self.running challengers is the M
+        # configuration, and to get to this point, we have already submitted tasks
+        # for (M - 1) configurations on N instances seed-pairs. The status of such
+        # (M - 1) * N tasks is tracked in self.run_tracker, that has a value of False
+        # if not processed nad true if such task has been processed.
```
mfeurer, line 388:
```suggestion
        # if not processed and true if such task has been processed.
```

smac/intensification/successive_halving.py
```diff
@@ -939,26 +955,16 @@ def _top_k(self,
         top_configs = configs_sorted[:k]
         return top_configs
 
-    def _count_remaining_instances_for_stage(
+    def _count_remaining_tasks_for_stage(
```
mfeurer, line 958: Maybe we should rename this to `_count_remaining_inst_seed_pairs_for_stage` to be more in SMAC lingo

smac/intensification/successive_halving.py
```diff
@@ -993,78 +1015,16 @@ def _count_remaining_instances_for_stage(
...
-        # currently running configuration
-        num_chal_available = (
-            len(self.success_challengers | self.fail_challengers | self.do_not_advance_challengers)
-            + self.fail_chal_offset + len(running_configs)
-        )
-        if num_chal_available == self.n_configs_in_stage[self.stage] and n_insts_remaining <= 0:
-            return True
-        else:
-            return False
+        return total_pending_configurations + pending_instances_to_launch
```
mfeurer, line 1027: Is this adding number of configurations and number of instance-seed pairs together? I guess it works, but could this be changed to make the whole code more intuitive?

mfeurer: My criticism is that the function returns a number greater than zero if something still needs to be run, but that's not what the name suggests. This is used in two places, and in both places it is more or less used as a bool. Is it possible to just return a bool and adapt the remaining code to this?

mfeurer (changes requested):

mfeurer (approved):

PR #712, opened 2020-12-08
Only suggest jobs if workers available

franchuterivera:
Check if workers are idle, and only then launch jobs in the case of the simple intensifier.

### DIFF
--- a/smac/intensification/simple_intensifier.py
+++ b/smac/intensification/simple_intensifier.py
@@ -2,12 +2,12 @@ import typing
 
 import numpy as np
 
+from smac.configspace import Configuration
 from smac.intensification.abstract_racer import AbstractRacer, RunInfoIntent
 from smac.optimizer.epm_configuration_chooser import EPMChooser
+from smac.runhistory.runhistory import RunHistory, RunInfo, RunValue
 from smac.stats.stats import Stats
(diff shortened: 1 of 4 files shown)


franchuterivera:
Even after this fix, there are configurations running because of the use of STOP. Example:

```
# Here I print what config id is being proposed (the last configuration)
propopsing to run 21 remaining time=5.285488605499268 and cutoff=30.0
submitted 21 as future <Future: pending, key: run_wrapper-59b951143eece8eae0650f2c67f748a0>
# Here I print the result of this configuration.
...
, instance='{"task_id": "breast_cancer"}', instance_specific='0', seed=0, cutoff=0, capped=False, budget=0.0, source_id=0), RunValue(cost=1.0, time=0.0, status=<StatusType.STOP: 8>, starttime=1607614234.5278208, endtime=1607614234.527821, additional_info={}))
```

Because this run is marked with status stop, this line https://github.com/automl/SMAC3/blob/60584751985b200c69e2bbb32d61f00f7d1d55d4/smac/optimizer/smbo.py#L458
I have moved this line lower to register when there is a stop in the runhistory. After this approach, no more running jobs are left.

mfeurer (approved):

PR #714, opened 2020-12-21
Karlkurzer swig documentation update

karlkurzer:
minor update to the documentation ensuring consistency


mfeurer:
We have lifted the requirement for SWIG to be version 3.0.X, the random forest now also compiles with later versions of SWIG (4.0.X)

PR #715, opened 2021-01-11
Moving to GitHub actions

franchuterivera:
Some relevant comments:
-  I have to move to use --forked as intensification capping testing was failing. That is, any launched run was expected to finish in 1 second yet it took 2 seconds. In my debug it turns out that we change the logging in couple of testing methods, which causes every other testing method to take more time than required. I believe isolated execution with --forked will solve this at multiple levels
-  The other problematic thing was python 3.9 which seems to have a bug with distributed as documented in https://github.com/dask/distributed/issues/4168 .


.pre-commit-config.yaml
```diff
@@ -0,0 +1,20 @@
+repos:
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.761
+    hooks:
+      - id: mypy
+        args: [--show-error-codes, --ignore-missing-imports, --disallow-untyped-decorators, --disallow-incomplete-defs, --disallow-untyped-defs, --follow-imports, skip]
```
mfeurer, line 6: Could you please keep the comments on which arguments we cannot use?

extras_require.json
```diff
@@ -10,5 +10,21 @@
     "sphinx",
     "sphinx_rtd_theme",
     "sphinx-gallery==0.5.0"
+  ],
+  "test": [
+    "pytest>=4.6",
+    "mypy",
+    "pytest-timeout",
+    "flaky",
+    "openml",
```
mfeurer, line 19: I don't think we require openml here as a dependency.

extras_require.json
```diff
@@ -10,5 +10,21 @@
...
+    "pre-commit",
+    "pytest-forked",
+    "pytest-cov"
+  ],
+  "examples": [
```
mfeurer, line 24: I don't think these are required?

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,101 @@
...
+      run: |
+        echo "::set-output name=BEFORE::$(git status --porcelain -b)"
+    - name: Conda Run tests
+      timeout-minutes: 45
+      if: matrix.use-conda == true
+      run: |
+        export OPENBLAS_NUM_THREADS=1
+        export OMP_NUM_THREADS=1
+        export MKL_NUM_THREADS=1
+        # We activate conda as metalearning uses python directly, so we need
```
mfeurer, line 71: This looks like a copy-paste which is not needed here.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        pip install -e .[all]
```
mfeurer, line 15: Hey, could you briefly explain how this works without installing SWIG? I'm actually wondering the same thing for Auto-sklearn right now...

ci_scripts/run_examples.sh
```diff
@@ -1,74 +0,0 @@
-cd examples
```
mfeurer, line 1: Hey, what happened to this script? Are we still testing all the examples?

test/test_intensify/test_intensify.py
```diff
@@ -804,7 +804,8 @@ def target(x):
         # the basis configuration (config3) not better than incumbent, so can move on
         self.assertEqual(inc, self.config1)
         self.assertEqual(self.stats.inc_changed, 2)
-        self.assertEqual(intensifier.stage, IntensifierStage.RUN_INCUMBENT)
+        self.assertEqual(intensifier.stage, IntensifierStage.RUN_INCUMBENT,
+                         msg=self.rh.data.items())
```
mfeurer, line 808: I assume this is only for debugging?

franchuterivera: Yes! because this check was failing and I could not understand why. So I let this for the future in case this happens again. This was solved via --forked.

ci_scripts/run_examples.sh
franchuterivera: Yes! While making the documentation:https://github.com/franchuterivera/SMAC3/runs/1681990495?check_suite_focus=true

But not, if we have a shell script. I missed this files and would add a Github action for this.

.github/workflows/docs.yml
franchuterivera: Swig is there, sorry for the confusion. It is needed when you run, and I have moved this line lower to clarify on this.

extras_require.json
franchuterivera: Sorry, removed.

.github/workflows/pytest.yml
franchuterivera: Sorry, removed.

extras_require.json
franchuterivera: Sorry, removed.

.pre-commit-config.yaml
franchuterivera: Done!

mfeurer (approved):

PR #719, opened 2021-03-10
Adding license_file to setup.cfg

cyprienc:
Hi,

I've added the entry ``license_file`` into the file ``setup.cfg`` so that it's also packaged in.
And I've also bumped the micro-version. 
Thanks,

Cyprien


cyprienc:
Let me know if the changelog has to be updated.

mfeurer:
Yes, that would be great. Could you please also make your change point to the development branch?

PR #722, opened 2021-04-08
new example, gradient boosting from sklearn

jeschment:
As discussed with @mlindauer, I added another example.


mlindauer:
@mfeurer would you agree with this configspace? Or do you see room for improvements?

mfeurer:
Hey @mlindauer the space looks fine, I'd do it a bit different:

* `min_samples_split`: I find an integer easier to interpret. Also, going from 0.01 to 1 is a suboptimal range. If we have 5k training data, the lower bound you can configure for a split is 50 samples. And the upper bound can basically mandate to do a split on 5k samples. I'd use [2, 200] on l logscale or something similar.
* `subsample` is usually tuned between 0.1 and 0.5

However, I'm wondering, what does this example add over https://automl.github.io/SMAC3/master/examples/SMAC4HPO_rf.html#sphx-glr-examples-smac4hpo-rf-py ? It's a bit shorter, so we maybe want to replace the other one, but the idea is the same, isn't it?

Please let me know if whether I should also review the rest of the example.

mlindauer:
@mfeurer Thanks for your feedback. I was thinking tuning gradient boosting could be more interesting than RFs these days. However, it's not hurting us to have more examples, right? 

@jeschment could you please implement the changes proposed by @mfeurer and see whether the results improve?

mfeurer:
@mlindauer indeed. Also, sklearn has a much better interface than the pyrfr. Maybe we can delete the other example then?

mlindauer:
@mfeurer fair point if we would have used the pyrfr in the other example you linked. But in fact, we also used sklearn there. (However, I'm also happy if we delete the other example.)

jeschment:
@mfeurer Thanks again for your feedback. 
@mlindauer The change of the subsample tuning space didn't alter the performance much.
If the min_sample_split is changed to [2,200], the performance decreases and using the bounds [2,2000] leads to similar performance as using [0.01, 1]. 
So the proposed changes do not improve the performance, but they don't make it worse either. Shall I implement them in the pull request nevertheless?

PR #734, opened 2021-07-16
Save runhistory/stats after each change

renesass:
Instead of calling self.runhistory.save_json and stats.save() only once after the optimization is finished, this implementation calls it in the end of incorporate_run_results. Hence, every time a new trial is finished, runhistory and states are saved.


benjamc (commented):
does it introduce overhead? @AndreBiedenkapp

KEggensperger (approved):

PR #765, opened 2021-08-18
Interrupt optimization through IncorporateRunResultCallback

thomashlvt:
I wanted to implement a hook in autosklearn for the user to be able to implement his own stopping strategy. @eddiebergman suggested [here](https://github.com/automl/auto-sklearn/issues/1223) to use the `IncorporateRunResultCallback` of SMAC. However, currently the callback cannot interrupt the optimization loop.

I now added the possibility to gracefully abort the optimization by returning `False` from the callback (when returning None the optimization still continues for backward compatibility). I also added a unit test to test this.

Please let me know if there is anything I should change/add/fix. I'm very happy to contribute :)

### DIFF
--- a/smac/callbacks.py
+++ b/smac/callbacks.py
@@ -1,4 +1,5 @@
 from typing import TYPE_CHECKING
+from typing import Optional
 
 if TYPE_CHECKING:
     from smac.optimizer.smbo import SMBO
@@ -31,12 +32,16 @@ How to add a new callback
 
 class IncorporateRunResultCallback:
 
-    """Callback to react on a new run result. Called after the finished run is added to the runhistory."""
+    """Callback to react on a new run result.
+
+    Called after the finished run is added to the runhistory.
+    Optionally return `False` to (gracefully) stop the optimization.
+    """
 
     def __call__(
             self, smbo: 'SMBO',
             run_info: RunInfo,
             result: RunValue,
             time_left: float,
-    ) -> None:
+    ) -> Optional[bool]:
         pass
--- a/test/test_smbo/test_smbo.py
+++ b/test/test_smbo/test_smbo.py
@@ -50,7 +50,8 @@ class TestSMBO(unittest.TestCase):
     def setUp(self):
         self.scenario = Scenario({'cs': test_helpers.get_branin_config_space(),
                                   'run_obj': 'quality',
-                                  'output_dir': 'data-test_smbo'})
+                                  'output_dir': 'data-test_smbo',
+                                  "runcount-limit": 5})
         self.output_dirs = []
         self.output_dirs.append(self.scenario.output_dir)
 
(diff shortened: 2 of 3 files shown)


test/test_smbo/test_smbo.py
```diff
@@ -50,7 +50,8 @@ class TestSMBO(unittest.TestCase):
     def setUp(self):
         self.scenario = Scenario({'cs': test_helpers.get_branin_config_space(),
                                   'run_obj': 'quality',
-                                  'output_dir': 'data-test_smbo'})
+                                  'output_dir': 'data-test_smbo',
+                                  "runcount-limit": 5})
         self.output_dirs = []
```
eddiebergman, line 55: Was there a need to include a "runcount-limit"?

eddiebergman (commented):
Hi @thomashlvt,

Glad you could get this done so quickly. I looked through the code and the test and it seems fairly straightforward and non-disruptive which is very nice! I have added one question in the review.

I imagine this feature will wanted to be documented somewhere but there should be someone more involved with SMAC to review this properly.

If there is no activity on this soon I will bump someone to take a look

test/test_smbo/test_smbo.py
thomashlvt: If the `TestCallback` would not return `False` at some point, or if stopping the optimization loop would not work, the test would run forever. This is not the case, but I thought it'd be better to have this safety net. I'm not sure if this has an impact on the other tests. Let me know if you would like me to remove this, or constrain it to `test_incorporate_run_results_callback_stop_loop`.

eddiebergman: This makes sense to me but hopefully someone more familiar with all the tests can decide.

smac/callbacks.py
```diff
@@ -1,4 +1,5 @@
 from typing import TYPE_CHECKING
+from typing import Optional
```
eddiebergman, line 3: The double `from typing import X` can be reduced to 1

test/test_smbo/test_smbo.py
mfeurer: Looks good to me

mfeurer (commented):
Looks good to me and I agree with all suggestions from @eddiebergman

eddiebergman:
@renesass Would you have any suggestions on where to document this? I'd be happy to merge it functionally but I feel like this functionality will get lost and forgotten about unless it's written prominently in the docs somewhere.

My suggestion would be to make a new `early_stopping.rst` file under `doc` and add it to the toc tree of `manual.rst`. This way it has a dedicated header entry in the manual as something SMAC can do.  I don't think a full blown example is necessary but a single paragraph with a code snippet demonstrating it would be nice.

Creating a dedicated file and manual section for it also means less barrier to modifying the docs for this in the future.

renesass:
I'd suggest to add "Advanced Usage" below "Basic Usage" under the manual tree. We can add a small example in this file for the early stopping scenario too. In the future, we can always update this file instead of creating others.

eddiebergman:
Did we ever add documentation on this? Sorting out autosklearn issues and it would be good to be able to link to any documentation on this functionality?

renesass:
We didn't.

eddiebergman:
Cool no problem, we have it as an issue, we will bug you about it when we deal with it.

PR #773, opened 2021-09-13
Handle invalid neighbors in local search gracefully

filipbartek:
The iterator returned by `ConfigSpace.util.get_one_exchange_neighborhood` may visit an invalid configuration. If it does so, it either raises a `ValueError` (with the probability 0.05) or yields the configuration (otherwise). This pull request ensures that `LocalSearch._do_search` handles these situations gracefully.

1. If `ValueError` is raised, it is caught and ignored (instead of crashing `smac`).
2. If an invalid configuration is visited and scores better than the current candidate, it is ignored (instead of replacing the current candidate).

The script `run.sh` in [demo.zip](https://github.com/automl/SMAC3/files/7156367/demo.zip) demonstrates raising of `ValueError` from the iterator. The error is raised with a probability greater than 0 and smaller than 1.

Issue at ConfigSpace that discusses the non-determinism of the configuration validation: https://github.com/automl/ConfigSpace/issues/194

### DIFF
--- a/smac/configspace/__init__.py
+++ b/smac/configspace/__init__.py
@@ -3,6 +3,7 @@ from functools import partial
 from ConfigSpace import ConfigurationSpace, Configuration, Constant, \
     CategoricalHyperparameter, UniformFloatHyperparameter, \
     UniformIntegerHyperparameter, InCondition
+from ConfigSpace.exceptions import ForbiddenValueError
 from ConfigSpace.read_and_write import pcs, pcs_new, json
 from ConfigSpace.util import get_one_exchange_neighbourhood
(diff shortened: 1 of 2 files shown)


dengdifan:
Looks good to me! Thanks for the contribution!

dengdifan (approved):

mfeurer:
@dengdifan just wanted to make you aware that this only fixes symptoms that are caused by a bug in the ConfigSpace and should be fixed there. Therefore, once they are fixed, this change should be rolled back (and you should already create an issue for that ore something else to remember to undo these changes).

PR #775, opened 2021-09-22
Propagate minR and maxR from scenario into intensifier

filipbartek:
Ensure that the options `minR`, `maxR`, and `use_ta_time` are propagated from the scenario into the intensifier in case the intensifier class is not provided explicitly to `SMAC4AC` (namely when `scripts/smac` is called).

### DIFF
--- a/smac/facade/smac_ac_facade.py
+++ b/smac/facade/smac_ac_facade.py
@@ -427,6 +427,9 @@ class SMAC4AC(object):
             'min_chall': scenario.intens_min_chall  # type: ignore[attr-defined] # noqa F821
         }
 
+        if intensifier is None:
+            intensifier = Intensifier
+
         if isinstance(intensifier, Intensifier) \
                 or (intensifier is not None and inspect.isclass(intensifier) and issubclass(intensifier, Intensifier)):
             intensifier_def_kwargs['always_race_against'] = scenario.cs.get_default_configuration()  # type: ignore[attr-defined] # noqa F821
@@ -436,11 +439,8 @@ class SMAC4AC(object):
         if intensifier_kwargs is not None:
             intensifier_def_kwargs.update(intensifier_kwargs)
 
-        if intensifier is None:
-            intensifier_instance = (
-                Intensifier(**intensifier_def_kwargs)  # type: ignore[arg-type] # noqa F821
-            )  # type: AbstractRacer
-        elif inspect.isclass(intensifier):
+        assert intensifier is not None
+        if inspect.isclass(intensifier):
             intensifier_instance = intensifier(**intensifier_def_kwargs)  # type: ignore[arg-type] # noqa F821
         else:
             raise TypeError(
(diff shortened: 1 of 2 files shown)


smac/facade/smac_ac_facade.py
```diff
@@ -436,11 +439,8 @@ def __init__(self,
         if intensifier_kwargs is not None:
             intensifier_def_kwargs.update(intensifier_kwargs)
 
-        if intensifier is None:
-            intensifier_instance = (
-                Intensifier(**intensifier_def_kwargs)  # type: ignore[arg-type] # noqa F821
-            )  # type: AbstractRacer
-        elif inspect.isclass(intensifier):
+        assert intensifier is not None
```
benjamc, line 442: we do not need an assert statement here as it is made sure that the intensifier cannot be None at this point (and I would rather use a ValueError or sth), could you remove it, @filipbartek?

smac/facade/smac_ac_facade.py
```diff
@@ -436,11 +439,8 @@ def __init__(self,
...
+        if inspect.isclass(intensifier):
```
benjamc, line 443: here we check if intensifier is a class but in line 433 we can have an instance. In that case the error would be raised. We should add a check if intensifier is an instance and if so, adopt it.

smac/facade/smac_ac_facade.py
```diff
@@ -427,6 +427,9 @@ def __init__(self,
             'min_chall': scenario.intens_min_chall  # type: ignore[attr-defined] # noqa F821
         }
 
+        if intensifier is None:
+            intensifier = Intensifier
+
         if isinstance(intensifier, Intensifier) \
```
benjamc, line 433: if intensifier is an instance, why do we need to define intensifier_def_kwargs? @dengdifan @renesass 
Or does it not matter because it does not hurt?

benjamc, line 433: if we would allow all these (None, intensifier class, intensifier instance), we should update the documentation to make it clear what to pass to __init__

benjamc:
Thanks @filipbartek for the pull request, it seems to make sense. Could you respond to my comments? After that we can merge.

smac/facade/smac_ac_facade.py
filipbartek: I have removed the assertion.

smac/facade/smac_ac_facade.py
filipbartek: Done.

smac/facade/smac_ac_facade.py
filipbartek: I have modified the code to only define `intensifier_def_kwargs` when `intensifier` is a class.

filipbartek: I have updated the documentation.

filipbartek:
> Could you respond to my comments? After that we can merge.

@benjamc, I have responded to all of your comments.

PR #794, opened 2021-11-05
Docs installation

dengdifan:
remove redundant dependencies checks


smac/__init__.py
```diff
@@ -18,12 +18,10 @@
 with open(os.path.join(os.path.dirname(__file__), 'extras_require.json')) as fh:
     extras_require = json.load(fh)
 
-extras_installed = set()
+extras_installed = set()  # type: set
```
mfeurer, line 21: Is this for-loop still required at all? Right now it is only a `continue`.

PR #802, opened 2021-12-15
Cleaned up workflow files, Adds PEP 561 compliance

eddiebergman:
This PR cleans up the github action workflow files to accomplish few things:

* Add's PEP 561 compliance (exports types so other packages can be aware of them)
* Allow manual `workflow_dispatch` on actions that might require it (can manually trigger them from github UI)
* Prevent the double trigger of actions by making `push` and `pull_request` and more strict.
    * A push to a pull request should no longer cause double the amount of tests to run (along with the other workflows that had `on: [push, pull_request]`
* Some general cleanup, giving names to some actions, adding some linebreaks to break up things, ...
* Disabling the `example.yaml` workflow as it did nothing but install SMAC. (Also removed the parameterization of it)
* `pytest.yaml`:
    * Now scheduled to auto run everyday instead of every week
    * Clean up the body of the steps and move some things to env var
    * Scaffold for matrix that includes windows and mac testing (currently excluded, see comments)
    * Includes tests for Python 3.10
    * Changed the boolean flags in the matrix to just be a categorical, easier to read

Once I figure out how to make the tests work windows and macos, I will create a new PR and enable windows/mac testing.


.github/workflows/examples.yml
```diff
@@ -1,22 +1,38 @@
...
+
   ubuntu:
+    if: false  # This workflow aciton does nothing at the moment
 
-    runs-on: ubuntu-18.04
-    strategy:
-      matrix:
-        python-version: [3.9]
-      fail-fast:  false
+    runs-on: ubuntu-lateset
```
benjamc, line 25: typo? lateset --> latest?

.github/workflows/examples.yml
```diff
@@ -1,22 +1,38 @@
...
-    - uses: actions/checkout@v2
-    - name: Setup Python ${{ matrix.python-version }}
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Setup Python 3.9
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: '3.9'
```
benjamc, line 34: should we always use the latest python version?

benjamc: and does the python version ref in line 23/39
   ```$CONDA/bin/conda create -n testenv --yes pip wheel gxx_linux-64 gcc_linux-64 swig python=${{ matrix.python-version }}```
need to be changed to ${{ python-version }} ?

.github/workflows/examples.yml
```diff
@@ -1,22 +1,38 @@
+# NOTE: This workflow seems to do nothing but install the library through conda
```
benjamc, line 1: do we need this then? do we test somewhere else if it can be installed through conda? why is it called examples?

benjamc: I saw we test it in pytest.yml.
Remove this file?

benjamc:
Looks good!
Should we remove .github/workflows/examples.yml?
@renesass @dengdifan @mfeurer

eddiebergman:
I would remove that file as it doesn't really do anything but I wasn't sure, so I just deactivated it for now

mfeurer:
That file was there to run the commandline examples. It would be better to reactivate this behavior to ensure that these examples are working as expected.

eddiebergman:
That's not what it's doing but isn't that covered with `sphinx-gallery` in the docs workflow?

eddiebergman:
I also fixed the PEP 561 compliance check and then also added the compliance.

This means a user using SMAC with ```import smac``` will have access to the type information of smac's API in their editor when they use it.

mfeurer:
> That's not what it's doing but isn't that covered with sphinx-gallery in the docs workflow?

No, `sphinx-gallery` should only cover the python examples. The command line examples are not covered by that, but should be tested as well. And it used to do this, not sure when this was disabled.

renesass:
That's on me. When integrated new docs, we couldn't get the tests working and decided to exclude it for now. But yeah, we will integrate them again in 1.1.2.

eddiebergman:
Cool, I'll leave those tests as a seperate PR, this one is ready to merge whenever you guys are happy with it.

PR #813, opened 2022-01-17
PiBO

hvarfner:
#### New Added unit tests for PriorAcquisitionFunction (and required supporting classes)
- Tests for the _compute_prior for different shapes
- Tests for standard computation for both EI (does not require rescaling) and TS (does require rescaling)
- Tests for discretizing the prior (relevant for random forests 
#### Added unit test for LocalAndSortedPriorRandomSearch 
- Since this one is pretty much tested (it just appends two different random searches to one another), there is really only one thing to test to me, but I may be wrong
#### Removed the need to sample the prior to get the co-domain of the prior
- Moved to closed-form in ConfigSpace
#### Changed input kwargs in facade
- Now called user_prior_kwargs (never got acquisition_function_kwargs to work, since the underlying acquisition function gets arguments it does not expect, like decay_beta)
#### Confirmed issue combining inactive with Normal / Beta hyperparameters 
- Due to what I believe is some old TPE support, Normal parameters are not normalized to 0-1 range (presumably since they can be used without upper and lower bounds
- Inactive parameters are flagged with a -1 value (right?) for Uniform parameters, which does not mesh with the possibly unbounded nature of Normal parameters
- As such, one could force Normal parameters to be bounded, normalize, and allow for PiBO with inactive parameters, but there may be some other issue regaring this that I fail to see. Simply put, I don't know if it's OK to forcibly bound and normalize Normal parameters in ConfigSpace.
#### Added beauitified example usage
- May not be the most intuitive, but at least it showcases how to use it
#### And of course, right branch this time


smac/facade/smac_ac_facade.py
```diff
@@ -83,6 +84,8 @@ def __init__(self,
                  acquisition_function: Optional[Type[AbstractAcquisitionFunction]] = None,
                  acquisition_function_kwargs: Optional[Dict] = None,
                  integrate_acquisition_function: bool = False,
+                 user_priors: bool = False,
+                 user_prior_kwargs: Optional[Dict] = None,
```
benjamc, line 88: @dengdifan just use one arg and put `user_priors` into `user_prior_kwargs`?

smac/optimizer/acquisition.py
```diff
@@ -178,6 +179,161 @@ def _compute(self, X: np.ndarray) -> np.ndarray:
...
+
+        # check if the acquisition function is LCB or TS - then the acquisition function values
+        # need to be rescaled to assure positiveness & correct magnitude
+        if isinstance(self.acq, IntegratedAcquisitionFunction):
+            acquisition_type = self.acq.acq
+        else:
+            acquisition_type = self.acq
+
+        self.rescale_acq = isinstance(acquisition_type, LCB) or isinstance(acquisition_type, TS)
+        self.iteration_number = 0
```
dengdifan, line 231: You also need to set `self._required_updates` to fit the `acquisition_type` accordingly (see `EI.__init__` as an example)
Additionally, `eta` also needs to be included.

benjamc:
Hi @hvarfner ,
thanks very much for working in the requested changes and the thorough tests!
@dengdifan will also have a look in case you and I have missed something.
But otherwise it looks fine for me to merge :)

dengdifan:
Thanks for the PR.
Overall it looks fine to me except for some small issues. Please have a check. After that, I think it is fine to merge  this PR

hvarfner:
Sounds great, thanks! I will make sure to provide a fixed version by EoD, when the new version of ConfigSpace is done, too!

smac/optimizer/acquisition.py
hvarfner: @dengdifan Can you please elaborate here? I'm not sure I follow completely.

hvarfner: Nevermind, I think I've got it!

hvarfner: @dengdifan Sorry, I think I got confused again. Since we overwrite update in PriorAcquisitionFunction, there should be no need for self._required_updates there, right? The underlying acquisition function already has its self._required_updates, and is accordningly called as 
`        self.acq.update(**kwargs)`
in line 244. Shouldn't this be enough?

hvarfner:
@benjamc @dengdifan I will try to make the example a bit more sophisticated (and readthedocs-friendly, as I don't think it is). Otherwise, I think the changes are taken care of, but please enlighten me otherwise!

hvarfner:
Hi,

There was one more change that needed to be made. Categorical parameters with probabilities (which all categorical parameters will now have as per default) cannot be written to a pcs file. I do not know the ins and outs of the format, and as was suggested by @mfeurer , this could be worked around by simply not generating such a file and relying strictly on the JSON. I took the liberty of implementing this (about 4 lines) myself, so please check if this is something you think is reasonable.

Moreover, I wanted a more sophisticated example to better showcase the intended usage. Please let me know what you think of this.

I do not plan for any more changes from now on, so please let me know if you see the need for anything else! @benjamc @dengdifan

benjamc (approved):

benjamc:
Hi @hvarfner ,
the example looks great! The explanations why which prior is used are really helpful.
Of course, we should update the link to pibo in the example (and everywhere else) once it is available.
Now we just need to wait for the ConfigSpace PRs.
Thanks a lot for your effort! 
🌞

benjamc:
We need to wait for the new ConfigSpace release and then see if the tests are running through. After that, we can merge. :sun_with_face:

eddiebergman:
I'll release a new ConfigSpace later today :)

hvarfner:
Great, thanks!

eddiebergman:
You can check progress [here](https://github.com/automl/ConfigSpace/runs/5675656077), I bumped the version notes and now the wheels are building for Windows/Linux/Mac ... Once those are done I can publish a release :)

eddiebergman:
One more point: Considering it's a fairly major change to ConfigSpace, I bumped from `0.4.21` -> `0.5.0`. This means the `requirements.txt` will need to be updated. I did this change as it will also effect many other libraries that rely on ConfigSpace, it should hopefully prevent anyone being safe and relying on the `0.4.x` from experiencing sudden issues.

eddiebergman:
Sorry more notes, the dependency cycle continues. Autosklearn can't use new ConfigSpace until there's a new SMAC release. It's not urgent but just that I will be nudging this along while possible.

hvarfner:
Hi,

If I'm not mistaking, we waited for configspace to be in place before, right? Is there more that needs to be done? If so, please let me know what I can do!

eddiebergman:
Hi @hvarfner,

As previously mentioned, the requirements file needs to be updated so that the right version of ConfigSpace gets pulled in. If you look at the test errors, you can see that `ConfigSpace has no variable BetaFloatHyperparameter` and other such errors. Hopefully there'll be no bugs to fix with that but we'll have to see :)

Best,
Eddie

renesass:
We have another PR going on right now, which updates the files based on updated tools (black, flake8, ...). We would love to have PiBO in this "cleaned" state, although it takes a bit more time. I can't promise it but most likely we can merge it Thursday.

Sorry for the late responses - a lot going on recently.

mlindauer:
Hi Rene and Eddie,
ICLR is in a few weeks and of course, we want to advertise piBO as part of SMAC (and Hypermapper). So please make sure that this is done before the ICLR starts -- earlier would be even better because people have watched Carl's yt video already.

smac/facade/smac_ac_facade.py
```diff
@@ -361,12 +370,27 @@ def __init__(
...
 
+        if user_priors:
+            # a solid default value for decay_beta - empirically founded
+            default_beta = scenario.ta_run_limit / 10
+            decay_beta = acq_def_kwargs.get("decay_beta", default_beta)
+            discretize = isinstance(model_instance, RandomForestWithInstances) or isinstance(
+                model_instance, RFRImputator
+            )
+            acquisition_function_instance = PriorAcquisitionFunction(
+                acquisition_function=acquisition_function_instance, **user_prior_kwargs, **acq_def_kwargs
```
renesass, line 381: Both decay_beta and discretize are not used. Should they be part of "user_prior_kwargs"?

renesass (changes requested):

smac/facade/smac_ac_facade.py
renesass: After this issue is fixed, we can merge it. :)

hvarfner: Great to hear! I'll check it right away.

hvarfner: Addressed the issue now, thanks for spotting.

PR #817, opened 2022-02-01
Multi-Objective

renesass:
Since the first Multi-Objective version was not working, this PR should fix it.


smac/utils/multi_objective.py
```diff
@@ -0,0 +1,30 @@
+from typing import Union, List, Tuple
+import numpy as np
+
+
+def normalize_costs(values: Union[np.ndarray, list],
+                    bounds: List[Tuple[float, float]] = None) -> np.ndarray:
```
eddiebergman, line 6: Some doc on `normalize_costs` requiring a 2 dimensional array, list would be nice :)

Also, just incase you're interested with respect to `list` or `List` for typing: https://stackoverflow.com/a/62033243

I've resigned to just using `List` everywhere for now and update it later once the deprecation warnings start to hit (years from now)

smac/utils/multi_objective.py
```diff
@@ -0,0 +1,30 @@
...
+
+    if isinstance(values, list):
+        values = np.array(values)
+
+    normalized_values = np.empty_like(values)
+    for axis in range(values.shape[1]):
```
eddiebergman, line 12: Seems off to me, does `values.shape[1]` really contain all the axes. Maybe it's meant to be `column`?

```python
values = np.array([[1,2,3],[4,5,6]])
values.shape[1] == 3  # Yet there's only two axes?
list(range(values.shape[1]))) # [0,1,2]
```

smac/utils/multi_objective.py
```diff
@@ -0,0 +1,30 @@
...
+        data = values[:, axis]
+        
+        if bounds is not None and len(bounds) == values.shape[1]:
```
eddiebergman, line 15: If bounds can be None, maybe the parameter type should read `Optional[...]`

smac/epm/gaussian_process.py
```diff
@@ -112,12 +112,7 @@ def _train(self, X: np.ndarray, y: np.ndarray, do_optimize: bool = True) -> 'Gau
         X = self._impute_inactive(X)
         if self.normalize_y:
             y = self._normalize_y(y)
-        if len(y.shape) == 1:
-            self.n_objectives_ = 1
-        else:
-            self.n_objectives_ = y.shape[1]
-        if self.n_objectives_ == 1:
-            y = y.flatten()
+        y = y.flatten()
```
benjamc, line 115: y now always is a vector/array and not a scalar, right?

smac/facade/smac_ac_facade.py
```diff
@@ -558,8 +554,8 @@ def __init__(self,
             'scale_perc': 5,
         }
 
+        # TODO: consider other sorts of multi-objective algorithms
```
benjamc, line 557: is this TODO up to date? by allowing arbitrary AggregationStrategy we can have other MO algorithms?

smac/optimizer/smbo.py
```diff
@@ -126,7 +127,7 @@ def __init__(self,
 
         self.initial_design_configs = []  # type: typing.List[Configuration]
 
-        # TODO consider if we need an additional EPMChooser for multi-objective optimization
+        # TODO: consider if we need an additional EPMChooser for multi-objective optimization
```
benjamc, line 130: do we?

smac/utils/io/cmd_reader.py
```diff
@@ -550,10 +550,10 @@ def _add_smac_options(self) -> None:
                                     "the target algorithm crashes.")
         smac_opts.add_argument("--limit-resources", "--limit_resources",
                                dest='limit_resources',
-                               default=True, type=truthy,
+                               default=False, type=truthy,
```
benjamc, line 553: are there any issues with resources if we don't limit them?

smac/utils/multi_objective.py
```diff
@@ -0,0 +1,57 @@
...
+            min_value = bounds[col][0]
+            max_value = bounds[col][1]
+        else:
+            min_value = np.min(data)
+            max_value = np.max(data)
+
+        denominator = max_value - min_value
+
+        # Prevent divide by zero
+        if denominator == 0:
```
benjamc, line 47: maybe don't exactly compare to 0 but `denominator < epsilon` to avoid numerical issues.
E.g. by using np.isclose

smac/utils/multi_objective.py
```diff
@@ -0,0 +1,57 @@
...
+    normalized_values = []
+    for col in range(values.shape[1]):
+        data = values[:, col].astype(float)
+
+        if bounds is not None and len(bounds) == values.shape[1]:
+            min_value = bounds[col][0]
+            max_value = bounds[col][1]
+        else:
+            min_value = np.min(data)
+            max_value = np.max(data)
```
benjamc, line 42: just to be sure, we are always using the current min/max and do not calculate them once?

smac/runhistory/runhistory.py
```diff
@@ -143,8 +166,8 @@ class RunHistory(object):
     """
 
     def __init__(
-            self,
-            overwrite_existing_runs: bool = False,
+        self,
+        overwrite_existing_runs: bool = True,
```
benjamc, line 170: why do we overwrite existing runs by default? was does this do/what is the effect?

smac/epm/gaussian_process.py
benjamc: y always is a scalar

smac/facade/smac_ac_facade.py
benjamc: leave, is up to date

smac/optimizer/smbo.py
benjamc: not right now

smac/utils/io/cmd_reader.py
benjamc: - [ ] add info that it only works on linux

smac/utils/multi_objective.py
benjamc: yes

smac/runhistory/runhistory.py
benjamc: new run for same metadata not updated, but will be now

renesass:
Yes, I was just removing some warnings on the way while fixing the tests. The tools will be updated in the next PR.

test/test_intensify/test_intensify.py
```diff
@@ -147,6 +162,7 @@ def target(x):
             capped=True,
             budget=0.0,
         )
+        print(type(run_info.cutoff))
```
renesass, line 165: Remove

renesass (commented):
I don't find the error why the intensification fail.

- You can pass list/array/float/int to runhistory and internally it's saved as list or float (if num_obj == 1)
- average_cost, sum_cost, min_cost however use `normalize_cost` to turn the internal costs (which can be a list) to a single value
- Therefore, everything should work as no multi-objective was implemented
- I changed `limit_resources` and `deterministic` to True and changed the scenario objects in the tests accordingly.
- However, `test_intensify.py` and `test_successive_halving.py` use the TAFunc directly. I set use_pynisher to False so I can test it on Mac. Setting it to True doesn't change the errors in the tests on GitHub.

Any help will be appreciated.

smac/tae/base.py
```diff
@@ -247,18 +247,19 @@ def run_wrapper(
 
         # Catch NaN or inf.
         if (
-            self.run_obj == 'runtime' and not np.isfinite(runtime)
-            or self.run_obj == 'quality' and not np.all(np.isfinite(cost))
+            self.run_obj == "runtime"
+            and not np.isfinite(runtime)
+            or self.run_obj == "quality"
+            and not np.all(np.isfinite(cost))
```
eddiebergman, line 253: While the order of operations works here, I would bracket these statements together.
```python
if (
    (self.run_obj == "runtime" and not np.isfinite(runtime))
    or (self.run_obj == "quality" and not np.all(np.isfinite(cost)))
):

```

smac/utils/multi_objective.py
```diff
@@ -0,0 +1,57 @@
...
+    -------
+    np.ndarray
+        Normalized costs
+    """
+
+    if isinstance(values, list):
+        values = np.array(values)
+
+    if len(values.shape) == 1:
+        values = np.expand_dims(values, axis=0)
```
eddiebergman, line 31: Why does `costs = [3,4,5]` need to be expanded? Is this treated as the cost of a single run with three multi-objective costs `[3,4,5]` or as three separate runs with each having a cost `3`, `4` and `5`?

smac/utils/multi_objective.py
```diff
@@ -0,0 +1,57 @@
...
+
+    normalized_values = []
+    for col in range(values.shape[1]):
+        data = values[:, col].astype(float)
+
+        if bounds is not None and len(bounds) == values.shape[1]:
+            min_value = bounds[col][0]
+            max_value = bounds[col][1]
+        else:
```
eddiebergman, line 40: What if `bounds is not None and len(bounds) != values.shape[1]`, should probably raise an explicit error in this case.

smac/utils/multi_objective.py
eddiebergman: I second this if the gap between `max_value` and `min_value` is exceptionally small, it's likely due to the cost being slightly unstable. If your costs are that sensitive then "wtf".
```python
if denominator < 1e-10:
  ...
```

Alternatively, check the denominator for this small difference and issue a warning but continue as it already does.

smac/utils/multi_objective.py
```diff
@@ -0,0 +1,57 @@
...
+
+        # Prevent divide by zero
+        if denominator == 0:
+            # Return ones
+            normalized_values.append(np.ones_like(data))
+        else:
+            numerator = data - min_value
+            normalized_values.append(numerator / denominator)
+
+    normalized_values = np.array(normalized_values)
```
eddiebergman, line 54: Seems like this line isn't needed as they're already `np.ndarray`

smac/runhistory/runhistory.py
```diff
@@ -372,10 +450,12 @@ def incremental_update_cost(self, config: Configuration, cost: Union[np.ndarray,
 
         config_id = self.config_ids[config]
         n_runs = self.num_runs_per_config.get(config_id, 0)
-        old_cost = self._cost_per_config.get(config_id, np.zeros(self.num_obj).squeeze())
+        old_cost = self._cost_per_config.get(config_id, 0.0)
```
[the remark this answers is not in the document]
renesass: Can you elaborate?

smac/utils/multi_objective.py
renesass: normalized_values is only a native python list

smac/utils/multi_objective.py
renesass: [3,4,5] means three runs. Therefore, we want to have [[3], [4], [5]]. Added examples in the docstring.

examples/python/plot_scalarized_multi_objective.py
```diff
@@ -46,7 +51,9 @@ def is_pareto_efficient_simple(costs):
     is_efficient = np.ones(costs.shape[0], dtype=bool)
     for i, c in enumerate(costs):
         if is_efficient[i]:
-            is_efficient[is_efficient] = np.any(costs[is_efficient] < c, axis=1)  # Keep any point with a lower cost
+            is_efficient[is_efficient] = np.any(
+                costs[is_efficient] < c, axis=1
+            )  # Keep any point with a lower cost
```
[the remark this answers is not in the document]
renesass: That should be correct.

smac/optimizer/multi_objective/abstract_multi_objective_algorithm.py
```diff
@@ -5,13 +5,16 @@
 
 class AbstractMultiObjectiveAlgorithm(ABC):
```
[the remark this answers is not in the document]
renesass: Makes no sense because this only makes sense for aggregations. Multi-Objective can also be done non-aggregated.

examples/python/spear_qcp/scenario.txt
```diff
@@ -2,6 +2,7 @@ algo = python -u ./spear_qcp/target_algorithm/scripts/SATCSSCWrapper.py --mem-li
 paramfile = ./spear_qcp/target_algorithm/spear-python/spear-params-mixed.pcs
 execdir = .
 deterministic = 0
+limit_resources = 1
```
benjamc, line 5: why do we limit resources here?

smac/facade/func_facade.py
```diff
@@ -16,14 +16,16 @@
...
-              scenario_args: typing.Mapping[str, typing.Any] = None,
-              tae_runner_kwargs: typing.Optional[typing.Dict[str, typing.Any]] = None,
-              **kwargs: typing.Any) -> typing.Tuple[Configuration, typing.Union[np.ndarray, float], SMAC4HPO]:
+def fmin_smac(
+    func: typing.Callable,
+    x0: typing.List[float],
+    bounds: typing.List[typing.Iterable[float]],
+    maxfun: int = -1,
+    rng: typing.Union[np.random.RandomState, int] = None,
+    scenario_args: typing.Mapping[str, typing.Any] = None,
```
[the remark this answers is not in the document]
benjamc: 👍

smac/optimizer/multi_objective/abstract_multi_objective_algorithm.py
```diff
@@ -5,13 +5,16 @@
...
     """
-    a general interface for mutli-objective optimizer, dependeing on different streagies, it can be applied to rh2epm
-    or epmchooser
+    A general interface for mutli-objective optimizer, dependeing on different streagies,
```
benjamc, line 8: typos 🤓

smac/optimizer/multi_objective/abstract_multi_objective_algorithm.py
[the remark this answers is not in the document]
benjamc: let's keep it as is

smac/runhistory/runhistory.py
```diff
@@ -282,41 +311,76 @@ def add(
...
+        all_costs = []
+        for (costs, _, status, _, _, _) in self.data.values():
+            if status == StatusType.SUCCESS:
+                if not isinstance(costs, Iterable):
+                    costs = [costs]
+
+                all_costs.append(costs)
+
+        all_costs = np.array(all_costs, dtype=float)
+        self.objective_bounds = [(np.inf, -np.inf)] * self.num_obj
```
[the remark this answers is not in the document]
benjamc: ❓  👍

smac/runhistory/runhistory.py
```diff
@@ -393,11 +473,11 @@ def get_cost(self, config: Configuration) -> float:
             Computed cost for configuration
         """
         config_id = self.config_ids.get(config)
-        return self._cost_per_config.get(config_id,                                # type: ignore[arg-type] # noqa F821
-                                         np.full(self.num_obj, np.nan).squeeze())  # type: ignore[arg-type] # noqa F821
+        return self._cost_per_config.get(config_id, np.nan)  # type: ignore[arg-type] # noqa F821
```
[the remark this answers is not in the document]
benjamc: ❓ 👍

smac/runhistory/runhistory2epm.py
```diff
@@ -659,7 +745,16 @@ def _build_matrix(self, run_dict: typing.Mapping[RunKey, RunValue],
...
                 X[row, :] = conf_vector
-            y[row, 0] = run.cost
+
+            if self.num_obj > 1 and self.multi_objective_algorithm is not None:
+                # Let's normalize y here
+                # We use the objective_bounds calculated by the runhistory
+                y_ = normalize_costs(run.cost, runhistory.objective_bounds)
+                y_ = self.multi_objective_algorithm(y_)
+                y[row, 0] = y_
+            else:
```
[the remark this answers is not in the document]
benjamc: 👍

smac/tae/base.py
```diff
@@ -213,6 +215,7 @@ def run_wrapper(
             )
         cutoff = None
         if run_info.cutoff is not None:
+            # cutoff = np.iinfo(np.int32).max
```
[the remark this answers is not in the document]
benjamc: remove?

test/test_files/restore_scenario_one.txt
```diff
@@ -2,4 +2,6 @@ paramfile = examples/commandline/branin/configspace.pcs
 run_obj = quality
 runcount_limit = 5
 output_dir = test/test_files/test_restore_state
-algo = python examples/quickstart/branin/branin.py
+algo = python examples/commandline/branin.py
+limit_resources = 1
```
benjamc, line 6: same as above, what does it do?

benjamc (approved):

smac/utils/multi_objective.py
dengdifan: But this line result in [[3, 4, 5]]

examples/python/spear_qcp/scenario.txt
renesass: Because `limit_resources` was default before. Otherwise, test will fail.

smac/runhistory/runhistory.py
```diff
@@ -282,41 +311,76 @@ def add(
...
+        """Update the objective bounds based on the data in the runhistory."""
+        all_costs = []
+        for (costs, _, status, _, _, _) in self.data.values():
+            if status == StatusType.SUCCESS:
+                if not isinstance(costs, Iterable):
+                    costs = [costs]
+
+                all_costs.append(costs)
+
+        all_costs = np.array(all_costs, dtype=float)
```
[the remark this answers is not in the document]
renesass: The costs are always the same shape (enforced in the `add` method). I added an assert to make sure.

smac/runhistory/runhistory.py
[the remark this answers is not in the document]
renesass: Yes, because of the annoying mypy errors :)

smac/runhistory/runhistory.py
```diff
@@ -282,41 +311,76 @@ def add(
...
+        self.objective_bounds = [(np.inf, -np.inf)] * self.num_obj
+
+        if len(all_costs) == 0:
+            return
+
+        min_values = np.min(all_costs, axis=0)
+        max_values = np.max(all_costs, axis=0)
+
+        for i, (min_v, max_v) in enumerate(zip(min_values, max_values)):
+            self.objective_bounds[i] = (min_v, max_v)
```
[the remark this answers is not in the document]
renesass: Hope you are more satisfied with this solution :)

.github/workflows/pytest.yml
```diff
@@ -138,5 +138,7 @@ jobs:
       if: matrix.code-cov && always()
       uses: codecov/codecov-action@v2
       with:
+        # token: ${{secrets.CODECOV_TOKEN}}
+        version: "v0.1.15"
```
eddiebergman, line 142: Why this version?

PR #824, opened 2022-02-24
Update tools

renesass:


renesass:
Windows and MacOS have to be removed because pynisher is not working on either of them.

mlindauer:
> Windows and MacOS tests have to be removed because pynisher is not working on either of them.

but without pynisher, SMAC should work on both of them, right?

renesass:
> > Windows and MacOS tests have to be removed because pynisher is not working on either of them.
> 
> but without pynisher, SMAC should work on both of them, right?

Yes. Theoretically, pynisher also should work on MacOS. However, it seems like some pickling fails on Mac. This issue will be investigated further.

benjamc:
- [ ] revert  [47819c2](https://github.com/automl/SMAC3/pull/824/commits/47819c2925122728a0b0408fe72449e55b28f337): runhistory can only work with scalars. Change signature/type/content of target dict from `np.ndarray` to `float`

renesass:
`_cost_per_config` is filled by `average_cost`, which should always return a float. In case of MO, `normalize_costs` is used to get only one cost value in the end.

Issue #833, opened 2022-04-18
[Bug?] RunHistory2EPM4EIPS is incrementing time twice before taking logarithm


mhlr:
The time is being incremented in both `_build_matrix`:

https://github.com/automl/SMAC3/blob/863e4290054847ba2688521b8cc2e44c15a1493a/smac/runhistory/runhistory2epm.py#L784

and `transform_response_values`

https://github.com/automl/SMAC3/blob/863e4290054847ba2688521b8cc2e44c15a1493a/smac/runhistory/runhistory2epm.py#L763

dengdifan:
Hi,
thanks for pointing that out! This is indeed a bug, we will fix that in the next version

PR #837, opened 2022-05-05
Multi obj epm

dengdifan:
This PR makes it easier for developing new Multi-Objective models. Also, it will explicitly raise an error if the acquisition function is incompatible with the epm models


smac/epm/base_uncorrelated_mo_model.py
```diff
@@ -0,0 +1,185 @@
...
+            types=types,
+            seed=seed,
+            instance_features=instance_features,
+            pca_components=pca_components,
+        )
+        if model_kwargs is None:
+            model_kwargs = {}
+        self.target_names = target_names
+        self.num_targets = len(self.target_names)
+        print(seed, model_kwargs)
```
benjamc, line 77: Should we keep the print statement?

smac/runhistory/runhistory.py
```diff
@@ -379,8 +379,9 @@ def _add(self, k: RunKey, v: RunValue, status: StatusType, origin: DataOrigin) -
 
             # if budget is used, then update cost instead of incremental updates
             if not self.overwrite_existing_runs and k.budget == 0:
-                # assumes an average across runs as cost function aggregation, this is used for algorithm configuration
-                # (incremental updates are used to save time as getting the cost for > 100 instances is high)
+                # assumes an average across runs as cost function aggregation, this is used for
+                # algorithm configuration (incremental updates are used to save time as getting
+                # Zthe cost for > 100 instances is high)
                 self.incremental_update_cost(self.ids_config[k.config_id], v.cost)
             else:
```
benjamc, line 385: So the cost in runhistory will be still averaged for the new uncorrelated MO?

smac/runhistory/runhistory2epm.py
```diff
@@ -725,5 +725,5 @@ def transform_response_values(self, values: np.ndarray) -> np.ndarray:
         -------
```
benjamc, line 725: We do we not add 1 anymore?

dengdifan: This aims at solving the issue #833

benjamc: Related to #833 , todo: create a separate PR for that:)

benjamc:
We should create an acquisition function that works with this type of surrogate model :)

PR #845, opened 2022-05-20
Link JMLR paper

mfeurer:


benjamc (approved):

benjamc:
We should also update the bibtex reference right? It still mentions the arxiv version. I could do that later

benjamc (changes requested):
TODO: update bibtex reference

Issue #609, opened 2020-03-04
Reference pSMAC correctly


mfeurer:
We should reference pSMAC as https://open.library.ubc.ca/cIRcle/collections/ubctheses/24/items/1.0167184

benjamc:
TODO: Create an example for pSMAC and document it better.

PR #855, opened 2022-06-23
Update pSMAC

### LINKED ISSUE #821: PSMAC test fails
`test/test_facade/test_psmac_facade.py` fails.

benjamc:
- fix #821 
- add reference to pSMAC as mentioned in #609 
- add example
- extend pSMAC capability: can pass a SMAC facade now
- add ResultMerger which can aggregate runhistories from different directories and create a new trajectory from them

- [ ] Missing for pSMAC: Capability to write complete SMAC run data. Trajectory and runhistory already can be written by the ResultMerger, only the proper call in pSMAC is missing (and the rest of the files that should be written).
- [x] check if every facade works with pSMAC --> add tests


renesass:
Please also update the changelog. :)

PR #872, opened 2022-07-04
Fix Multi-Objective cache

### LINKED ISSUE #852: Integrate Multi-Objective caching
The runhistory caching does not work for MO right now.

Steps to do:
- Call `update_cost` if `num_obj` > 1 in _add
- Check if costs are cached in intensifier

Idea: Save metrics and average/normalize only if `get_cost()` is called.

renesass:
Closes #852.

I slightly rearranged the runhistory so that it's more intuitive.

What else did I do?
`average_cost/min_cost/sum_cost` return a list of floats (instead of a float) now.
For example, for `average_cost`, each objective is averaged separately. Imagine you have two runs with [100, 200] and [0, 0] then you'd get [50, 100]. However, once you call `get_cost`, this (cached) [50, 100] is normalized based on all passed entries. That means that `get_cost` returns a float in the end. 

> Unfortunately, the intensifier and also some other methods work with `average_cost` or `sum_cost` (which should actually be a private method). So, I have to call `normalize_cost` in the intensifier again. Since `_cost_per_config`(cache for `average_cost`) is always storing the current state with all objective values, we are caching it correctly and normalize it once it's needed in the `get_cost` method.

Edit: I added an argument in `average_cost/min_cost/sum_cost` to return the normalized (single float) values. Makes the code easier and scalable.

One more thing which really is weird: In the runhistory2epm we iterate over the run dictionary (runkey, runvalue). But here we use the `run.cost` directly and hence we have to call `normalize_costs` in the case of MO again. I wonder why this is done this way? @mfeurer

### DIFF
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -13,6 +13,7 @@ from smac.runhistory.runhistory import RunHistory, RunInfo, RunValue
 from smac.stats.stats import Stats
 from smac.utils.io.traj_logging import TrajLogger
 from smac.utils.logging import format_array
+from smac.utils.multi_objective import normalize_costs
 
 _config_to_run_type = Iterator[Optional[Configuration]]
 
@@ -89,7 +90,6 @@ class AbstractRacer(object):
         maxR: int = 2000,
         adaptive_capping_slackfactor: float = 1.2,
         min_chall: int = 1,
-        num_obj: int = 1,
     ):
 
         self.logger = logging.getLogger(self.__module__ + "." + self.__class__.__name__)
@@ -128,12 +128,6 @@ class AbstractRacer(object):
         self.repeat_configs = False
         # to mark the end of an iteration
         self.iteration_done = False
-
-        if num_obj > 1:
-            raise ValueError(
-                "Intensifiers only support single objective optimization. For multi-objective problems,"
-                "please refer to multi-objective intensifiers"
-            )
 
     def get_next_run(
         self,
@@ -350,6 +344,15 @@ class AbstractRacer(object):
         # if it dominates the incumbent
         chal_perf = run_history.average_cost(challenger, to_compare_runs)
         inc_perf = run_history.average_cost(incumbent, to_compare_runs)
+
+        # Multi-Objective setting: We have to normalize the values here
+        if type(chal_perf) == list or type(inc_perf) == list:
+            assert type(chal_perf) == type(inc_perf)
+            chal_perf = run_history.get_cost(challenger)
+            inc_perf = run_history.get_cost(incumbent)
+        else:
+            assert type(chal_perf) == float
+            assert type(inc_perf) == float
 
         # Line 15
         if np.any(chal_perf > inc_perf) and len(chall_runs) >= self.minR:
--- a/smac/intensification/abstract_racer.py
+++ b/smac/intensification/abstract_racer.py
@@ -296,9 +296,10 @@ class AbstractRacer(object):
         # reasons)
         chall_inst_seeds = run_history.get_runs_for_config(challenger, only_max_observed_budget=True)
         chal_sum_cost = run_history.sum_cost(
-            config=challenger,
-            instance_seed_budget_keys=chall_inst_seeds,
+            config=challenger, instance_seed_budget_keys=chall_inst_seeds, normalize=True
         )
+        assert type(chal_sum_cost) == float
+
         cutoff = min(curr_cutoff, inc_sum_cost * self.adaptive_capping_slackfactor - chal_sum_cost)
         return cutoff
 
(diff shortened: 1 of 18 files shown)


smac/intensification/abstract_racer.py
```diff
@@ -89,7 +89,6 @@ def __init__(
         maxR: int = 2000,
         adaptive_capping_slackfactor: float = 1.2,
         min_chall: int = 1,
-        num_obj: int = 1,
```
timruhkopf, line 92: needs documentation

smac/intensification/abstract_racer.py
```diff
@@ -129,12 +128,6 @@ def __init__(
         # to mark the end of an iteration
         self.iteration_done = False
 
-        if num_obj > 1:
-            raise ValueError(
-                "Intensifiers only support single objective optimization. For multi-objective problems,"
-                "please refer to multi-objective intensifiers"
```
timruhkopf, line 135: Where exactly are you pointing to?

smac/intensification/hyperband.py
```diff
@@ -98,7 +98,6 @@ def __init__(
         min_chall: int = 1,
         incumbent_selection: str = "highest_executed_budget",
         identifier: int = 0,
-        num_obj: int = 1,
```
timruhkopf, line 101: documentation

smac/intensification/intensification.py
```diff
@@ -144,7 +144,6 @@ def __init__(
         maxR: int = 2000,
         adaptive_capping_slackfactor: float = 1.2,
         min_chall: int = 2,
-        num_obj: int = 1,
```
timruhkopf, line 147: documentation

smac/optimizer/multi_objective/aggregation_strategy.py
```diff
@@ -9,22 +11,24 @@
 
 class AggregationStrategy(AbstractMultiObjectiveAlgorithm):
     """
-    An abstract class to aggregate multi-objective losses to a single objective losses, which can then be utilized
-    by the single-objective optimizer.
+    An abstract class to aggregate multi-objective losses to a single objective losses,
```
timruhkopf, line 14: [...] losses to a scalar loss.

If you do not consider the change, then at least change "single objective losses" to "... loss" i.e. singular

smac/utils/multi_objective.py
```diff
@@ -1,66 +1,41 @@
...
+    normalized_costs : list[float]
+        Normalized costs based on the bounds. If no bounds are given, ones are returned.
+        Also, if min and max bounds are the same, the value is set to 1.
     """
-    _values: np.ndarray
-    if isinstance(values, list):
-        _values = np.array(values)
-    else:
-        _values = values
+    if bounds is None:
```
timruhkopf, line 23: why does the normalize function return to kinds of things - the costs and ones (if no bounds are present) 

I would think, that if we do not have bounds then we would not normalize it, but rather return the ground truth costs instead, making this function consistent.

tests/test_utils/test_multi_objective.py
```diff
@@ -10,61 +12,34 @@
...
     def setUp(self):
-        self.bounds_1d = [(0, 1)]
-        self.bounds_2d = [(0, 1), (50, 100)]
+        self.bounds = [(0, 50), (50, 100)]
+        self.bounds_invalid = [(0, 0), (5, 5)]
 
     def test_normalize_costs(self):
-        # Normalize between 0..1 given data only
-        v = np.array([[5, 2], [10, 0]])
+        # If no bounds are passed, we get ones back
```
timruhkopf, line 19: you will have to change this here as well if norm no longer returns ones for bounds=none

timruhkopf (commented):
Normalization returns a one vector if budget = None. This may not be intuitive and desirable. 

+ we might want to make normalization part of the aggregation strategy instead (making it explicit) 

+ We could as well call the aggregationStrategy ScalarizationStrategy, implying that we want to make a scalar objective from the multi objective

smac/optimizer/multi_objective/aggregation_strategy.py
```diff
@@ -35,16 +39,18 @@ class MeanAggregationStrategy(AggregationStrategy):
...
 
-    def __call__(self, values: np.ndarray) -> float:
+    def __call__(self, values: list[float]) -> float:
         """
         Transform a multi-objective loss to a single loss.
 
         Parameters
         ----------
-            values (np.ndarray): Normalized values.
+        values : list[float]
```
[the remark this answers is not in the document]
renesass: I don't see a reason why the values should not be normalized here.

smac/optimizer/multi_objective/parego.py
```diff
@@ -8,32 +10,32 @@
...
+        theta = self.rng.rand(len(values))
 
         # Normalize st all theta values sum up to 1
         theta = theta / (np.sum(theta) + 1e-10)
 
         # Weight the values
         theta_f = theta * values
-
-        return np.max(theta_f, axis=1) + self.rho * np.sum(theta_f, axis=1)
+        return np.max(theta_f, axis=0) + self.rho * np.sum(theta_f, axis=0)
```
mfeurer, line 41: I'm somewhat surprised to see that the summation over an axis changes without a respective test changing. Does this mean that there is no test for ParEGO?

smac/utils/multi_objective.py
```diff
@@ -1,66 +1,40 @@
...
-        if denominator < 1e-10:
-            # Return ones
-            normalized_values.append(np.ones_like(data))
+        if q < 1e-10:
+            cost = 1.0
         else:
-            numerator = data - min_value
-            normalized_values.append(numerator / denominator)
+            cost = p / q
+        costs += [cost]
```
mfeurer, line 38: Why not `costs.append()`?

PR #874, opened 2022-07-11
Add: workflow to update pre-commit versions

eddiebergman:
Add a workflow that runs nightly to check for pre-commit version updates and creates a PR if there are any.


renesass:
Maybe we do it weekly instead?

eddiebergman:
Sure sure

PR #885, opened 2022-10-18
Fix parego

KEggensperger:
Fixes random weight (re-)generalization. 

Before the weights were generated for each call to build_matrix, now we only re-generate them for every iteration.


renesass:
Thank you all for your effort on this PR! It's looking good so I'm gonna merge it and release 2.0.0a2 asap.

renesass (approved):

PR #887, opened 2022-10-25
corrected log prob for tophat prior

ekalosak:


renesass:
@hvarfner Can you approve this change?

hvarfner:
Sure. Want me to look or just approve?

renesass:
It's basically one line only so having a quick look might help. Because in your original implementation, you just returned 0.

hvarfner:
Now, I actually looked.  This should not be related to my work, since it is the standard GP over functions and not the PiBO work with priors over the optimum. I still don't mind having a look at it, but I am not the one who implemented this (assuming that there hasn't been a mixup somewhere).

renesass:
Oh, I'm sorry, I thought the priors came together with PiBO. But @mfeurer did actually implement this long time ago.

mfeurer:
Hey, just had a look. This should be fine as the result is still a constant.

PR #892, opened 2022-12-07
Version 2.0.0b1

renesass:
This version contains primarily the new intensifier changes.

# Documentation
https://automl.github.io/SMAC3/development/

# Required for release candidate

- [x] Intensifier rework:
    - [x] Support for multi-threating
    - [x] Consider trials in runhistory
- [x] SH/HB
    - [x] Easier implementation
    - [x] Consider trials in runhistory
- [x] Fix multi-tasking on macs
- [x] Update changelog
- [x] Update docs
- [x] Specify some docstrings (inconsistent API in runhistory)
- [x] Reach 87% test coverage
- [x] Benchmarking

# Optional
- [ ] Integrate BOinG + TurBO
- [ ] Integrate HydraFacade
- [ ] Bug when evaluating only few trials?
- [ ] Run profiler
- [ ] Use latest budget if highest budget not yet evaluated (or give option)
- [ ] Option to restart crashed ones

# Intensifier

- [x] Decide which trial to suggest next if we are waiting for all busy workers (try new config instead of new instance, seed pair of already evaluated config) 
- [x] Should directly operate on the `RunHistory` (no `process_results` method anymore)
- [x] When should the model be trained? static approach: after n evaluated configs
- [x] If user provided instances, seeds in the RunHistory: Those are preferred to be evaluated first
- [x] `get_next_config`: Generator, yields a config. Count how many configs yielded inside the generator.
- [x] Intensifier: Generator, yield trial.
- [x] Get rid off t_intensify (does not work in multi-processing) --  what was the idea behind it? Make sure that the computation time of the surrogate model and the intensifier are balanced?

# Discussions

* Continue runs.
    * Many components are depending on states. Idea: Save/load states s.t. the optimization can pick-off where it stopped.
    * Have the option to re-run crashed ones.
    * HyperparameterFacade starts a local search when continuing a run. However, the limit already was exceeded. 
    * HB/SH does not work because an incumbent is given and it's stuck in the first stage.
* Ask-and-Tell Interface
    * Call asks multiple times before calling tell
    * Find a way to incorporate the trials from the user when only using ``tell``. Partially works for ``intensifier`` already.
    * It does not make sense to tell SMAC trials in advance when using SH. Reason: It's heavily depending on a budget+instance combination and even if the user provides it, SMAC have to wait till the other trials have been finished too.
* Constraints instead of imputation?
* Facade: Build the facade automatically based on the scenario inputs (like if budgets are defined use successivehalving e.g.) -> AutoFacade? -> Log which components are used
* Random_design.check should not be in acquition but in SMBO main loop, since it obleviates the necessity of computing epm and acquition to begin with?
* Ask and intensification inversion. Currently, the ask method is passed to the intensifier. But ideally, the ask method defines what configuration to be evaluated exactly on what problem instance. This implies the inverse relation: the intensifier should be called from within ask?
* Intensification is one way of defining a fidelity (as number of problem instances to evaluate on) but it shouldn’t be at the heart of SMAC, since nowadays the multiple dataset optimization is no longer as prominent.
* 10000 challengers: Never touch the surrogate model again? This usually happens due to the intensification percentage being at 0.5, the model fitting taking quite long, and the functions taking no time. If these three things don't apply, this is indeed an issue
* _collect_data in smbo.py: Training only on the highest fidelity or mixed fidelity? -> Docs
* Problem with _get_x_best and instances: Only the config with the lowest cost is used?!
* Tools to visualize things.


renesass:
Okay guys, I just made my last pass and fixed some last minute issues. Please have a look and confirm ASAP.

sarah-krebs (approved):

tests/test_model/test_rf.py
```diff
@@ -215,44 +215,44 @@ def test_with_ordinal():
...
+#     rs = np.random.RandomState(1)
+
+#     cv = sklearn.model_selection.KFold(shuffle=True, random_state=rs, n_splits=2)
+
+#     for do_log in [False, True]:
+#         if do_log:
+#             targets = np.log(y)
+#             model = RandomForest(
+#                 configspace=_get_cs(X.shape[1]),
+#                 ratio_features=1.0,
```
benjamc, line 231: Commented out bc should be tested in the RF repo? (Same for GP?)

benjamc (approved):

tests/test_model/test_rf.py
renesass: it's because Boston is not supported anymore... changing the dataset means changing all of the values

PR #900, opened 2023-01-11
Fix 531 correctly create integers in initial design

mfeurer:
This PR fixes a bug in which integer hyperparameters would not be created correctly in the initial design. This went unnoticed, except in the case where it was part of a condition and the function deactivate_inactive_hyperparameters was called.

This PR fixes this bug by mapping the random float that was sampled by, for example and LHD or Sobol sequence, to the float representation of an integer by calling _transform and _inverse_transform of the integer hyperparameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - Updated documentation based on new implementation.
 - Added benchmark to compare different versions.
 
+## Bugfixes
+- Correct handling of integer hyperparameters in the initial design.
+
 
 # 2.0.0a2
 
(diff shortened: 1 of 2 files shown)


sarah-krebs (approved):
Thanks for fixing this issue!

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 - Updated documentation based on new implementation.
 - Added benchmark to compare different versions.
 
+## Bugfixes
+- Correct handling of integer hyperparameters in the initial design.
```
benjamc, line 11: Add issue number

benjamc (approved):

CHANGELOG.md
sarah-krebs: Sure, done.

PR #901, opened 2023-01-19
Local search fix

jeroenrook:
- Search stops when maximum number of steps is reached.
- Search stops when it is stuck in a local optimum.
- Size of sampled neighbours to compute the acquisition value over is reset to `self.vectorization_min_obtain` instead of a fixed size of 2.
- Extended logging output message.

### DIFF
--- a/smac/optimizer/acquisition/maximizer.py
+++ b/smac/optimizer/acquisition/maximizer.py
@@ -189,7 +189,7 @@ class LocalSearch(AcquisitionFunctionMaximizer):
         vectorization_max_obtain: int = 64,
     ):
         super().__init__(acquisition_function, config_space, rng)
-        self.max_steps = max_steps
+        self.max_steps = np.inf if max_steps is None else max_steps
         self.n_steps_plateau_walk = n_steps_plateau_walk
         self.vectorization_min_obtain = vectorization_min_obtain
         self.vectorization_max_obtain = vectorization_max_obtain
@@ -455,17 +455,25 @@ class LocalSearch(AcquisitionFunctionMaximizer):
                 if not active[i]:
                     continue
                 if obtain_n[i] == 0 or improved[i]:
-                    obtain_n[i] = 2
+                    obtain_n[i] = self.vectorization_min_obtain
                 else:
                     obtain_n[i] = obtain_n[i] * 2
                     obtain_n[i] = min(obtain_n[i], self.vectorization_max_obtain)
                 if new_neighborhood[i]:
                     if not improved[i] and n_no_plateau_walk[i] < self.n_steps_plateau_walk:
-                        if len(neighbors_w_equal_acq[i]) != 0:
+                        if len(neighbors_w_equal_acq[i]) > 0:
                             candidates[i] = neighbors_w_equal_acq[i][0]
                             neighbors_w_equal_acq[i] = []
+                        else:
+                            # Local optima: All neighbours are worse
+                            self.logger.debug(
+                                f"Local search {i}: Stop search since there are no better points in the neighborhood.")
+                            active[i] = False
+                            continue
+
                         n_no_plateau_walk[i] += 1
-                    if n_no_plateau_walk[i] >= self.n_steps_plateau_walk:
+
+                    if n_no_plateau_walk[i] >= self.n_steps_plateau_walk or local_search_steps[i] >= self.max_steps:
                         active[i] = False
                         continue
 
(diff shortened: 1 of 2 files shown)


benjamc (approved):

smac/optimizer/acquisition/maximizer.py
```diff
@@ -455,17 +455,25 @@ def _do_search(
...
                 if obtain_n[i] == 0 or improved[i]:
-                    obtain_n[i] = 2
+                    obtain_n[i] = self.vectorization_min_obtain
                 else:
                     obtain_n[i] = obtain_n[i] * 2
                     obtain_n[i] = min(obtain_n[i], self.vectorization_max_obtain)
                 if new_neighborhood[i]:
                     if not improved[i] and n_no_plateau_walk[i] < self.n_steps_plateau_walk:
-                        if len(neighbors_w_equal_acq[i]) != 0:
+                        if len(neighbors_w_equal_acq[i]) > 0:
```
dengdifan, line 464: I am not sure if we really need this. Assuming that we have lots of hyperparameters, as each time we only sample 8 neighbors, it is quite likely that all these 8 neighbors are all worse than the current one while the acquisition function values can be further reduced with the other neighbors

smac/optimizer/acquisition/maximizer.py
```diff
@@ -455,17 +455,25 @@ def _do_search(
...
                             candidates[i] = neighbors_w_equal_acq[i][0]
                             neighbors_w_equal_acq[i] = []
+                        else:
+                            # Local optima: All neighbours are worse
```
dengdifan, line 468: I am not so sure about that. Assuming that we have lots of hyperparameters, as each time we only sample 8 neighbors, it is quite likely that all these 8 neighbors are all worse than the current one while the acquisition function values can be further reduced with the other neighbors

jeroenrook: I follow your line of reasoning, but I would argue that this concern should be addressed in the neighbourhood generator (by, for example, increasing the numbers of neighbours to sample) and not in the local search procedure. In fact, we primarily sample more than one neighbourhood value to increase the chance of getting an improvement, right? Or are there other motivations for doing this? 

Furthermore, before the local search was vectorized in (v0.11.0), the search did also stop when it reached a local optimum.

mlindauer:
@mfeurer could you please also review that PR. If I remember correctly, you implemented the plateau walks for the local search.

mfeurer:
Will have a look tomorrow. I remember this helping quite a bit, but don't have data on this any more.

smac/optimizer/acquisition/maximizer.py
mfeurer: So this basically disables the plateau walking, right?

One would assume that increasing the number of neighbors would do the same thing as disabling plateau walking. While I agree, the issue would be that looking at more neighbors at the same time, this would increase the overhead of SMAC because these extra points would need to go through the random forest and acquisition function.

Maybe @frank-hutter can provide further intuition as he proposed this functionality back then.

mfeurer (commented):
This makes sense to me except for the drop of the plateau walking. Also, why does this go to 1.4.1 and not to 2.0?

smac/optimizer/acquisition/maximizer.py
jeroenrook: Plateau walk still occurs as long as there are neighbours with the same acquisition value as the current candidate. Only when all neighbours have an acquisition value that is worse than the value of the current candidate, I propose to consider it as a local optimum and stop the search. 
In the previous implementation the neighbourhood of the same current candidate will be sampled (which also include categorical and ordinal parameters where the neighbourhood sampling is deterministic (right?)). This also results in an increased overhead of (guaranteed futile) computations per plateau walk. Furthermore, considering you actually are in a local optimum, is it worthwhile to do 10 plateau walks (assuming the default value and no previous plateau walks in the search trajectory) before stopping the search?

mfeurer: Interesting points. There are two things I extract from this:
1. Plateau walk must be disabled for candidates that return all neighbors. I'm afraid that this needs to be done on the level of the config space?
2. We need to benchmark this to see whether it improves performance or not. @alexandertornede you mentioned that you are working on a benchmarking setup in some other context?

benjamc:
As this is a performance improvement and not a bugfix we will not add it to 1.4. The improvement went into 2.0.

PR #902, opened 2023-01-19
Random forest speed up v1.4

jeroenrook:
Random forest with instances predicts the marginalized costs by using a C++ implementation in `pyrfr`.

Please let me know if the `try .. except` construction should be kept.


benjamc:
As this is a performance improvement instead of a bugfix we will not integrate in 1.4, but it is already integrated in 2.0. :)

Issue #921, opened 2023-02-18
`InvalidParameterError` when using MCMCGaussianProcess and scikit-learn >= 1.2.X


jamesdolezal:
#### Description
Optimizers which use the `MCMCGaussianProcess` will encounter the following error with scikit-learn 1.2.X:

```
sklearn.utils._param_validation.InvalidParameterError: The 'n_restarts_optimizer' parameter of GaussianProcessRegressor must be an int in the range [0, inf). Got -1 instead.
```

The issue is in [`MCMCGaussianProcess._get_gp()`](https://github.com/automl/SMAC3/blob/bf32ca81b8ce06510605159950ec804d90d32c82/smac/epm/gaussian_process/mcmc.py#L279), where the the argument `n_restarts_optimizer` is set to -1, with the comment "Do not use scikit-learn's optimization routine."

In scikit-learn version 1.2.1, the `n_restarts_optimizer` requires a value between 0 and infinity (see [their docs](https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.GaussianProcessRegressor.html)). In versions <1.2 this was not enforced, but parameters are [now being validated in 1.2.X](https://github.com/scikit-learn/scikit-learn/blob/8c9c1f27b7e21201cfffb118934999025fd50cca/sklearn/gaussian_process/_gpr.py#L222), which is causing the error to be raised.

The solution would be to change:

```
n_restarts_optimizer=-1,  # Do not use scikit-learn's optimization routine
``` 

to 

```
n_restarts_optimizer=0,  # Do not use scikit-learn's optimization routine
``` 

in `smac.epm.mcmc:284`.

eddiebergman:
Hi @jamesdolezal,

We are currently in the process of delivering a new SMAC that's still in development and being trialled in a few projects before being released. In this latest version of SMAC, [this parameter is correctly set!](https://github.com/automl/SMAC3/blob/e5f9875341cc2a2cac69373a31d0352d91581adf/smac/model/gaussian_process/mcmc_gaussian_process.py#L299).

My guess is you're using the `1.4` version of SMAC ([the release history is here](https://pypi.org/project/smac/#history)) which is the current default of `pip install smac`.

In the meantime, you could do a [monkey patch](https://stackoverflow.com/questions/5626193/what-is-monkey-patching) along theses lines:
```python
from smac.epm.gaussian_process.mcmc import MCMCGaussianProcess

[...]

Issue #531, opened 2019-09-11
Initial design fails with GreaterThanCondition


mfeurer:
Example code to reproduce:

```
import numpy as np
import sklearn.datasets
import sklearn.metrics
from sklearn.neural_network import MLPClassifier

from ConfigSpace import (
    UniformIntegerHyperparameter,
    CategoricalHyperparameter,
    GreaterThanCondition,
    ConfigurationSpace,
    UniformFloatHyperparameter,
)

from smac.scenario.scenario import Scenario
from smac.facade.smac_hpo_facade import SMAC4HPO


cs = ConfigurationSpace()
alpha = UniformFloatHyperparameter(
    name='alpha', lower=0.000001, upper=1, default_value=0.0001, log=True,
)
lrate = UniformFloatHyperparameter(
    name='lrate', lower=0.000001, upper=1, default_value=0.001, log=True,
)
momentum = UniformFloatHyperparameter(
    name='momentum', lower=0.5, upper=1, default_value=0.9,
)
n_layers = UniformIntegerHyperparameter(
    name='n_layers', lower=1, upper=3,
)
number_neurons_1 = UniformIntegerHyperparameter(
    name='neurons_1', lower=16, upper=256, default_value=32, log=True,
)
number_neurons_2 = UniformIntegerHyperparameter(
    name='neurons_2', lower=16, upper=256, default_value=32, log=True,
)
number_neurons_3 = UniformIntegerHyperparameter(
    name='neurons_3', lower=16, upper=256, default_value=32, log=True,
)
lrate_schedule = CategoricalHyperparameter(
    name='lrate_schedule', choices=['constant', 'invscaling', 'adaptive'],
)
cs.add_hyperparameters([
    alpha, lrate, momentum, number_neurons_1, number_neurons_2,
    number_neurons_3, n_layers, lrate_schedule,
])
neurons_2_condition = GreaterThanCondition(number_neurons_2, n_layers, 1)
neurons_3_condition = GreaterThanCondition(number_neurons_3, n_layers, 2)
cs.add_conditions([neurons_2_condition, neurons_3_condition])


# Load data
X, y = sklearn.datasets.load_breast_cancer(return_X_y=True)
X_train, X_valid, y_train, y_valid = \
        sklearn.model_selection.train_test_split(X, y, random_state=1)


def eval_mlp(config, seed):
    n_layers = config['n_layers']
    layer_sizes = []
    for i in range(1, n_layers + 1):
[...]

PR #922, opened 2023-02-20
V1.4.1

eddiebergman:
# 1.4.1

## Bugfixes
- Correct handling of integer hyperparameters in the initial design (#531).
- Pass correct parameter to GaussianProcessRegressor (#921).


jamesdolezal:
Just wanted to give a quick heads up that there's one more `n_restarts_optimizer` that would need to be changed from -1 to 0 in [epm.gaussian_process.gp:143](https://github.com/automl/SMAC3/blob/bf32ca81b8ce06510605159950ec804d90d32c82/smac/epm/gaussian_process/gp.py#L143). But after that change, I no longer get the `InvalidParameterErrror` reported in #921 and everything else works for our use case.

renesass:
Since you guys update 1.4, it also makes sense to fix the installation guide (see #923). :)

smac/epm/gaussian_process/gpytorch.py
```diff
@@ -104,6 +108,12 @@ def __init__(
             The number of components to keep when using PCA to reduce dimensionality of instance features. Requires to
             set n_feats (> pca_dims).
         """
+        if _scipy_objective_and_grad is None:
```
dengdifan, line 111: Since we have a this function under our own codebase, I will simply delete this check

dengdifan:
Documentation is still failing. When I run `make doc` on my own machine, it showed that:
`
(api/smac.optimizer.acquisition.maximizer: line    1) broken    https://www.programiz.com/python-programming/closure - HTTPSConnectionPool(host='www.programiz.com', port=443): Max retries exceeded with url: /python-programming/closure (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f5472ad2610>: Failed to establish a new connection: 
`
@eddiebergman Do you have any idea why this would happen?

renesass:
Seems like the link checker can not connect to the website. Just remove it from the docs?

PR #929, opened 2023-02-27
Ac bench

benjamc:
- AC Benchmark added
- Updated plotting
- Updated benchmark results (run on my machine)


benchmark/src/datasets/ac_branin.py
```diff
@@ -0,0 +1,31 @@
...
+
+import numpy as np
+import pandas as pd
+from sklearn import datasets
+from src.datasets.dataset import Dataset
+
+
+class ACBraninDataset(Dataset):
+    def __init__(self) -> None:
+        self._data = pd.read_csv("/home/benjamin/Dokumente/code/tmp/SMAC3/benchmark/src/datasets/ac_branin_instances_features.csv", skipinitialspace=True)
```
timruhkopf, line 13: you probably will not want to load this from your computer :)

timruhkopf (changes requested):
The Trajectory_walltime plot misses 1.4.0 for some cases

you load the file from disk not repo

benchmark/src/datasets/ac_branin.py
benjamc: done

PR #931, opened 2023-02-28
Feature/add version to make publish

sarah-krebs:
- Update version numbers to 2.0.0b2
- Use bash in makefile (else `read` command does not work)
- Fix make publish to use version set in makefile (else the command to install the test-release does not work for pre-versions)
- Fix make publish to use package-name


eddiebergman:
Seems like there's a lot of places to update versions around the repo, it could be good to collect these locations and maybe print them out in `make publish` or somewhere. I'm not sure if I had to go in and update the version numbers, I'd know where to find them all

sarah-krebs:
Good Point. I'll add that to `make publish`.

helegraf (approved):
LGTM

PR #951, opened 2023-03-14
fix(runhistory): set id of the sampled config

aron-barm:
### Context
[Auto-sklearn](https://github.com/automl/auto-sklearn) is heavily dependent on SMAC. I'm currently trying to bump up our dependency to the latest version of SMAC.

### Problem Description
Auto-sklearn used to rely on the fact that SMAC set the [config_id attribute](https://github.com/automl/ConfigSpace/blob/1d93d98e4b0b73a0d3a6799f05f11f4c1a4bb5f9/ConfigSpace/configuration_space.pyx#L1651) of a sampled Configuration object before the `submit_trial()` method was called on the function runner object in the [main optimization loop](https://github.com/automl/SMAC3/blob/731854e8effe72d76cb07636458e783143f32dc5/smac/main/smbo.py#L272-L319).

This functionality got removed since 2.0, and now the config_ids of the sampled Configurations are all None.

### Question
Is there any good reason why you got rid of it?

### Solution
If there isn't, then this brief PR tries to resolve it in a simple way. The config_id gets set to the ID of the Configuration object already present in the run_history.

### DIFF
--- a/smac/main/smbo.py
+++ b/smac/main/smbo.py
@@ -278,9 +278,6 @@ class SMBO:
                 # Sample next trial from the intensification
                 trial_info = self.ask()
 
-                # Set the ID of the sampled config object
-                trial_info.config.config_id = self.runhistory.config_ids[trial_info.config]
-
                 # We submit the trial to the runner
                 # In multi-worker mode, SMAC waits till a new worker is available here
                 self._runner.submit_trial(trial_info=trial_info)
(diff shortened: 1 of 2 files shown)


renesass:
I think it makes more sense to add it after this line here
https://github.com/automl/SMAC3/blob/731854e8effe72d76cb07636458e783143f32dc5/smac/runhistory/runhistory.py#L222

aron-barm:
Thanks for the quick feedback René!
I moved the logic of setting the `config_id` to runhistory.py, right below the line you linked, but outside of the `if` clause.

PR #963, opened 2023-04-11
ci: set dependabot to development

eddiebergman:
Just makes dependabot target development with it's changes. This should prevent divergences between `main` and `development`. I set myself as the reviewer and assignee of these. Please let me know if anyone else want's to be notified and specifically assigned to these.


benjamc (approved):

benjamc:
thanks for taking care of this:)

alexandertornede (approved):
Thanks, Eddie!

PR #970, opened 2023-04-14
fix(logging): Prevent automatic logging setup at init

### LINKED ISSUE #948: [Feature] Enable method to disable custom logging of SMAC.
The parameter [`logging_level`](https://github.com/automl/SMAC3/blob/731854e8effe72d76cb07636458e783143f32dc5/smac/facade/abstract_facade.py#L109) is always used with the function [`setup_logging`](https://github.com/automl/SMAC3/blob/731854e8effe72d76cb07636458e783143f32dc5/smac/facade/abstract_facade.py#L113), no matter what is passed in. This prevents users from just using the logging that they have set up themselves. I'm not sure exactly why this is done but my guess is because the `logger` get's passed around for some multi-processing reasons.

I made a PR in #947

eddiebergman:
Prevent SMAC doing automatic logging setup at import, in general it's very bad practice for libraries to automatically perform computation at import time as simply `import smac` performs unknown changes to a users environemnt.

* Closes #948

### DIFF
--- a/smac/__init__.py
+++ b/smac/__init__.py
@@ -23,10 +23,6 @@ version = "2.0.0"
 
 
 try:
-    from smac.utils.logging import setup_logging
-
-    setup_logging(0)
-
     from smac.callback import Callback
     from smac.facade import (
         AlgorithmConfigurationFacade,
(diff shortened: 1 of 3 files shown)


alexandertornede (approved):
LGTM

PR #981, opened 2023-04-19
Fix validate method of SMAC to use the seed given in the scenario

helegraf:
Previously, if no seed was given, the seed 0 was used.


alexandertornede (approved):
LGTM

PR #988, opened 2023-04-25
fix broken references

dengdifan:
fix the broken links in doc


dengdifan:
Seems like that only the links that end with "pdf" work (checks also failed on researchgate)

alexandertornede:
Then let's stay with your link. Thanks for fixing this!

alexandertornede (approved):
LGTM

PR #993, opened 2023-05-02
Add a workaround to be able to pass a dataset via dask.scatter

nabenabe0928:
# Background

When we run SMAC using a tabular benchmark or a surrogate benchmark, SMAC incurs a significant slowdown due to the (de-)serialization problem.

The problem here is that when a given target function refers to such a benchmark, the target function must be serialized to be passed to each process and be deserialized when each process uses it.
However, (de-)serialization is unnecessary because both tabular and surrogate benchmarks are static.
For this reason, we would like to use dask.scatter to be able to keep such a benchmark in memory.

This PR gives the feature to do so.

**NOTE**: This PR will be merged to the `main` branch, but we probably do not want to do it, so please take care of the branch to be merged.


smac/facade/abstract_facade.py
```diff
@@ -275,7 +275,7 @@ def tell(self, info: TrialInfo, value: TrialValue, save: bool = True) -> None:
         """
         return self._optimizer.tell(info, value, save=save)
 
-    def optimize(self) -> Configuration | list[Configuration]:
+    def optimize(self, shared_data=None) -> Configuration | list[Configuration]:
```
eddiebergman, line 278: Can we add a proper type to it? I'm not sure if it's of a special type or just a `dict[str, Any]`? A brief comment about the argument would also be helpful :)

```suggestion
    def optimize(self, shared_data: dict[str, Any] | None = None) -> Configuration | list[Configuration]:
```

smac/main/smbo.py
```diff
@@ -244,7 +245,7 @@ def update_acquisition_function(self, acquisition_function: AbstractAcquisitionF
             assert config_selector._acquisition_maximizer is not None
             config_selector._acquisition_maximizer.acquisition_function = acquisition_function
 
-    def optimize(self) -> Configuration | list[Configuration]:
+    def optimize(self, shared_data=None) -> Configuration | list[Configuration]:
```
eddiebergman, line 248: Same here regarding the typing and docstring

smac/runner/abstract_runner.py
```diff
@@ -76,7 +76,7 @@ def __init__(
                 assert isinstance(scenario.crash_cost, float)
                 self._crash_cost = [scenario.crash_cost for _ in range(self._n_objectives)]
 
-    def run_wrapper(self, trial_info: TrialInfo) -> tuple[TrialInfo, TrialValue]:
+    def run_wrapper(self, trial_info: TrialInfo, **shared_dict) -> tuple[TrialInfo, TrialValue]:
```
eddiebergman, line 79:
```suggestion
    def run_wrapper(self, trial_info: TrialInfo, **shared_dict: Any) -> tuple[TrialInfo, TrialValue]:
```

smac/runner/dask_runner.py
```diff
@@ -154,8 +154,9 @@ def run(
         instance: str | None = None,
         budget: float | None = None,
         seed: int | None = None,
+        **shared_dict,
```
eddiebergman, line 157:
```suggestion
        **shared_dict: Any,
```

smac/runner/target_function_runner.py
```diff
@@ -93,6 +93,7 @@ def run(
         instance: str | None = None,
         budget: float | None = None,
         seed: int | None = None,
+        **shared_dict,
```
eddiebergman, line 96:
```suggestion
        **shared_dict: Any,
```

eddiebergman (commented):
Seems good but likely linters will shout at you for some pointed out reasons.

This more generally seems like a method to pass your own kwargs to a target function, i.e.

```python
def my_target_function(..., *, my_extra_object: str):
    ...
    
smac.optimize(
    my_target_function,
    shared_data={"my_extra_object": "hello_world"}
)
```

The problem here is that I could easily use the functionality for this purpose, without knowing that under the hood, this is issuing a `dask.scatter`. I don't really know the implications of `dask.scatter` so that may not be an issue though.

Seeing as this is a relatively "advanced" feature, it might make sense to make the argument more obscure on purpose? For example,
`smac.optimize(dask_scatter={"x": y}). It's immediately clear it's dask specific and what functionality it invokes. For anyone unaware, it would encourage them to read the docstrings and force them to understand the choice they are making.

One example is it would also help people understand the case where it's not a `DaskParallelRunner` and why suddenly their `my_extra_object = "hello"` never makes it to their target function?

All that being said, perhaps there should be some extra arguments along the lines of what you have done but are just passed through `client.submit` and not necessarily using `dask.scatter`

smac/facade/abstract_facade.py
```diff
@@ -275,18 +275,29 @@ def tell(self, info: TrialInfo, value: TrialValue, save: bool = True) -> None:
...
         return self._optimizer.tell(info, value, save=save)
 
-    def optimize(self) -> Configuration | list[Configuration]:
+    def optimize(self, *, data_to_scatter: dict[str, Any] | None = None) -> Configuration | list[Configuration]:
         """
         Optimizes the configuration of the algorithm.
 
+        Parameters
+        ----------
+        dask_data_to_scatter: dict[str, Any] | None
```
eddiebergman, line 284: Just a correct to the param name :)

smac/main/smbo.py
```diff
@@ -269,6 +280,17 @@ def optimize(self) -> Configuration | list[Configuration]:
...
+        if isinstance(self._runner, DaskParallelRunner) and dask_data_to_scatter is not None:
+            dask_data_to_scatter = dict(
+                data_to_scatter=self._runner._client.scatter(data_to_scatter, broadcast=True)
+            )
+        elif dask_data_to_scatter is not None:
+            raise ValueError(
+                "dask_data_to_scatter is valid only for DaskParallelRunner, "
+                f"but {dask_data_to_scatter} was provided for {self._runner.__class__.__name__}"
+            )
+
```
eddiebergman, line 293: The error you're getting from the checks are from here, this suggestion should fix it.

```python
elif any(dask_data_to_scatter):
    ...
```

eddiebergman (changes requested):

eddiebergman (approved):

eddiebergman:
Added @benjamc as a reviewer since I do not have god-mode to accept changes and 2 reviewers are required. I thought you might be the most familiar? If not, please recommend someone else who knows dask :))

benjamc:
Thank you for the PR!
Could you also update the `CHANGELOG.md`?

smac/facade/abstract_facade.py
```diff
@@ -386,7 +397,7 @@ def get_config_selector(
         scenario: Scenario,
         *,
         retrain_after: int = 8,
-        retries: int = 16,
+        retries: int = 1000,
```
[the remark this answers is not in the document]
nabenabe0928: Thanks for raising this point, I do not know why it sneaked into this PR, but this is a heuristic for my own experiment, so I will get it back.

nabenabe0928:
> Thank you for the PR! Could you also update the `CHANGELOG.md`?

Done!

smac/facade/abstract_facade.py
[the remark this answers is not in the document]
benjamc: Thanks!

benjamc (approved):

benjamc (approved):

Issue #992, opened 2023-04-28
Basic SMAC usage with n_workers > 2 crashes at run time


AndrewJamesTurner:
#### Description

Basic SMAC usage with n_workers > 2 crashes at run time.

#### Steps/Code to Reproduce

Using a fresh Python 3.10.6 virtual environment with smac==2.0.0 installed. 

```
import time
from ConfigSpace import Configuration, ConfigurationSpace
from smac import HyperparameterOptimizationFacade, Scenario


def train(config: Configuration, seed):

    time.sleep(1)

    return 10


def main():

    config_space = ConfigurationSpace({"a": (1, 10), "b": (1, 10)})

    scenario = Scenario(
        config_space,
        deterministic=False,
        walltime_limit=60,
        n_workers=2
    )

    smac = HyperparameterOptimizationFacade(
        scenario,
        train
    )

    smac.optimize()


if __name__ == "__main__":
    main()
```

If n_workers is set to 1, it runs fine.

If the time.sleep(1) is removed from train it works fine

#### Expected Results

Everything to work fine

#### Actual Results

KeyError in internal get_config_id method. 

```
[INFO][proxy.py:71] To route to workers diagnostics web server please install jupyter-server-proxy: python -m pip install jupyter-server-proxy
[INFO][scheduler.py:1624] State start
[INFO][scheduler.py:3871]   Scheduler at:     tcp://127.0.0.1:43079
[INFO][scheduler.py:3886]   dashboard at:  http://127.0.0.1:8787/status
[INFO][nanny.py:367]         Start Nanny at: 'tcp://127.0.0.1:34521'
[INFO][nanny.py:367]         Start Nanny at: 'tcp://127.0.0.1:34481'
[DEBUG][selector_events.py:54] Using selector: EpollSelector
[DEBUG][selector_events.py:54] Using selector: EpollSelector
[INFO][worker.py:1434]       Start worker at:      tcp://127.0.0.1:37641
[INFO][worker.py:1435]          Listening to:      tcp://127.0.0.1:37641
[INFO][worker.py:1438]           Worker name:                          1
[INFO][worker.py:1440]          dashboard at:            127.0.0.1:44043
[INFO][worker.py:1441] Waiting to connect to:      tcp://127.0.0.1:43079
[INFO][worker.py:1442] -------------------------------------------------
[INFO][worker.py:1443]               Threads:                          1
[INFO][worker.py:1445]                Memory:                  15.63 GiB
[...]

PR #997, opened 2023-05-04
check if config in rh when storing state

dengdifan:
Check if the config is stored under runhistory before querying it, should solve https://github.com/automl/SMAC3/issues/992


benjamc (changes requested):

smac/intensifier/intensifier.py
```diff
@@ -81,7 +81,11 @@ def uses_instances(self) -> bool:  # noqa: D102
 
     def get_state(self) -> dict[str, Any]:  # noqa: D102
         return {
-            "queue": [(self.runhistory.get_config_id(config), n) for config, n in self._queue],
+            "queue": [
+                (self.runhistory.get_config_id(config), n)
+                for config, n in self._queue
+                if self.runhistory.has_config(config)
```
renesass, line 87: So now you basically just ignore the config?

dengdifan: They are the configs that are put in the queue but not executed yet. Currently, our RH cannot (and do not necessarily) record the potential future configurations, I think it is fine to simply drop them.

PR #1002, opened 2023-05-16
Feature/dask client

benjamc:
Add info message that `scenario.n_workers` is ignored when directly passing a dask client. We need to set the number of workers in the `dask_client` instead.


eddiebergman (changes requested):
Might be better to make it a warning rather than `INFO` as `INFO` is not generally displayed in `logging.basicConfig()`. Main reason being is a user could wrongly expect SMAC do be doing something other than it is.

benjamc:
Good idea with the warning

eddiebergman (approved):

PR #1007, opened 2023-05-16
Feature/GitHub actions pr draft

timruhkopf:
Try to make dist-test, conda-test and source-test in pytest.yml conditional on the draft status of the pull request, to avoid execution of those test. 

On success, we will need to verify, that the state change to "ready for review" will trigger the tests once again.

### DIFF
--- a/examples/1_basics/2_svm_cv.py
+++ b/examples/1_basics/2_svm_cv.py
@@ -72,7 +72,9 @@ if __name__ == "__main__":
     # Next, we create an object, holding general information about the run
     scenario = Scenario(
         classifier.configspace,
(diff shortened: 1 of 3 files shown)


examples/1_basics/2_svm_cv.py
benjamc, line 1: What's up with this example?

timruhkopf: that's an accidental commit from the configspace default stuff. i don't intent to merge this pr eventually, it is merely here to demonstrate that the GH-Actions are not executed & the checks are executed on change from draft to propper pr (which failed btw)

PR #1018, opened 2023-05-26
Raise an error for invalid scatter data

nabenabe0928:
When `data_to_scatter={}`, it causes an error in Dask, so I fixed this error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 2.0.2
+
+## Improvements
+- Add an error when we get an empty dict data_to_scatter so that we can avoid an internal error caused in Dask precautiously
+
 # 2.0.1
 
 ## Improvements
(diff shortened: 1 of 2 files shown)


alexandertornede:
Thanks for this PR, @nabenabe0928! Overall, it looks good to me. Could you please add this to the CHANGELOG.md as well? :)  

Let's see whether @benjamc has another opinion as she is the expert on the DASK runner.

benjamc (changes requested):
Looks good, can you update the changelog as Alex suggested?:)

nabenabe0928:
@benjamc @alexandertornede 
Thanks for the review:)
I updated the CHANGELOG.md!

benjamc (approved):

benjamc:
Thank you, @nabenabe0928! 🌞

Issue #1020, opened 2023-05-30
Setting trial_walltime_limit when using a slurm cluster


FlorianPommerening:
When using both a `trial_walltime_limit` and also parallelizing work on a cluster, I ran into a problem where the trials do not seem to run at all.

#### Description
The code below optimizes a single parameter `x` over a parabola. In every trial, I also write a log file to leave some trace of the trial running. The workers and the scenario are allowed to run for 10 minutes. There are 10 workers to handle 400 trials, which should be plenty, considering that each trial only writes a file, computes a simple arithmetic expression and returns.

If I now add a `trial_walltime_limit` to the scenario, the trials do not write to the files anymore and the incumbent is never updated. The value of the limit doesn't matter, I set it to 1000 (higher than the limit of the whole optimization to make this clear but lower values also cause the error).

#### Steps/Code to Reproduce
I used this code to run 
https://gist.github.com/FlorianPommerening/6a5a529a97c4b80d9a2ea5fff7f4a4d4

#### Expected Results
I would expect a `log.<x>` file to be created for each value `<x>` that is used in a trial. The incumbent should be updated several times to eventually reach a value close to 0.5.

#### Actual Results

No `log.<x>` files are created, the incumbent only updates once at the beginning (because there is no incumbent yet).

The main log reads:
```
[WARNING][abstract_facade.py:192] Provided `dask_client`. Ignore `scenario.n_workers`, directly set `n_workers` in `dask_client`.
[INFO][abstract_initial_design.py:147] Using 0 initial design configurations and 1 additional configurations.
[INFO][abstract_intensifier.py:305] Using only one seed for deterministic scenario.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][smbo.py:319] Finished 0 trials.
[INFO][abstract_intensifier.py:515] Added config 467e1c as new incumbent because there are no incumbents yet.
[INFO][smbo.py:319] Finished 50 trials.
[INFO][smbo.py:319] Finished 100 trials.
[...]

PR #1023, opened 2023-06-01
Raise error if resource limitation and parallelization is requested

benjamc:
- This is not a fix sadly
- Related to #1020 
- Dask cannot pickle jobs using processes themselves (this is what I suspect why it does not work)
- Throw error if both resource limitation and parallization is requested

### DIFF
--- a/docs/advanced_usage/9_parallelism.rst
+++ b/docs/advanced_usage/9_parallelism.rst
@@ -21,6 +21,39 @@ SMAC supports multiple workers natively via Dask. Just specify ``n_workers`` in 
     When using multiple workers, SMAC is not reproducible anymore.
 
 
+..warning::
+
+    You cannot use resource limitation (pynisher, via the `scenario` arguments `trail_walltime_limit` and `trial_memory_limit`).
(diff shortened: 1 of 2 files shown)


alexandertornede (approved):
LGTM

docs/advanced_usage/9_parallelism.rst
```diff
@@ -21,6 +21,39 @@ SMAC supports multiple workers natively via Dask. Just specify ``n_workers`` in
     When using multiple workers, SMAC is not reproducible anymore.
 
 
+..warning::
+
```
sarah-segel, line 25: Seems like two black spaces are missing (before and after warning).

sarah-segel:
LGTM, except for that one comment.

eddiebergman:
Just a note, I've had no issues running pynisher inside of a subprocess running on a dask worker. This is likely not the issue

docs/advanced_usage/9_parallelism.rst
benjamc: You are right, fixed it

PR #1043, opened 2023-06-15
fix(configspace): shows all keys in print

### LINKED ISSUE #1041: [BUG] [QUESTION] How to setup AlgorithmConfigurationFacade for multiple algorithms and hyperparameters?
Dear **Authors**, we hope you are doing well.

#### Description
We are exploring the use of SMAC for the optimisation of multiple algorithms and their hyperparameters. After successfully experimenting with the HyperparameterOptimizationFacade for a single algorithm, we are now looking to extend this to multiple algorithms.

We believe that we should use the AlgorithmConfigurationFacade for this purpose as suggested in the SMAC3 paper (SMAC4AC now AlgorithmConfigurationFacade). However, we found that the documentation for this feature is sparse, leaving us uncertain about the correct approach. We kindly request some guidance on using the AlgorithmConfigurationFacade correctly. To give back to the community, and you may be busy on a plethora of sides, we could also plan to provide examples for enhancing the documentation in a pull request over the summer.

One specific question we have is regarding the setup of the configspace for the AlgorithmConfigurationFacade. We have searched through the documentation and test cases but could not find relevant examples. Any exemples around?

As a final step in our experimentations, we plan not to only optimise algorithms and hyperparmeters, but actual custom Scikit-learn pipeline candidate for a specific use case. We would like to confirm whether our intended use of AlgorithmConfigurationFacade aligns with this goal or walk towards the right direction? If so, how does the configuration of the candidates can be made based on a given search space (probably something with the condition operations ConfigSpace allows users to do?) ?

#### Steps/Code to Reproduce
We have not encountered a problem with any specific code; this is a general usage question.

#### Expected Results
We are hoping for guidance on how to use the AlgorithmConfigurationFacade for tuning multiple algorithms and their hyperparameters. Specifically, we would appreciate an example or a detailed explanation of how to correctly set up the configspace for this purpose.

#### Actual Results
Currently, we are unable to proceed due to the limited understanding and lack of examples in the documentation.

#### Versions
We are using SMAC with Poetry on a MacBook Pro m2 (2022) running on OSX Ventura stable.

We thank you very much in advance for your time!
Cheers

eddiebergman:
Previously this would only show keys present in the challenger and not in the incumbent. Also by using string `[key]` indexing instead of `.get(key)`, there could be a `KeyError` if the `key` was not present in the `Configuration`.

* Closes #1041

### DIFF
--- a/smac/utils/configspace.py
+++ b/smac/utils/configspace.py
@@ -174,8 +174,8 @@ def print_config_changes(
 
     lines = []
     for k in sorted(all_keys):
-        inc_k = incumbent.get(k)
-        cha_k = challenger.get(k)
+        inc_k = incumbent.get(k, "-inactive-")
+        cha_k = challenger.get(k, "-inactive-")
         lines.append(f"--- {k}: {inc_k} -> {cha_k}" + " (unchanged)" if inc_k == cha_k else "")
 
     msg = "\n".join(lines)
--- a/smac/utils/configspace.py
+++ b/smac/utils/configspace.py
@@ -169,7 +169,9 @@ def print_config_changes(
     if incumbent is None or challenger is None:
         return
 
-    all_keys = set(incumbent.keys().union(challenger.keys()))
+    inc_keys = set(incumbent.keys())
+    challenger_keys = set(challenger.keys())
+    all_keys = inc_keys.union(challenger_keys)
 
     params = sorted([(param, incumbent.get(param), challenger.get(param)) for param in all_keys])
     for key, inc_val, challenger_val in params:
--- a/smac/utils/configspace.py
+++ b/smac/utils/configspace.py
@@ -176,9 +176,7 @@ def print_config_changes(
     for k in sorted(all_keys):
         inc_k = incumbent.get(k)
         cha_k = challenger.get(k)
-        lines.append(
-            f"--- {k}: {inc_k} -> {cha_k}" + " (unchanged)" if inc_k == cha_k else ""
-        )
+        lines.append(f"--- {k}: {inc_k} -> {cha_k}" + " (unchanged)" if inc_k == cha_k else "")
 
     msg = "\n".join(lines)
     logger.debug(msg)
--- a/smac/utils/configspace.py
+++ b/smac/utils/configspace.py
@@ -169,12 +169,16 @@ def print_config_changes(
     if incumbent is None or challenger is None:
         return
 
-    params = sorted([(param, incumbent[param], challenger[param]) for param in challenger.keys()])
-    for param in params:
-        if param[1] != param[2]:
-            logger.info("--- %s: %r -> %r" % param)
-        else:
-            logger.debug("--- %s Remains unchanged: %r", param[0], param[1])
+    all_keys = set(incumbent.keys().union(challenger.keys()))
+
+    params = sorted(
+        [(param, incumbent.get(param), challenger.get(param)) for param in all_keys]
+    )
+    for (key, inc_val, challenger_val) in params:
+        msg = f"--- {key}: {inc_val} -> {challenger_val}"
+        if inc_val == challenger_val:
+            msg += " (unchanged)"
+        logger.debug(msg)
 
 
 # def check_subspace_points(
--- a/smac/utils/configspace.py
+++ b/smac/utils/configspace.py
@@ -171,10 +171,8 @@ def print_config_changes(
 
     all_keys = set(incumbent.keys().union(challenger.keys()))
 
-    params = sorted(
-        [(param, incumbent.get(param), challenger.get(param)) for param in all_keys]
-    )
-    for (key, inc_val, challenger_val) in params:
+    params = sorted([(param, incumbent.get(param), challenger.get(param)) for param in all_keys])
+    for key, inc_val, challenger_val in params:
         msg = f"--- {key}: {inc_val} -> {challenger_val}"
         if inc_val == challenger_val:
             msg += " (unchanged)"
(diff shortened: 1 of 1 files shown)


smac/utils/configspace.py
```diff
@@ -169,12 +169,16 @@ def print_config_changes(
     if incumbent is None or challenger is None:
         return
 
-    params = sorted([(param, incumbent[param], challenger[param]) for param in challenger.keys()])
-    for param in params:
-        if param[1] != param[2]:
-            logger.info("--- %s: %r -> %r" % param)
-        else:
-            logger.debug("--- %s Remains unchanged: %r", param[0], param[1])
+    all_keys = set(incumbent.keys().union(challenger.keys()))
```
simonprovost: In reference to #1041, I would recommend a more efficient way for obtaining the unique keys from `incumbent` and `challenger`.  Instead of calling `.keys()` explicitly twice, the first set on both dictionaries using the sequential union, will automatically provide the unique keys:

```python
all_keys = set(incumbent).union(challenger)
```

In this expression, both `set(incumbent)` and `challenger` in the `union` method will use the keys of the respective dictionaries. This way, we avoid redundant calls to `.keys()` and make the code more readable. 

Moreover, while Python may optimize away the unnecessary `.keys()` calls (i.e, most probably in the compiler's component), removing them now can potentially improve efficiency and certainly clarifies the intention of the code 👌

smac/utils/configspace.py
```diff
@@ -169,12 +169,16 @@ def print_config_changes(
...
+
+    params = sorted(
+        [(param, incumbent.get(param), challenger.get(param)) for param in all_keys]
+    )
+    for (key, inc_val, challenger_val) in params:
+        msg = f"--- {key}: {inc_val} -> {challenger_val}"
+        if inc_val == challenger_val:
+            msg += " (unchanged)"
+        logger.debug(msg)
```
simonprovost: In reference to #1041, I propose simplifying the generation of debug messages and the procedure of the entire loop. The current implementation generates a list of parameter tuples using a list comprehension (essentially a loop) and then iterates over this list using a distinct for-in loop. Instead, these operations can be consolidated into a single loop that explicitly iterates over the sorted keys:

```python
    for key in sorted(all_keys):
        incumbent_val = incumbent.get(key)
        challenger_val = challenger.get(key)
        logger.debug(f"--- {key}: {incumbent_val} -> {challenger_val}"
                     f"{' (unchanged)' if incumbent_val == challenger_val else ''}")
```

This change simplifies the code by removing the need for the intermediate `params` list (i.e, variable, and list comprehension), and potentially improves efficiency by reducing the number of loops (yet they were not nested so not a big deal anyway). Nonetheless, it also enhances readability by generating the complete debug message string within the logger.debug() call, using a ternary expression to optionally append ' (unchanged)' rather than concatenation.

Your thoughts?

simonprovost (changes requested):
Regarding #1041, I made a few suggestions to provide a function that is likely simpler, simpler to comprehend, and with fewer redundant operations. Let me know if I am in any event wrong :) Cheers!

Happy to provide some help for your lab's project! This is exciting.

smac/utils/configspace.py
eddiebergman: Yup might be a bit faster, I'll implement it in, thank you :)

smac/utils/configspace.py
eddiebergman: Yup this should be faster, good suggestion :) I was just trying to not deviate too much from the original code.

I think in practice, it would be faster to build up the entire string and debug message at once, rather than logging each line. Given the `logger.debug` requires a syscall, doing as little of these as possible in a parallel set up makes sense

simonprovost:
> @simonprovost Thanks for the suggestions, I wasn't really thinking too much about optimization here, considering it's relatively small impact. I think the biggest speedup would probably be from calling `logger.debug` once with the generated string :)
> 
> I tried to maintain some readability as well, the ternary works but with the long variable names it becomes quite noisy.

Cheers! Indeed, this is not the most crucial aspect to optimise for SMAC. I believe I simply seized the opportunity to contribute. More likely in the following months ☺️

Cheers for your promptness with all that. have a great weekend!

eddiebergman:
> > @simonprovost Thanks for the suggestions, I wasn't really thinking too much about optimization here, considering it's relatively small impact. I think the biggest speedup would probably be from calling `logger.debug` once with the generated string :)
> > I tried to maintain some readability as well, the ternary works but with the long variable names it becomes quite noisy.
> 
> Cheers! Indeed, this is not the most crucial aspect to optimise for SMAC. I believe I simply seized the opportunity to contribute. More likely in the following months relaxed
> 
> Cheers for your promptness with all that. have a great weekend!

Thanks for engaging :) It's nice to get input that isn't a vague issue now and then

smac/utils/configspace.py
```diff
@@ -169,12 +169,17 @@ def print_config_changes(
...
-        else:
-            logger.debug("--- %s Remains unchanged: %r", param[0], param[1])
+    inc_keys = set(incumbent.keys())
+    all_keys = inc_keys.union(challenger.keys())
+
+    lines = []
+    for k in sorted(all_keys):
+        inc_k = incumbent.get(k)
+        cha_k = challenger.get(k)
+        lines.append(f"--- {k}: {inc_k} -> {cha_k}" + " (unchanged)" if inc_k == cha_k else "")
```
dengdifan, line 179: if any key becomes `None`, can we rephrase that to something like "inactivated" and "activated"?

eddiebergman: Makes sense!

dengdifan (approved):
Thanks for the PR. Looks good to me!

alexandertornede (approved):
LGTM, feel free to merge @eddiebergman !

PR #1045, opened 2023-06-15
Fix: target runner with partial func

eddiebergman:
When encoding the smac instance to disk, part of that process is to use the `__code__` attribute of a function to essentially *hash* it to disk, allowing SMAC to detect if the function has changed between invocations.

However when using a `partial`, these do not have a `__code__` attribute, moreover, they can have extra variables attached to them, as per the reason you would use a `partial`.

This PR fixes this and makes sure that even the arguments used in the `partial` are considered when using `overwrite=False`, and SMAC attempts to restart from it's last state on disk.


eddiebergman:
Anticipating failing doc build, fixed in #1043

PR #1047, opened 2023-06-20
fixed bug for conditional params

hstojic:
hello,

I found a bug in the print statement which prevents the whole loop from running (I'm using `MultiFidelityFacade`), when there are conditional hyperparameters in the challenger, but not in the incumbent, then the line I have changed would raise a key error - using `get` method instead simply gives None for nonexistent keys, which to me seems like an OK solution for this kind of print statement

cheers

### DIFF
--- a/smac/utils/configspace.py
+++ b/smac/utils/configspace.py
@@ -169,7 +169,7 @@ def print_config_changes(
     if incumbent is None or challenger is None:
         return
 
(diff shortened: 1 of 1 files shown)


eddiebergman:
Hi @hstojic, thanks for taking the time to fix this but it's already been accounted for in #1043 and waiting for merging :) Will close but we do appreciate the time!

hstojic:
great to hear that you noticed it already

PR #1051, opened 2023-06-26
Fix: Deprecation warnings configspace `0.7.1`

eddiebergman:
A while ago ConfigSpace (`~ 0.5.0`), the `Configuration` and `ConfigurationSpace` classes were made to behave like [`Mapping`](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes), i.e. they act like a dictionary with `["lookup"]`, `len()`, `values()`, `keys()`, `items()` and `.get(key, default)` plus whatever other non-mutating dictionary semantics there are.

In `0.7.1`, I changed two core files of ConfigSpace from `.pyx` to `.py` to play better with editors. As part of this I also deprecated the following methods which are basically Java style "getters" for what is already provided by following the Mapping interface.

### Notes
While updating the various locations of these methods, I found a few places that rely on traversal-order which are no order invariant. It might be wise to convert these to use the hyperparamter names and use dicts, relying on sort order of the names when an array is needed.

### Deprecations
```python
# configuration_space.py
class ConfigurationSpace(Mapping[str, HyperParameter]):


    def get_hyperparameter(self, name: str) -> Hyperparameter:
        warnings.warn(
            "Prefer `space[name]` over `get_hyperparameter`",
            DeprecationWarning,
            stacklevel=2,
        )
        return self[name]

    def get_hyperparameters(self) -> list[Hyperparameter]:
        warnings.warn(
            "Prefer using `list(space.values())` over `get_hyperparameters`",
            DeprecationWarning,
            stacklevel=2,
        )
        return list(self._hyperparameters.values())

    def get_hyperparameters_dict(self) -> dict[str, Hyperparameter]:
        warnings.warn(
            "Prefer using `dict(space)` over `get_hyperparameters_dict`",
            DeprecationWarning,
            stacklevel=2,
        )
        return self._hyperparameters.copy()

    def get_hyperparameter_names(self) -> list[str]:
        warnings.warn(
            "Prefer using `list(space.keys())` over `get_hyperparameter_names`",
            DeprecationWarning,
            stacklevel=2,
        )
        return list(self._hyperparameters.keys())


class Configuration(Mapping[str, Any]):

    def get_dictionary(self) -> dict[str, Any]:
        warnings.warn(
            "`Configuration` act's like a dictionary."
            " Please use `dict(config)` instead of `get_dictionary`"
            " if you explicitly need a `dict`",
            DeprecationWarning,
            stacklevel=2,
        )
        return dict(self)

### DIFF
--- a/smac/utils/configspace.py
+++ b/smac/utils/configspace.py
@@ -52,11 +52,22 @@ def get_types(
     -------
     The bounds for the instance features are *not* added in this function.
     """
+    # NOTE (eddiebergman): It would be better return a dict[str, tuple[int, float, float]]
+    #   where the str is the hyperparameters name and the dictionary is in sorted order by
+    #   the hyperparameters name.
+    #
+    # We are currently relying on the order of the hyperparameters of the space always
+    #   being the same. Any future extension which might incorporate changing/shrinking spaces
+    #   or even just changes to ConfigSpace can break this subtly.
+    #
+    # To counter-act this, I can implement this in if wanted but needs someone who's aware
+    #   of it's use in the kernels to verify that this will be okay.
+
     # Extract types vector for rf from config space and the bounds
-    types = [0] * len(configspace.get_hyperparameters())
+    types = [0] * len(configspace)
     bounds = [(np.nan, np.nan)] * len(types)
 
-    for i, param in enumerate(configspace.get_hyperparameters()):
+    for i, param in enumerate(configspace.values()):
         parents = configspace.get_parents_of(param.name)
         if len(parents) == 0:
             can_be_inactive = False
--- a/benchmark/src/models/svm.py
+++ b/benchmark/src/models/svm.py
@@ -43,7 +43,7 @@ class SVMModel(Model):
         """Creates a SVM based on a configuration and evaluates it on the
         iris-dataset using cross-validation."""
         assert self.dataset is not None
-        config_dict = config.get_dictionary()
+        config_dict = dict(config)
         if "gamma" in config:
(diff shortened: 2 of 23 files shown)

### CODE REVIEW

**Comment by dengdifan** [comment]
Thanks for the PR. Could you please also update the minimal required ConfigSpace version?

File: smac/utils/configspace.py [hunk 1]
```diff
@@ -52,11 +52,22 @@ def get_types(
     -------
     The bounds for the instance features are *not* added in this function.
     """
+    # NOTE (eddiebergman): It would be better return a dict[str, tuple[int, float, float]]
+    #   where the str is the hyperparameters name and the dictionary is in sorted order by
+    #   the hyperparameters name.
+    #
+    # We are currently relying on the order of the hyperparameters of the space always
```
dengdifan, line 59: In that case, can we ensure that the orders of the hp bounds are always correct (fit the order of confispace)?

smac/utils/configspace.py
```diff
@@ -52,11 +52,22 @@ def get_types(
...
+    #   being the same. Any future extension which might incorporate changing/shrinking spaces
+    #   or even just changes to ConfigSpace can break this subtly.
+    #
+    # To counter-act this, I can implement this in if wanted but needs someone who's aware
```
dengdifan, line 63: In that case, can we ensure that the orders of the hp bounds are always correct (fit the order of confispace)?

benjamc:
Hi @eddiebergman  can you check this again? :)

benjamc:
ConfigSpace 1.0 is on the way

PR #1054, opened 2023-07-04
fix scheduler file path

BertrandWen:
Fix the path of the scheduler file if output_directory is not none.

### DIFF
--- a/smac/runner/dask_runner.py
+++ b/smac/runner/dask_runner.py
@@ -91,7 +91,7 @@ class DaskParallelRunner(AbstractRunner):
             )
 
             if self._scenario.output_directory is not None:
-                self._scheduler_file = self._scenario.output_directory / ".dask_scheduler_file"
+                self._scheduler_file = self._scenario.output_directory + '/' + ".dask_scheduler_file"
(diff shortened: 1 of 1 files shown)


eddiebergman:
Hi @BertrandWen,

I'm not sure this fixes what you mention? However it seems like `self._scenario.output_directory` should never be `None`, hence no need for the `if` statement thats already there.

Anywho, why change the `/` operator to raw string concatenation with `... + '/' + ...`.

The `/` operator is cross platform while the second one will crash on windows which uses `'\'` as file separators.

Was there a bug that occured for you and this fixed it?

BertrandWen:
Sorry, I don't read the Pull Request requirement before. I am going to make a new PR for the development branch.
I'm not sure the `if` statement here if is necessary, but in my case, the `self._scenario.output_directory` and `".dask_scheduler_file"` is in `str` type and don't support the `/` operator, so I fix it.

And considering the cross-platform scenarios, I'm going to use `Path(self._scenario.output_directory, ".dask_scheduler_file)` in the new commit. It works perfectly in my test on OSX and Windows.

PR #1055, opened 2023-07-05
fix dask_scheduler_file path

BertrandWen:
The `self._scenario.output_directory` and `".dask_scheduler_file"` is in the `str` type and don't support the `/` operator, which will raise TypeError: unsupported operand type(s) for /: 'str' and 'str'.


dengdifan:
Thanks for the PR and sorry for the late response, it looks good to me!

PR #1057, opened 2023-07-20
Handle configspace as dictionary in mlp example

sarah-segel:
- Handle configspace as dictionary in mlp and parego example
- Adapt sgd loss to newest scikit-learn version ("log" not accepted anymore")

### DIFF
--- a/examples/2_multi_fidelity/1_mlp_epochs.py
+++ b/examples/2_multi_fidelity/1_mlp_epochs.py
@@ -84,18 +84,18 @@ class MLP:
         # For deactivated parameters (by virtue of the conditions),
         # the configuration stores None-values.
         # This is not accepted by the MLP, so we replace them with placeholder values.
-        lr = config["learning_rate"] if config["learning_rate"] else "constant"
-        lr_init = config["learning_rate_init"] if config["learning_rate_init"] else 0.001
-        batch_size = config["batch_size"] if config["batch_size"] else 200
+        lr = config.get("learning_rate") if config.get("learning_rate") else "constant"
+        lr_init = config.get("learning_rate_init") if config.get("learning_rate_init") else 0.001
+        batch_size = config.get("batch_size") if config.get("batch_size") else 200
 
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore")
 
             classifier = MLPClassifier(
-                hidden_layer_sizes=[config["n_neurons"]] * config["n_layer"],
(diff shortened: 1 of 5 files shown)


examples/2_multi_fidelity/1_mlp_epochs.py
```diff
@@ -84,18 +84,18 @@ def train(self, config: Configuration, seed: int = 0, budget: int = 25) -> float
         # For deactivated parameters (by virtue of the conditions),
         # the configuration stores None-values.
         # This is not accepted by the MLP, so we replace them with placeholder values.
-        lr = config["learning_rate"] if config["learning_rate"] else "constant"
-        lr_init = config["learning_rate_init"] if config["learning_rate_init"] else 0.001
-        batch_size = config["batch_size"] if config["batch_size"] else 200
+        lr = config.get("learning_rate") if config.get("learning_rate") else "constant"
+        lr_init = config.get("learning_rate_init") if config.get("learning_rate_init") else 0.001
+        batch_size = config.get("batch_size") if config.get("batch_size") else 200
```
eddiebergman, line 89: Just an FYI, you can do
```python
lr = config.get("learning_rate", "constant")
lr_init = config.get("learning_rate_init", 0.001)
batch_size = config.get("batch_size", 200)
```

examples/2_multi_fidelity/1_mlp_epochs.py
```diff
@@ -84,18 +84,18 @@ def train(self, config: Configuration, seed: int = 0, budget: int = 25) -> float
...
 
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore")
 
             classifier = MLPClassifier(
-                hidden_layer_sizes=[config["n_neurons"]] * config["n_layer"],
-                solver=config["solver"],
+                hidden_layer_sizes=[config.get("n_neurons")] * config.get("n_layer"),
```
eddiebergman, line 95: If `n_neurons` is going to be in there, you can still do `config["n_neurons"]`. In fact it would error if either of these were to not be present. Likewise for the two lines below

eddiebergman (commented):
You should look at the definition of `Mapping.get` which applies to dicst and to `Configuration` now. It's roughly something like:

```python

def get(self, key: str, default = None):
    try:
        return self[key]
    except KeyError:
        return default
```

eddiebergman (approved):

PR #1061, opened 2023-08-04
[#1056] Add example on intensify for cross-validation.

### LINKED ISSUE #1056: SMAC for cross validation examples
We need to create an example showing that each fold of cross-validation can be considered as an instance in SMAC, similar to https://github.com/automl/SMAC3/blob/main/examples/2_multi_fidelity/2_sgd_datasets.py

mfeurer:
Closes #1056

I hope the example explains itself. There are two things that I would have loved to add, but that currently do not work (will create separate issues for them):

1. A comparison against requiring all folds evaluated (option for this is missing from the intensifier).
2. ~~Removing the warning that instance features are missing (SMAC currently gives an error when passing instance features).~~ This now works after upgrading the pyrfr package.

### DIFF
--- a/examples/4_advanced_optimizer/4_intensify_crossvalidation.py
+++ b/examples/4_advanced_optimizer/4_intensify_crossvalidation.py
@@ -6,7 +6,11 @@ An example of optimizing a simple support vector machine on the digits dataset. 
 [simple example](examples/1_basics/2_svm_cv.py), in which all cross-validation folds are executed
 at once, we use the intensification mechanism described in the original 
 [SMAC paper](https://link.springer.com/chapter/10.1007/978-3-642-25566-3_40) as also demonstrated
-by [Auto-WEKA](https://dl.acm.org/doi/10.1145/2487575.2487629). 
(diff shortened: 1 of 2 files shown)


examples/4_advanced_optimizer/4_intensify_crossvalidation.py
```diff
@@ -0,0 +1,119 @@
...
+        return cost
+
+
+if __name__ == "__main__":
+    classifier = SVM()
+
+    # Next, we create an object, holding general information about the run
+    scenario = Scenario(
+        classifier.configspace,
+        n_trials=50,  # We want to run max 50 trials (combination of config and seed)
```
dengdifan, line 79: Here `n_trials` should be the combination of config, instance, and seed. (otherwise, users might think that `n_trials` indicates n configurations if `deterministic=True`)

dengdifan:
Thanks for the PR. Overall, it looks good to me, except for a minor point. Could you please address this? Otherwise, we can merge this PR

examples/4_advanced_optimizer/4_intensify_crossvalidation.py
mfeurer: I updated the docstring and also improved the docstring of the example file.

PR #1066, opened 2023-08-10
Propagate the Scenario random seed to get_random_design

bpkroth:
The random seed specified in the Scenario, is propagated to most other places, but not the default ProbabilityRandomDesign child component.

This is part of an effort to track down some sources of non-determinism.

### DIFF
--- a/smac/facade/hyperparameter_optimization_facade.py
+++ b/smac/facade/hyperparameter_optimization_facade.py
@@ -178,7 +178,7 @@ class HyperparameterOptimizationFacade(AbstractFacade):
         probability : float, defaults to 0.2
             Probability that a configuration will be drawn at random.
         """
-        return ProbabilityRandomDesign(probability=probability)
+        return ProbabilityRandomDesign(probability=probability, seed=scenario.seed)
 
     @staticmethod
     def get_multi_objective_algorithm(  # type: ignore
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 2.0.3
+
+## Bugfixes
+
+- Propagate the Scenario random seed to `get_random_design` (#1066)
+
 # 2.0.2
 
 ## Improvements


sarah-segel:
As you are saying there is non-determinism, could you maybe provide a few more details on where/how you observed that? 
That would be great.

smac/facade/hyperparameter_optimization_facade.py
```diff
@@ -178,7 +178,7 @@ def get_random_design(  # type: ignore
         probability : float, defaults to 0.2
             Probability that a configuration will be drawn at random.
         """
-        return ProbabilityRandomDesign(probability=probability)
+        return ProbabilityRandomDesign(probability=probability, seed=scenario.seed)
```
dengdifan, line 181: I agree that this needs to be correctly passed to `ProbabilityRandomDesign`. However, since the default parameter for  `seed` is set as 0, ignoring this part might not result in any non-determinism behaviors. Could you please provide more details regarding the non-determinism things?

bpkroth:
> As you are saying there is non-determinism, could you maybe provide a few more details on where/how you observed that? That would be great.

So sorry, got caught up in other things and kept forgetting to come back to this.

Yes, I was seeing a bunch of non-determinism across either machines or versions of Python, but I couldn't nail down a good simplified test case to reproduce it.

Basically, I had some simple unit test cases that would attempt to do a very simple optimization operation but once they got out of the random sampling phase would take divergent paths.  Same seeds are used in all cases and the function itself is also deterministic, so I would expect the optimizer to produce the same set of suggestions each time, but on occasion it would produce something else.  My guess is that somewhere in the stack the random seed was not getting propagated or there's unintended shared state that is getting polluted somehow.

When I wrap up with some of my current tasks I'll take another crack at that.

smac/facade/hyperparameter_optimization_facade.py
bpkroth: https://github.com/automl/SMAC3/pull/1066#issuecomment-1749127210

sarah-segel:
@bpkroth Thank you for those insights! In case you gain any further insights on this, it would be great if you could keep us updated.

sarah-segel (approved):

PR #1067, opened 2023-08-15
Don't use mutable default argument.

bpkroth:
If this isn't respected, then additional_configs can be built up overtime when multiple optimizers are created in a single process.

This can occur when the use_default_config option is set True and/or when additional_configs are provided.

When different ConfigSpaces are provided this can result in an error due to mismatched defaults, for instance.

In other cases it can result in non-deterministic behavior due to the random ordering of the configs that get added (e.g., with different parallel pytest runs).

- [x] Cleanup other sources of this error as well elsewhere in the code.
- [x] Add pylint checks for this issue.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,3 +43,16 @@ repos:
         name: flake8 smac
         files: smac
         exclude: "scripts|tests"
+
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
+        args:
+          [
+            "-rn", # Only display messages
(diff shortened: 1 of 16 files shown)


helegraf:
Thanks for spotting and addressing this issue! We will discuss how we want to adapt the pre-commit hooks.

helegraf:
Could you please add a line to the `CHANGELOG.md` mentioning the bugfix?

bpkroth:
> Could you please add a line to the `CHANGELOG.md` mentioning the bugfix?

Hey, did that again.  Anything else?  Thanks!

benjamc:
Closed by #1097

PR #1069, opened 2023-08-16
Fix config rejection for #1068

### LINKED ISSUE #1068: Racing: Configs worse than existing incumbent are not rejected `bug`
#### Description

I wanted to use the tell-interface of the HPOFacade (which uses racing for intensification, the related class being `Intensifier`) to warmstart the optimizer with data from previous runs which were all collected on the same seed, no instances were used.

#### Steps/Code to Reproduce

https://gist.github.com/PhilippBordne/722c336dd0981859bfade198171f9c29

#### Expected Results
All configurations worse than the incumbent should be in the `_rejected_config_ids` list after using the tell interface for warmstarting. They should still be in it after `__post_init__` of `AbstractIntensifier` was run when starting the actual optimization loop. `__post_init__` will call `update_incumbents(config)` on the runhistory again and thus also update the `_rejected_config_ids` list.

#### Actual Results
The configs worse than the incumbent are in the `_rejected_config_ids` list after using the tell interface for warmstarting if they were "told" in an order that improves the cost with every new config. But they are not in the list after running `__post_init__` (They are all removed during the `update_incumbents` cycle on existing configs in `RunHistory` at the end of `__post_init__`). This can lead the subsequent optimization run to fail because the configs which are erroneously not in the `_rejected_config_ids` list will fill up the queue and make the intensifier use up all its retries on them. 
For details see the linked code to reproduce.

#### Proposed fix
The issue is caused by the current configs being _removed_ from `_rejected_config_ids` if the incumbent doesn't change during `update_incumbents(config)`, which is also the case if newly evaluated configs are worse than the existing incumbent. This can be fixed by inserting a simple check if the config is among the incumbents after `update_incumbents`.
A PR implementing this is incoming.

#### Versions

2.0.2

PhilippBordne:
Fixes #1068 to make sure newly evaluated configs are rejected in case they are worse than the existing incumbent.

### DIFF
--- a/smac/intensifier/abstract_intensifier.py
+++ b/smac/intensifier/abstract_intensifier.py
@@ -572,7 +572,15 @@ class AbstractIntensifier:
         if len(previous_incumbents) == len(new_incumbents):
             if previous_incumbents == new_incumbents:
                 # No changes in the incumbents
-                self._remove_rejected_config(config_id)
+                # need this clause to make sure set difference isn't empty
+                if config_id in new_incumbent_ids:
+                    # TODO: can this case even happen? Intuitively a config should not be among incumbents and rejected
+                    # config IDs at the same time when starting incumbent update. (add/remove rejected config methods
+                    # are only called within update_incumbents.)
+                    self._remove_rejected_config(config_id)
+                else:
+                    # config worse than incumbents and thus rejected
+                    self._add_rejected_config(config_id)
                 return
             else:
                 # In this case, we have to determine which config replaced which incumbent and reject it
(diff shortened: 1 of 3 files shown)


eddiebergman (approved):
We discussed this previously and I did a check beforehand. However should still be approved by someone from Hannover

@alexandertornede @dengdifan

dengdifan:
Thanks for the PR. Looks good to me!

smac/intensifier/abstract_intensifier.py
```diff
@@ -571,8 +571,12 @@ def update_incumbents(self, config: Configuration) -> None:
 
         if len(previous_incumbents) == len(new_incumbents):
             if previous_incumbents == new_incumbents:
-                # No changes in the incumbents
-                self._remove_rejected_config(config_id)
+                # No changes in the incumbents, we need this clause because we can't use set difference then
```
dengdifan, line 574: However, since all the configs are unique and disordered in `_rejected_config_ids`,  I cannot find a reason not to store `config_ids` as Set()

PR #1071, opened 2023-08-24
chore: log warn on condition when custom dask client is provided

SimonThormeyer:
Currently, the warning is always emitted when a custom dask client is provided.
IMO it would be better to just emit it when ` n_workers` is `> 1`.


dengdifan:
Thanks for the PR. It looks good to me. Could you please merge this into the development branch and update the changelog accordingly? Thanks!

SimonThormeyer:
@dengdifan Done

PR #1084, opened 2023-11-16
fix(RandomForest): Always cast `seed` to `int`

eddiebergman:
In cases where seeds were generated from some numpy objects, sometimes
you'd get back an `np.integer`, which causes the program to crash when
communicating with `pyrfr` through `swig`. It seems that swig doesn't
know that it's an _int-like_ and so we explicitly cast it to `int`.


benjamc (approved):
LGTM

Issue #1038, opened 2023-06-13
Make warmstarting easier


benjamc:
Provide a helper function receiving `list[Configuration]` and float costs.
Use `tell` under the hood and convert stuff into `TrialInfo` and `TrialValue`.

armandieu:
Hello @benjamc, with a group of friend from University studying optimization, we would like to collaborate on this issue if possible. 
Can you please tell me more about the details.

eddiebergman:
Heyo, just leaving this for context #950 as I've provided some sample code there in a project where I wrap SMAC to be able to do this with my own trial-like objects.

eddiebergman:
I tried this out in the Colab notebook at the fall school and can verify it's at least consistent with the `tell()` interface, i.e. use one instantiation to generate 10 trial/value pairs. Create a new optimizer instance and tell it the 10 trial/value pairs. Confirm that the 11th trial produced by both is the same.

Doesn't solve the original problem but good to confirm its correctness that it can be used to warmstart correctly.

armandieu:
Hello Dear @eddiebergman , can you please share this Colab notebook to have an idea of where to start. Thanks.

armandieu:
Hello,
For example purposes, I was thinking of a method included in the helper class similar to this. 
```python
#self smac class
#configList: List of Configuration
#costList: List of floats
def warm_starting(self, configList, costList):
  trialInfos = [SMACTrialInfo(config=config) for config in configList]
  trialValues = [SMACTrialValue(cost=cost) for cost in costList]
  for i in range(len(trialInfo)):
    self.facade.tell(info=trialInfos[i], value=trialValues[i], save=True)
```
As I am learning from the documentation, please let me know if it makes sense.

armandieu:
Hello SMAC community, as part of this approach, I have made an example of how the implementation of a warmstarting from a list of configuration and cost values will look like. 
Any comments and direction are really appreciated.

```python
from ConfigSpace import Configuration
from smac.runhistory.dataclasses import TrialValue, TrialInfo
# For the moment we assume the values 1, 2, -0.5 are points the model must know to solve it faster.
x_values = [1, 2, -0.5]
# In this case I initialize the cost for each point as specific values, but it could be the result of the
# target function evaluated in the point model.train(config)
cost_list = [3, 4, 1]
# Finally we will transform the x_values to warm start into TrialInfo objects which also needs. Here we assume x is an hyperparameter already setup in the configspace of the model
[...]

PR #1090, opened 2023-12-18
Added warmstarting method to facade

armandieu:
Hello dear community,
I am sending this approach for the pull request: #1038

I hope although small, it makes a contribution to the community.

### DIFF
--- a/smac/facade/abstract_facade.py
+++ b/smac/facade/abstract_facade.py
@@ -290,7 +290,23 @@ class AbstractFacade:
             Whether the runhistory should be saved.
         """
         return self._optimizer.tell(info, value, save=save)
+    
+    def warmstarting(self, infos, values, save: bool = True) -> None:
+        """Adds the result of a list of trials to the runhistory and updates the intensifier.
 
+        Arguments:
+            infos (List TrialInfo): A list of Trial Info which describes the trials from which to process the results.
+            values (List float): A list containing per item the relevant information regarding the execution of a trial.
+            save (bool, optional): Whether the runhistory should be saved. Defaults to True.
+        """
+        # Check length of info is same as Values
+        if(len(infos)!=len(values)):
+            raise ValueError('Not matching lenght, the lenght of Info list must match the values list')
+        trialValues = [TrialValue(cost=cost) for cost in values]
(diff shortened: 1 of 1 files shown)


smac/facade/abstract_facade.py
```diff
@@ -290,7 +290,23 @@ def tell(self, info: TrialInfo, value: TrialValue, save: bool = True) -> None:
...
+
+        Arguments:
+            infos (List TrialInfo): A list of Trial Info which describes the trials from which to process the results.
+            values (List float): A list containing per item the relevant information regarding the execution of a trial.
+            save (bool, optional): Whether the runhistory should be saved. Defaults to True.
+        """
+        # Check length of info is same as Values
+        if(len(infos)!=len(values)):
+            raise ValueError('Not matching lenght, the lenght of Info list must match the values list')
+        trialValues = [TrialValue(cost=cost) for cost in values]
```
dengdifan, line 305: [TrialValue](https://github.com/automl/SMAC3/blob/main/smac/runhistory/dataclasses.py#L94) also requires some additional arguments (such as time, status, and additional_info). Therefore, it would be nice to check the content of values and then decide how to handle it.
For instance, if `values` is a list of `dict`, then we initialize `TrialValue` with `**value` , or if `values` is a list of `TrialValue`, then we do not need to transform that to `TrialValue`.

smac/facade/abstract_facade.py
```diff
@@ -290,7 +290,23 @@ def tell(self, info: TrialInfo, value: TrialValue, save: bool = True) -> None:
...
+        # Use the tell interface to the optimizer
+        for i in range(len(infos)):
```
dengdifan, line 307: So I would recommend using a `zip` here and reinitializing `trial_Values` at each iteration.

dengdifan:
Thanks for the PR. 
Overall it looks good to me. There are some minor points to be solved. 
Additional to that, could you please switch the target branch from `main` to `development`, then update the [changelog](https://github.com/automl/SMAC3/blob/main/CHANGELOG.md) and add a test case for this function?

benjamc:
Addressed by this example #1120

PR #1102, opened 2024-04-16
test: Fix upper bounds for `test_transformer()` and no random in `test_transformer_conditional()`

eddiebergman:
When testing the [new ConfigSpace branch](https://github.com/automl/ConfigSpace/tree/typing-cython-3) to check if it would break anything in SMAC, this test failed just due to a change on randomization.

I've fixed the test to not rely on the randomization of sampling.

### DIFF
--- a/tests/test_runhistory/test_runhistory_encoder.py
+++ b/tests/test_runhistory/test_runhistory_encoder.py
@@ -14,6 +14,7 @@ from smac.runhistory.encoder.encoder import RunHistoryEncoder
 from smac.runner.abstract_runner import StatusType
 
 from ConfigSpace import Configuration
+from ConfigSpace.hyperparameters import CategoricalHyperparameter
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


tests/test_runhistory/test_runhistory_encoder.py
```diff
@@ -39,70 +41,127 @@ def test_transform(runhistory, make_scenario, configspace_small, configs):
...
+        else 1.0
+        for hp in configspace_small.get_hyperparameters()
+    }
+    # Need to ensure they match the order in the Configuration vectorized form
+    sorted_by_indices = sorted(
+        upper_bounds.items(),
+        key=lambda x: configspace_small._hyperparameter_idx[x[0]],
+    )
+    upper = np.array([upper_bound for _, upper_bound in sorted_by_indices])
+    lower = 0.0
```
benjamc, line 67: I guess the lower bound is always 0 no matter the type of HP?

benjamc (changes requested):
some import is missing

eddiebergman:
Whoops sorry, should be corrected now.
The CodeFactor check failed because there's a #TODO in there. This TODO only makes sense once the new ConfigSpace is in use. Should I remove it?

benjamc:
Yeah just leave the comment/make it obvious that this is relevant for ConfigSpace 1.0?

tests/test_runhistory/test_runhistory_encoder.py
eddiebergman: Yes, but theoretically someone can implement their own HP where that's not the case. Hence the change in the above `#TODO`

benjamc (approved):

Issue #987, opened 2023-04-25
Create an example on warmstarting SMAC with a fixed meta-dataset


alexandertornede:
We should have an example on how to warmstart SMAC from a fixed meta-dataset.

benjamc:
Closed by #1120

PR #1120, opened 2024-06-25
[doc] #1038, #987 added warm starting using ask and tell

timruhkopf:
Added warm-starting documentation example (didn't warrant its own advanced_usage example)
It is super simple and would only add unnecessary code.


benjamc (changes requested):
We need to show how we reduce the initial design. Because normally when people want to warmstart, they do not want an initial design, often, it is replaced.

PR #1150, opened 2024-10-09
Fix differential evolution

hadarshavit:
Fix for differential evolution acquisition function optimisation.
Fixing a bug that it didn't work when the configspace has categorical/integer hyperparameters.


benjamc:
Can you merge the current development branch into your branch?

benjamc (commented):
In principle LGTM. Could you describe a bit more why we need this fix?
Why did it work before?

Oh and can you also update the `CHANGELOG.md`?
Thanks!

hadarshavit:
When using categorical hyperparameters differential evolution returned an exception. 
For example, modifying the SVM example (examples/1_basics/2_svm_cv.py) to:
```
    initial_design = HyperparameterOptimizationFacade.get_initial_design(scenario, n_configs=5)
    acq_opt = DifferentialEvolution(classifier.configspace)

    # Now we use SMAC to find the best hyperparameters
    smac = HyperparameterOptimizationFacade(
        scenario,
        classifier.train,
        initial_design=initial_design,
        acquisition_maximizer=acq_opt,
        overwrite=True,  # If the run exists, we overwrite it; alternatively, we can continue from last state
    )
```

Results in:
`ValueError: Got unexpected float value while trying to transform a vector representation into a value in ['linear' 'poly' 'rbf' 'sigmoid'].Expected integers but got [0.17368738] (dtype: float64)
`

The fix simply transforms float vectors (like the ones scipy's differential evolution outputs) to valid configs.
The test (tests/test_acquisition/test_maximizers.py) has a configspace of 2 float hyperparameters, and therefore passes.

benjamc:
Can you add this as a test in `test_maximizers.py`? Then I'll merge
Thank you!

benjamc (approved):

PR #1173, opened 2024-11-26
Adding in CPU time to wherever time is defined as variable "cpu_time"

thijssnelleman:
Changed the TrialValue to hold CPU time and have SMBO use this value to calculate its budgets

- Same behaviour as the original 'time' variable, except in its calculation tries to estimate CPU time with process_time instead.
- Added in to TrialValue object, RunHistory and SMBO so it can be recorded and correctly calculate the remaining CPU budget

Pytest are passing (on my machine)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Improvements
 - Add logger information on handling of stopIteration error (#960)
 - Replace deprecated ConfigSpace methods (#1139)
+- Separated Wallclock time measurements from CPU time measurements and storing them under new 'cpu_time' variable (#????)
 
 ## Dependencies
 - Allow numpy >= 2.x (#1146)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 ## Improvements
 - Add logger information on handling of stopIteration error (#960)
 - Replace deprecated ConfigSpace methods (#1139)
-- Separated Wallclock time measurements from CPU time measurements and storing them under new 'cpu_time' variable (#????)
+- Separated Wallclock time measurements from CPU time measurements and storing them under new 'cpu_time' variable (#1173)
 
 ## Dependencies
 - Allow numpy >= 2.x (#1146)
--- a/smac/main/smbo.py
+++ b/smac/main/smbo.py
@@ -81,6 +81,7 @@ class SMBO:
         # Stats variables
         self._start_time: float | None = None
         self._used_target_function_walltime = 0.0
+        self._used_target_function_cputime = 0.0
 
         # Set walltime used method for intensifier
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## Improvements
 - Add logger information on handling of stopIteration error (#960)
 - Replace deprecated ConfigSpace methods (#1139)
+- Separated Wallclock time measurements from CPU time measurements and storing them under new 'cpu_time' variable (#????)
```
benjamc, line 13: update PR number

smac/main/smbo.py
```diff
@@ -137,6 +138,11 @@ def used_target_function_walltime(self) -> float:
         """Returns how much walltime the target function spend so far."""
         return self._used_target_function_walltime
 
+    @property
+    def used_target_function_cputime(self) -> float:
+        """Returns how much cputime the target function spend so far."""
```
benjamc, line 143: describe that CPU time generally means time spent on the hardware and not just CPU (whether its CPU, TPU, GPU or whatever)

smac/runhistory/dataclasses.py
```diff
@@ -98,6 +98,7 @@ class TrialValue:
     ----------
     cost : float | list[float]
     time : float, defaults to 0.0
+    cpu_time : float, defaults to 0.0
```
benjamc, line 101: describe CPU time (what it means and why/ in what way it differs from time)

smac/runhistory/runhistory.py
```diff
@@ -191,6 +192,8 @@ def add(
             Cost of the evaluated trial. Might be a list in case of multi-objective.
         time : float
             How much time was needed to evaluate the trial.
+        cpu_time : float
```
benjamc, line 195: Same here, add a little more info

smac/runhistory/runhistory.py
```diff
@@ -848,6 +856,7 @@ def load(self, filename: str | Path, configspace: ConfigurationSpace) -> None:
         self._n_id = len(self._config_ids)
 
         # Important to use add method to use all data structure correctly
+        # NOTE: These hardcoded indices can easily lead to trouble
```
benjamc, line 859: agreed

benjamc (changes requested):
is there any way to test the new cpu return? if yes, please add a little test

otherwise, codewise the PR LGTM, it only needs a bit more documentation.
You can basically copy/paste the description for cpu_time everywhere.

smac/runhistory/runhistory.py
thijssnelleman: https://github.com/automl/SMAC3/issues/1174

thijssnelleman:
Added in CPU time tests in target function runner tests

benjamc (approved):

Issue #1174, opened 2024-11-29
Replace hard coded indices in RunHistory


thijssnelleman:
#### Description
In the RunHistory class, when loading data from a file, quite some data on an entry level is being extracted by using hard coded indices of a list. This is very error prone and can lead to older runhistories no longer to work with a new release. 

#### Possible Solution
Store the information as a dictionary instead so we can do key/value matching for each stored attribute.

#### Versions
2.2.0

LukasFehring:
Are you referencing this https://github.com/automl/SMAC3/issues/1085 or something else?

benjamc:
Closed by #1180

PR #1175, opened 2024-12-02
1085 runhistory documentation

### LINKED ISSUE #1085: Unclarity in the Documentation
#### Description
When working with `smac` there are two issues I struggled with:
1. Some classes do not contain Class Docstrings, while others do. For example `AlgorithmConfiguraitonFacade` and `BlackBoxFacade` and `RunHistoryLogScaledEncoder`, do not have them. While some of them are basic in principle, this makes it significantly harder to understand what is the difference between different components
2. I did not find an explanation for the default optimization output (Maybe I missed it?). In order to understand the `runhistory` file, I had to find out the exact place of logging and check what the `dict` that is written as `json' looks like. Especially the data part 
```json
"data": [
    [
      1,
      null,
      662124363,
      null,
      0.06746031746031744,
      0.05632376670837402,
      1,
      1700216610.3126414,
      1700216610.3698032,
      {}
    ],
```
is not self-explanatory. Additionally the `data` is a list and `configs` is a dictionary with keys, increasing the difficulty of matching a configuration to the evaluated cost, since the configuration with key `3` has to be accessed as `data[2]`.
3. In the description of abstract components [AbstractIntensifier](https://automl.github.io/SMAC3/main/api/smac.intensifier.html), a link to [Advanced Usage](https://automl.github.io/SMAC3/main/advanced_usage/1_components.html) would help.
4.  [Advanced Usage](https://automl.github.io/SMAC3/main/advanced_usage/1_components.html) should be navigated by a menu (also makes links for 3 more useful).

PS: The issue template implies that one should give a label to issues, but I one does not have the necessary privileges.

benjamc:
Output of runhistory not well documented, related: #1174 

(First step to update docs, then code bc changing code has downstream consequences of libs like DeepCAVE)

Closes #1085

### DIFF
--- a/docs/advanced_usage/8_logging.md
+++ b/docs/advanced_usage/8_logging.md
@@ -87,7 +87,7 @@ The optimization process is portrayed in `optimization.json` with the following 
 ``` 
 ### runhistory.json
 The runhistory.json in split into four parts. `stats`, `data`, `configs`, and `config_origins`.
-`stats` contains overall broad stats on the different evaluated configruations:
+`stats` contains overall broad stats on the different evaluated configurations:
 ```json
   "stats": {
     "submitted": 73,
@@ -116,7 +116,7 @@ The runhistory.json in split into four parts. `stats`, `data`, `configs`, and `c
   ]
 ```
 
-`configs` contians a human readable dictionary of configurations. Note that this list starts by one, therefore the 0 entry from above is mapped to 1.
+`configs` contains a human readable dictionary of configurations. Note that this list starts by one, therefore the 0 entry from above is mapped to 1.
 ```json
   "configs": {
     "1": {
--- a/docs/advanced_usage/8_logging.md
+++ b/docs/advanced_usage/8_logging.md
@@ -116,7 +116,7 @@ The runhistory.json in split into four parts. `stats`, `data`, `configs`, and `c
   ]
 ```
 
-`configs` contains a human readable dictionary of configurations. Note that this list starts by one, therefore the 0 entry from above is mapped to 1.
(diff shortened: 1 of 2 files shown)


docs/advanced_usage/8_logging.md
```diff
@@ -27,6 +27,113 @@ The table shows you the specific levels:
...
+
+`configs` contains a human readable dictionary of configurations. Note that this list starts by one, therefore the 0 entry from above is mapped to 1.
+```json
+  "configs": {
+    "1": {
+      "x": -2.3312147893012
+    },
+```
+
+Lastly, `config_origins` contains the origin of a configuration.
```
benjamc, line 127: add, that origins can be the from the initial design or from different acquisition function maximization techniques

docs/advanced_usage/8_logging.md
```diff
@@ -27,6 +27,113 @@ The table shows you the specific levels:
...
+      1,                            # status
+      1733133181.2144582,           # start_time
+      1733133181.21695,             # end_time
+      {}                            # additional_info
+    ],
+    ...
+  ]
+```
+
+`configs` contains a human readable dictionary of configurations. Note that this list starts by one, therefore the 0 entry from above is mapped to 1.
```
benjamc, line 119: rephrase: The config ids are 1-based. (just leave out the last half-sentence, it confused me a bit)

LukasFehring: The problem is that in the data part, the values are actually not starting at config_index 1, but at config_index 0

benjamc: Ah okay thanks!

benjamc: weird that it is logged like this 😅

benjamc:
LGTM

Issue #1170, opened 2024-11-23
Avoid workers waiting for training of surrogate model to finish


bbudescu:
### Motivation

Here's how a cpu load graph looks like for a multi-objective optimization session using the multi-fidelity facade that ran for about 46 hours an a 64 core machine (no hyperthreading, `n_workers=64`), finishing almost 20k trials on a bit over 16k distinct configurations (two rungs).

![Screenshot 2024-11-23 at 11-59-52 Instances EC2 eu-west-1](https://github.com/user-attachments/assets/2d93f0a5-eae8-46ee-b1a3-5dca91357487)

One can see that cpu utilizations decreases to less than 50% after the first 12 hours. It then drops to under 40% after another 10 hours (by this time, 12.6k trials were finished in total).

### Previous Discussion
I thought that another cause for this degradation in performance might be Hyperband, and thought that using ASHA (https://github.com/automl/SMAC3/issues/1169) instead would help eliminate that hypothesis, however, after @eddiebergman's  https://github.com/automl/SMAC3/issues/1169#issuecomment-2493662911, I understand the problem is caused by workers waiting to get another suggestion from the surrogate model

### Potential solution
- train the random forest in a different thread
- replace a version of the RF with a newly trained one only when the training is done (similar to double buffering)
- workers should always get configs using the currently available RF, even if a new RF is training
- optionally: use an occupancy threshold, e.g., 90%, and allow worker threads to wait for training to finish only if the percentage of workers that are idle waiting for the new RF version is below 10%.
- optionally: add gpu support to accelerate training of Random Forest
- optionally: perhaps add the option to decrease the number of workers running the target function by 1 once the RF trainer occupies a CPU core for more than 50% of time

dengdifan:
Hi @bbudescu, 
Thanks for raising this issue and providing potential solutions. As far as I can see, the potential solutions that you proposed are mostly related to the RF.  We are also planning to replace the current RF packages: https://github.com/automl/SMAC3/issues/1116

However, the new RF model might still rely on the existing packages (e.g., sklearn). Since we are only a small team maintaining SMAC, we might not have enough manpower to customize a new random forest package. However, if you have any good ideas on how to implement this efficiently (in Python) or you would like to create a new PR regarding the RF replacement stuff. We are happy to help you (on how to integrate this into SMAC).

bbudescu:
Hi @dengdifan,

[...]

Issue #1169, opened 2024-11-22
[Feature Request] ASHA


bbudescu:
Asynchronous Successive Halving Algorithm (ASHA) is, as far as I can understand, a tweaked version of the Hyperband multi-fidelity intenstifier (aka [ early stopping | pruning | scheduler ] algorithm) that allows to make better use of the available computing power in the context of many cpu cores (improves cpu occupancy).

Here are a bunch of references:
- nice [explanation](https://d2l.ai/chapter_hyperparameter-optimization/sh-async.html) in a book that I found helpful for myself to understand how it works
- authors' 2018 [blog post](https://blog.ml.cmu.edu/2018/12/12/massively-parallel-hyperparameter-optimization/) introducing the method
- 2018 [paper](https://openreview.net/revisions?id=S1Y7OOlRZ) (rejected by ICLR); arxiv preprint [here](https://arxiv.org/abs/1810.05934)
- another docs entry offering a pretty nice [explanation](https://docs.determined.ai/latest/model-dev-guide/hyperparameter/search-methods/hp-adaptive-asha.html)

I think [here](https://github.com/liamcli/darts_asha) is the original repo in which the authors published the reference implementation, but a bunch of other optimizers implement it, too nowadays (e.g., [Oríon](https://orion.readthedocs.io/en/stable/user/algorithms.html#asha), [optuna](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.pruners.SuccessiveHalvingPruner.html)). I've personally used `ray[tune]`'s implementation before (docs [overview](https://docs.ray.io/en/latest/tune/api/schedulers.html#asha-tune-schedulers-ashascheduler), [api reference](https://docs.ray.io/en/latest/tune/api/doc/ray.tune.schedulers.AsyncHyperBandScheduler.html#ray.tune.schedulers.AsyncHyperBandScheduler)), and, at some point, I somehow ended up stepping through the [code](https://github.com/ray-project/ray/blob/f1316e8af8c8930cd062f78b1d46f7933f744bf8/python/ray/tune/schedulers/async_hyperband.py#L19) and their implementation looked surprisingly simple (yes, I know... famous last words, but after reading research papers there's sometimes this shock when you read the code and see that the implementation could have been explained in much simpler words).

[...]

PR #1178, opened 2024-12-05
train rf model in background thread

bbudescu:
work in progress on #1170

### DIFF
--- a/smac/intensifier/abstract_intensifier.py
+++ b/smac/intensifier/abstract_intensifier.py
@@ -62,7 +62,7 @@ class AbstractIntensifier:
     ):
         self._scenario = scenario
         self._config_selector: ConfigSelector | None = None
-        self._config_generator: Iterator[ConfigSelector] | None = None
+        self._config_generator: Iterator[Configuration] | None = None
         self._runhistory: RunHistory | None = None
 
         if seed is None:
--- a/smac/model/random_forest/random_forest.py
+++ b/smac/model/random_forest/random_forest.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, TYPE_CHECKING
+from threading import Thread, Condition, Event, Lock
 
 import numpy as np
 from ConfigSpace import ConfigurationSpace
@@ -11,8 +12,66 @@ from pyrfr.regression import default_data_container as DataContainer
 from smac.constants import N_TREES, VERY_SMALL_NUMBER
 from smac.model.random_forest import AbstractRandomForest
 
+if TYPE_CHECKING:
+    from pyrfr.regression import forest_opts as ForestOpts
+
 __copyright__ = "Copyright 2022, automl.org"
 __license__ = "3-clause BSD"
+
+
+class RFTrainer(Thread):
+    def __init__(self):
+        self._model: BinaryForest | None = None
+        # we could use a RWLockFair from https://pypi.org/project/readerwriterlock/, but it seems to be a bit of an
+        # overkill since critical section are rather short
+        self.model_lock = Lock()
+        self.model_available = Event()
+
+        self.data = None
+        self.opts = None
+        self.data_cv = Condition()
+
+        super().__init__(daemon=True)
+        self.start()
+
+    @property
+    def model(self):
+        self.model_available.wait()
+        with self.model_lock:
+            return self._model
+
+    def submit_for_training(self, data: DataContainer, opts: ForestOpts):
+        # use condition variable to wake up the trainer thread if it's sleeping
+        with self.data_cv:
+            assert data is not None
+            # overwrite with latest training data
+            self.data = data
+            self.opts = opts
+            self.data_cv.notify()
+
+    def run(self) -> None:
+        while True:
+            # sleep until new data is submitted for training
+            with self.data_cv:
+                while self.data is None:
+                    self.data_cv.wait()
+                data = self.data
+                self.data = None
+
+            # here we could (conditionally) call self.model_available.clear() in order to make _some_ worker threads
+            # wait for training to finish before receiving a new configuration to try, depending on CPU load; we might
+            # have to replace the Event by a Condition
+
+            _rf = regression.binary_rss_forest()
+            _rf.options = self.opts
+
+            _rf.fit(data, rng=self._rng)
+
+            with self.model_lock:
+                self._model = _rf
+
+            if not self.model_available.is_set():
+                self.model_available.set()
 
 
 class RandomForest(AbstractRandomForest):
@@ -85,7 +144,7 @@ class RandomForest(AbstractRandomForest):
         self._rf_opts.tree_opts.epsilon_purity = eps_purity
         self._rf_opts.tree_opts.max_num_nodes = max_nodes
         self._rf_opts.compute_law_of_total_variance = False
-        self._rf: BinaryForest | None = None
+        self._rf = RFTrainer()
         self._log_y = log_y
 
         # Case to `int` incase we get an `np.integer` type
@@ -142,16 +201,18 @@ class RandomForest(AbstractRandomForest):
         # self.X = X
         # self.y = y.flatten()
 
+        data = self._init_data_container(X, y)
+
         if self._n_points_per_tree <= 0:
             self._rf_opts.num_data_points_per_tree = X.shape[0]
         else:
             self._rf_opts.num_data_points_per_tree = self._n_points_per_tree
 
-        self._rf = regression.binary_rss_forest()
-        self._rf.options = self._rf_opts
+        self._rf.submit_for_training(data, self._rf_opts)
 
-        data = self._init_data_container(X, y)
-        self._rf.fit(data, rng=self._rng)
+        # call this to make sure that there exists a trained model before returning (actually, not sure this is
+        # required, since we check within predict() anyway)
+        # _ = self._rf.model
 
         return self
 
@@ -198,7 +259,9 @@ class RandomForest(AbstractRandomForest):
         if covariance_type != "diagonal":
             raise ValueError("`covariance_type` can only take `diagonal` for this model.")
 
-        assert self._rf is not None
+        rf = self._rf.model
+
+        assert rf is not None
         X = self._impute_inactive(X)
 
         if self._log_y:
@@ -207,7 +270,7 @@ class RandomForest(AbstractRandomForest):
 
             # Gather data in a list of 2d arrays and get statistics about the required size of the 3d array
             for row_X in X:
-                preds_per_tree = self._rf.all_leaf_values(row_X)
+                preds_per_tree = rf.all_leaf_values(row_X)
                 all_preds.append(preds_per_tree)
                 max_num_leaf_data = max(map(len, preds_per_tree))
                 third_dimension = max(max_num_leaf_data, third_dimension)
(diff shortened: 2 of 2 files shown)


eddiebergman:
Heyo, just a question why this is done as a thread and not a subprocess. Thread is easier by many stretches but has quite a few downsides compared to a process:

Thread:
* Will not be parallel, i.e. fights for computation of main process (it's a Python thing). It's more that it will go-back-and-forth between main process computation and this thread.
* Uses same memory as main process, easier to transfer data for training.
* Thread's can not be be interrupted and can cause the main process to not close, handling Ctrl+C or other system signals can be difficult.

Process:
* Will be parallel.
* Has seperate memory from main process, requires _sending_ data to subprocess. Usually quite cheap and handled trivially using something like the builtin `ProcessPoolExecutor`.
* System level signals propogate to subprocesses, no special handling required.

bbudescu:
> Will not be parallel, i.e. fights for computation of main process (it's a Python thing). It's more that it will go-back-and-forth between main process computation and this thread. This is due to the GIL.

Yes, that's what I was on about in https://github.com/automl/SMAC3/issues/1170#issuecomment-2519720753:

> Of course, because of the GIL, the main optimization loop gets less cpu time, but although I haven't tested, I expect that the main thread would spend most time in the training section anyway. And I don't think the other parts in the loop are very CPU intensive anyway. Or are they? Do you know if, perhaps the inference or something takes really long?

So, actually, I think with running it in a separate thread, instead of having e.g., 90% of the time of the main process doing training, we now might end up giving it only 50%, and leaving the other 50% for other stuff like reporting and inference etc. if there's a demand for that.

Also, having threads instead of processes (either through multiprocessing or dask), doesn't take away one core from the workers doing cost function evaluation. Perhaps that's not much of an advantage, though.

> Thread's can not be be interrupted and can cause the main process to not close, handling Ctrl+C or other system signals can be difficult.

I think this should already be fixed by [daemonizing]((https://github.com/automl/SMAC3/pull/1178/files#diff-b5f135e64e89aab7b4aeb65fce0d47242963bc04126cdf4085cb8b4c4b1a6f33R34)) the thread ([docs](https://docs.python.org/3/library/threading.html#threading.Thread.daemon), further [explanation](https://stackoverflow.com/questions/190010/daemon-threads-explanation)).

> Has seperate memory from main process, requires sending data to subprocess. Usually quite cheap and handled trivially using something like the builtin ProcessPoolExecutor.

Not sure how random forest serialization is handled. I know that for `Queues` objects are pickled, but I'm not sure how `concurrent.futures` pass return values. I know that `pyrfr.binary_rss_forest` exposes `ascii_string_representation()`, which can be used for serialization.

**LE:** Oh, wait... I just noticed that it also implements `__getstate__` and `__setstate__`, which rely on `ascii_string_representation` internally, so it should be pickleable.

Also, for the training data (`X` and `y` and the derived `data`), as I also wrote in https://github.com/automl/SMAC3/issues/1170#issuecomment-2519720753, one could also use shared memory arrays because they're contiguous and homogeneous `numpy` arrays, which can be buffered in `multiprocessing.Array`-like objects.

bbudescu:
> Thread's can not be be interrupted and can cause the main process to not close, handling Ctrl+C or other system signals can be difficult.

Ah, damn, I wasn't aware of https://github.com/python/cpython/issues/80116.

eddiebergman:
Heyo, some extra feedback:

* Daemonizing threads is a dangerous route too. They effectively detach themselves from the main process, and so _if_ there is a bug and the thread runs forever, well, good luck. I've also explored a lot of this in [`pynisher`](https://github.com/automl/pynisher), which is consequently what SMAC uses to run evaluations in a subprocess with the time (and memory?) limits. I just remember threads being the bane of lots of issues in the tests for that library.
* Re serialization for processes: You're indeed right that mmep of arrays is efficient and useful. Rather than re-solve the wheel, you could  use `joblib`, which is already a dependancy of scikit-learn. It already wraps a lot of the Python native ways of concurrent running of things, as well as implementing their own wrapper around `ProcessPoolExecutor` which natively handles `mmep` on objects that can be mem-mapped. It's a bit less low level, and may prevent certain operations but give the ubiqutous use of scikit-learn and how stable they've made it, it's probably best to rely on their implementations. (They also have threads fyi)

Tldr; concurrent programming in python is a mystical art with so many-caveats, and really experimentation is the only way to identify what works consistently. Try out `joblib` and if it works, I think that's the simplest solution with minimal chances for surprises later on.

bbudescu:
Hi @eddiebergman,

I didn't find a way to use `joblib` to streamline the implementation of the behavior I wanted, namely, to have a loop that continuously trains on new data as it arrives, and that always makes the current model available whenever `_predict` is called (i.g., even if another model is currently training on more recent data) without having to wait at all, so I ended up keeping the old code structure, but switched from `threading` to the builtin `multiprocessing`.

Now, the code I pushed isn't yet tested, neither manually/visually, nor by unit tests. Also, I'd like to add a flag that would allow the user to switch to the old, synchronous behavior, which I think would help test things, and also allow users to fall back on the old code until the new code is reliable enough. However, all the logic should be there (unless I made some gross mistakes), so that should give us at least a basis for discussion.

In this stage, I wanted to ask you about how the `AbstractModel._rng` member (a `np.random.RandomState` that is overwritten by a `pyrfr.regression.default_random_engine` within `RandomForest.__init__`) is supposed to be used outside of the training logic (I imagine it might be serialized when saving the experiment state?), and how to handle it best. Currently, I've basically made the main thread agnostic of the `rng` used for training (except for the seed that is fed to it on initialization). Should I pass it back to the main process through the `Queue` I'm using for models?  In this case, do you have any idea on how to handle serialization / [un-]pickling?

Also, given the non-deterministic nature of OS scheduling, running the same optimization session twice, even with the same rng seed might result in different results, especially if running with a different number of cores, but I understand that's already the case , as per the `warning` in the [Parallelism section of the Advanced Usage/Parallelism pagein the docs](https://automl.github.io/SMAC3/v2.2.0/advanced_usage/9_parallelism.html#parallelism):

> Warning: When using multiple workers, SMAC is not reproducible anymore.

Other than that, if you want to throw an eye on what I've implemented so far and, if you, perhaps, find any flagrant mistakes, any feedback and support would be greatly appreciated.

Thanks,

Bogdan

bbudescu:
Hi @eddiebergman,

I managed to bring the code to a point where it passes all the tests without hanging. Indeed, it wasn't a pretty job :).

Could you now, please, advise on what's needed for this PR to be merged?

Here are some potential improvements I could think of:

#### 1. Option to Disable Concurrency

Add a flag to revert to old implementation, and not just behavior. Note, I've already added the option to wait for training to be completed after every submission of fresh training data, but I was thinking of disabling multiprocessing altogether)

####  2. Unit Tests

Would you like me to add some unit tests? how do you think they should look like?

#### 3. Training Data IPC

In order to decrease risks, I could eliminate my own homebrew [`mmap`-backed implementation](https://github.com/bbudescu/SMAC3/blob/cb76f8c5512e4d93d7e2099a8b738cae93ccb3fc/smac/model/random_forest/multiproc_util/GrowingSharedArray.py) of a [dynamic array](https://en.wikipedia.org/wiki/Dynamic_array) and use some other way to pass the training data between the optimization loop and the training loop process.

You suggested `joblib` above, but I don't think the `futures`-like interface can really suit my need to implement a loop discarding all but the latest data (or, at least, I wasn't able to figure out how to use it to attain my goals). However, upon further investigation, I found out that, under the hood, `joblib` uses `cloudpickle` ([pypi](https://pypi.org/project/cloudpickle/), [src](https://github.com/cloudpipe/cloudpickle)) for more complex objects (like `numpy.array`s) that can't be handled by the builtin `pickle`. An alternative to `cloudpickle` would be `dill` ([pypi](https://pypi.org/project/dill/), [docs](https://dill.readthedocs.io/en/stable/), [src](https://github.com/uqfoundation/dill)), which has been around for quite some time.

I know you would prefer not to add further dependencies to the project, but I thought it doesn't hurt to ask :). I found the `multiprocess` package ([pypi](https://pypi.org/project/multiprocess/), [docs](https://multiprocess.readthedocs.io/en/stable/multiprocess.html#module-multiprocess.process), [src](https://github.com/uqfoundation/multiprocess)) that acts a drop-in replacement for the `multiprocessing` builtin module that uses `dill` for serialization.

This would also simplify the inter-process signalling required to release and unlink the shared memory, and would avoid potential leaks (see 5., below).

#### 4. RNG

Still haven't heard back from you about the `RandomForest._rng` member that overrides `AbstractModel._rng`

#### 5. Leaks, hangs

As per the code [here](https://github.com/bbudescu/SMAC3/blob/cb76f8c5512e4d93d7e2099a8b738cae93ccb3fc/tests/test_acquisition/test_maximizers.py#L483), in some complex dependency graphs, the garbage collector doesn't get to call the `__del__`'s of various objects, so execution just hangs. 

This shouldn't normally be an issue neither in the common use cases, nor when running tests with `pytest` (I needed to add code to explicitly kill the background process in the fixture teardown), but still, I think if this should be fixed if possible.

bbudescu:
Hi again,

I just added an option to disable multiprocessing altogether and switching back to the old behavior. Would this allow, perhaps, to allow the merging of the pull request, and somehow expose this to the user as an experimental reason, at his  own risk?

Thanks!

bbudescu:
I ran a session similar to the one the data from which the cpu load plot in https://github.com/automl/SMAC3/issues/1170#issue-2686007843 was generated, and also the performance reported in the motivation section of https://github.com/automl/SMAC3/issues/1169#issue-2682549690:

> I was checking out an optimization session running on 64 cores and noticed the cpu load drop below 50% after 12 hours. Not sure yet exactly how many trials it had gotten through at that point (after 22 hours it got through 12600 trials, and cpu load is under 40%).

Currently, it doesn't seem that the code I wrote has managed to fix it.

So here are some plots:

#### 1. The original 46h session
Screenshot 2024-11-23 at 11-59-52 Instances 
EC2 eu-west-1

#### 2. Another 48h session (smooth)
Screenshot 2024-12-16 at 14-49-13 Instances EC2 eu-west-1

#### 3.a. Current 24h Session
Screenshot 2024-12-18 at 08-33-17 Instances EC2 eu-west-1

#### 3.b. Current 24h Session (smooth)
Screenshot 2024-12-18 at 08-48-12 Instances EC2 eu-west-1

### Observation
After 24h the session got through 14450 trials on 64 cores, and cpu load is about 42%.

## Conclusion
Either the code is broken, or the bottleneck happens for a different reason, e.g., because of Hyperband.

**Next steps:**
- reproduce behavior quicker, i.e., without requiring a 64 core machine and wait for at least 12 hours to get a definite observation of a drop in cpu load
- do some profiling to measure on which lock most workers spend most time waiting on

bbudescu:
## Additional CPU load graphs

Hi guys. I ran another bunch of optimization sessions turning off various components of SMAC to try and identify where the bottleneck lies. Here are the results:

### 1. Hyperband disabled

#### 1.1. With Background Training
Screenshot 2024-12-21 at 11-17-23 Instances EC2 eu-west-1
6.3k trials in 48 hours

#### 1.2. Without Background Training (my smac3 clone)
Screenshot 2024-12-21 at 11-26-41 Instances EC2 eu-west-1
6.16k trials in 48 hours

#### 1.4. Without Background Training (original smac3 repo)
This was just to test if my code introduces any bottlenecks.
Screenshot 2024-12-21 at 11-31-31 Instances EC2 eu-west-1
6.24k trials in 48 hours

### 2. Random Sampling ()
- IN PROGRESS
- no Random Forest Surrogate Model

#### 2.1. Hyperband Disabled
Screenshot 2024-12-21 at 11-58-12 Instances EC2 eu-west-1
- ~ 5.5k-6k trials in 46 hours
- still running - has another 2 hours budget

#### 2.2. With Hyperband Enabled
- not started yet

PR #1208, opened 2025-02-27
#946 Purely operating Ask & Tell removes requirement for target_function

### LINKED ISSUE #946: [Feature] Remove `target_function` as required when using `ask` and `tell` interface
In the case where someone is using the `ask` and `tell` only, there's no reason for SMAC to know about the `target_function` at hand, however it's [required by the abstract facade](https://github.com/automl/SMAC3/blob/731854e8effe72d76cb07636458e783143f32dc5/smac/facade/abstract_facade.py#L98).

I can get around this pretty easily as I'm familiar enough with SMAC but it would nice not to have to do these hacky work-arounds.

```python
optimizer = HyperparameterOptimizationFacade(
    scenario=Scenario(...),
    target_function="dummy"  # Smac things this is a program it should run
)
```

The reason for this is we are building light wrappers around some optimizers that support `ask` and `tell` and **not** relying on SMAC to actually evaluate the `target_function` itself.

timruhkopf:
when using purely `ask` and `tell` instead of `optimize`, then smac does not need to know about the `target_function` and we can just drop the requirement by setting it to `None`. Should the user then however decide to call optimize, this must raise an ValueError. A few type checks and raises are necessary to ensure proper flow.

### DIFF
--- a/docs/advanced_usage/5_ask_and_tell.md
+++ b/docs/advanced_usage/5_ask_and_tell.md
@@ -15,5 +15,7 @@ and report the results of the trial.
     different budgets, they, obviously, can not be considered. However, all user-provided configurations will flow 
     into the intensification process.
 
+Notice: if you are exclusively using the ask-and-tell interface and do not use `smac.optimize()`, then smac no longer
+is responsible for the evaluation of the trials and therefor the Facade no longer will require a specified `target_algorithm` argument.
 
 Please have a look at our [ask-and-tell example](../examples/1%20Basics/3_ask_and_tell.md).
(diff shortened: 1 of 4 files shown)


benjamc (changes requested):
- [x] update changelog
- [x] Add short info to facade init about this behavior
Other than that, LGTM!

benjamc (approved):

docs/advanced_usage/5_ask_and_tell.md
```diff
@@ -15,5 +15,7 @@ and report the results of the trial.
     different budgets, they, obviously, can not be considered. However, all user-provided configurations will flow 
     into the intensification process.
 
+Notice: if you are exclusively using the ask-and-tell interface and do not use `smac.optimize()`, then smac no longer
+is responsible for the evaluation of the trials and therefor the Facade no longer will require a specified `target_algorithm` argument.
```
benjamc: small typo in therefor

benjamc (approved):

benjamc (approved):

benjamc (approved):

PR #1212, opened 2025-02-27
ask and tell without initial design

### LINKED ISSUE #1206: Documentation/Inspection ask/tell
- document that ask and tell does not care about n_trials (but the initial design does)
- should there be no initial design preset for using ask/tell?
- what happens if you use the random facade /sobol and you ask for more than the specified trials? does sobol depend on the number of trials? if it does, raise a warning

timruhkopf:
[bug-fix] when check out initial_design by setting n_configs=0  and scenario.n_trials, we will still get at least one config from the initial design on ask.

[documentation] updated how one would want to warm start with ask and tell, the meaning of n_trials (should we choose to warm start --> only relevant for max_ratio in the initial design) 
also pointed out that we can completely avoid passing target_function in the facade, if we only use ask and tell and not validate and optimize.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 - New SMAC logo
 - Fix doc link in README
 
+## Documentation
+- Ask and tell without initial design and warmstarting
+
+## Bugfixes 
+- Ask and tell without initial design may still return a config from the initial design.  
+
 # 2.3.0
 
 ## Documentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Ask and tell without initial design and warmstarting
 
 ## Bugfixes 
-- Ask and tell without initial design may still return a config from the initial design.  
+- Ask and tell without initial design may no longer return a config from the initial design - if it is not "removed".  
 
 # 2.3.0
 
--- a/examples/1_basics/3_ask_and_tell.py
+++ b/examples/1_basics/3_ask_and_tell.py
@@ -2,6 +2,10 @@
 # Flags: doc-Runnable
 
 This examples show how to use the Ask-and-Tell interface.
+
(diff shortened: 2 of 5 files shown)


examples/1_basics/3_ask_and_tell.py
```diff
@@ -2,6 +2,10 @@
 # Flags: doc-Runnable
 
 This examples show how to use the Ask-and-Tell interface.
+
+Notice, that the ask-and-tell interface will still use the initial design specified in the facade.
+Should you which to add your own evaluated configurations instead or deactivate the initial
```
benjamc: typo "which"

examples/1_basics/3_ask_and_tell.py
```diff
@@ -68,7 +72,14 @@ def train(self, config: Configuration, seed: int = 0) -> float:
         smac.tell(info, value)
 
     # After calling ask+tell, we can still optimize
-    # Note: SMAC will optimize the next 90 trials because 10 trials already have been evaluated
+    # Note: SMAC will optimize the next 90 trials because 10 trials already have been evaluated.
+    # If we however choose not to call optimize; e.g. because we want to manage heavy
+    # computation of model.train completely outside smac, but still use it to suggest new
+    # configurations, then n_trails will only be relevant for the initial design in combination
```
benjamc: typo "n_trails"

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 - New SMAC logo
 - Fix doc link in README
 
+## Documentation
+- Ask and tell without initial design and warmstarting
+
+## Bugfixes 
+- Ask and tell without initial design may still return a config from the initial design.
```
benjamc: reformulate bugfixed notice to what it fixed

benjamc (approved):

Issue #1118, opened 2024-06-24
Investigate behavior of intensifier


benjamc:
`n_retries`
why needed, logic flow

Issue #1088, opened 2023-12-06
SMAC stops before n_trials


hvarfner:
A rather basic question, possibly related to previous issue by mfeurer regarding `n_retries`:

Using the following setup (and latest pip version, 2.0.2):
```
initial_design = SobolInitialDesign(scenario, n_configs=10)

intensifier = Intensifier(scenario, max_config_calls=1, retries=500)

smac = HyperparameterOptimizationFacade(
    scenario, 
    recorded_trajectory.__call__, 
    initial_design=initial_design, 
    intensifier=intensifier,
    overwrite=True,
)

smac.optimize()
```
for a fully categorical, slightly constrained (one forbidden-clause which excludes ~20% of the search space of 300k configurations), SMAC stops after 64 iterations with the following message:

```
INFO][abstract_intensifier.py:590] Added config d89cda and rejected config 9e4bfe as incumbent because it is not better than the incumbents on 1 instances:
[INFO][smbo.py:319] Finished 50 trials.
[INFO][abstract_intensifier.py:590] Added config c60899 and rejected config d89cda as incumbent because it is not better than the incumbents on 1 instances:
[WARNING][config_selector.py:242] Could not return a new configuration after 16 retries.
[INFO][smbo.py:332] Shutting down because the stop flag was set.
```

Why is this, and can it be disabled so that the run can finish? I tried to examine the `Intensifier` but was unable to find the solution myself.

Any help would be much appreciated! 
Best,
Carl

mfeurer:
Hey @hvarfner, yes, I think this is highly related to my previous issue #1086. I agree that it would be great if SMAC does not decide by itself to stop the Bayesian optimization.

hvarfner:
@mfeurer Great, thanks for the clarification! Do you know if this can be disabled? I'm perfectly happy hacking around in the source code if needed.

mfeurer:
No, sorry, I don't know if this can be disabled or where the number of retries can be increased.

hvarfner:
Okay, thanks! I'll look at it on my end and let you know if I find a simple way around it.

alexandertornede:
Hi @hvarfner, 

sorry for the late reply, but we had some people being absent during December and there was the Christmas break.

[...]

Issue #1086, opened 2023-11-28
What is the reasoning behind n_retries


mfeurer:
The new SMAC rewrite introduced an argument [n_retries](https://github.com/automl/SMAC3/blob/541ee7e0383b491b86d1a23dcff669f2efad616d/smac/main/config_selector.py#L54), however, it is not clear to me why it was introduced. Could you please explain the reasoning why it is there?

I am asking because SMAC decided to stop an evaluation despite the budget not being exhausted. However, n_retries is not something that is exposed in the Facade, not something that is researched or described, and therefore very surprising to me.

PR #1216, opened 2025-03-14
Better handling of n_retries

hadarshavit:
Related issues: #1088  #1086 #1118 

Better handling of n_retries: if the first n_retires configs returned from the acq maximizer were already sampled, sample random configs instead, and check that those configs were not sampled before. If this also fails, then simply sample random configuration without checking if they were sampled.

This handling is inspired by HEBO: https://github.com/huawei-noah/HEBO/blob/2cce0838dc16b53aa8955f3ead025268ddd18a00/HEBO/hebo/optimizers/hebo.py#L170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # 2.3.1
 
+## Bugfixes
+- Addressing situations where the acquisition function suggests configurations that have already been sampled in prior iterations (#1216)
+
 ## Misc
 - New SMAC logo
 - Fix doc link in README
--- a/smac/main/config_selector.py
+++ b/smac/main/config_selector.py
@@ -241,7 +241,31 @@ class ConfigSelector:
                         logger.warning(
                             f"Could not return a new configuration after {self._max_new_config_tries} retries." ""
                         )
-                        return
+                        random_configs_retries = 0
+                        while counter < self._retrain_after and random_configs_retries < self._max_new_config_tries:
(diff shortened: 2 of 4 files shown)


daphne12345:
Hello @hadarshavit, thank you for the effort! It makes sense to me to sample a random configuration if n_retries is up but I don't understand the second part. Why do you want to run a configuration without knowing if it has been processed before? Isn't it likely that you sample a configuration that has already been run before?

hadarshavit:
Hi @daphne12345, the idea is that if the user still wants to run a configuration (for example if there is a call to ask), and all other configurations were already processed, SMAC should still returns some configuration (because the user asked for one). This is the same mechanism HEBO uses, but also raising an exception can make sense in such case.

benjamc:
Hi Hadar, 
we are still stumbling about the second block. Assuming we tried every configuration possible, shouldn't we then stop the optimization? I agree that raising an exception could be the way to go here.
(The first block absolutely makes sense)

hadarshavit:
I added an exception in case even the random sampling fails and a test for this case.

daphne12345:
Hello Hadar, thank you for the update. I'm wondering why the ConfigurationSpaceExhautionException is only raised if random_configs_retries < self._max_new_config_tries. Should it not be raised in case we reach the max_new_config_tries also with the random_config_retries?

benjamc (approved):

Issue #1191, opened 2025-01-13
Add bool in model indicating whether it is trained


benjamc:
`is_not_trained = isinstance(self._model, GaussianProcess) and not self._model._is_trained \
                or isinstance(self._model, RandomForest) and self._rf is None`

benjamc:
add `model_trained` as property to surrogate model abstract class (or `is_model_trained()` function)

timruhkopf:
PR #1232 closes this

PR #1232, opened 2025-04-17
[feature] #1191 is trained property to indicate the state of the model.

timruhkopf:
#1191

### DIFF
--- a/tests/test_model/test_abstract_model.py
+++ b/tests/test_model/test_abstract_model.py
@@ -23,6 +23,16 @@ def get_X_y(cs, n_samples, n_instance_features):
 def _train(X, Y):
     return None
 
+class MyAbstractModel(AbstractModel):
+    # dummy class to ensure that the required methods are implemented
+    @property
+    def is_trained(self):
+        return self._is_trained
+
+    def _train(self, X, Y):
+        self._is_trained = True
+        return self
+
 
 def test_no_pca(configspace_small, make_scenario):
     n_instances = 100
@@ -35,7 +45,7 @@ def test_no_pca(configspace_small, make_scenario):
         n_instances=n_instances,
         n_instance_features=n_instance_features,
     )
-    model = AbstractModel(configspace_small, scenario.instance_features, pca_components=7)
+    model = MyAbstractModel(configspace_small, scenario.instance_features, pca_components=7)
     # We just overwrite the function as mock here
     model._train = _train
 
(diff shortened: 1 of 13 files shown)


benjamc (changes requested):
Can you make sure that the property `is_trained` is an abstract property, thus required?
If this is difficult with the property type, you can also create an `abstractmethod`

timruhkopf:
> Can you make sure that the property `is_trained` is an abstract property, thus required? If this is difficult with the property type, you can also create an `abstractmethod`

I have inherited from ABC and made it an abstract property. Now we have a few property getters that will only return if the "_is_trained" attribute is True or False. But future work may want to access that differently, so it kind of makes sense.

tests/test_model/test_abstract_model.py
```diff
@@ -88,3 +98,36 @@ def test_pca(configspace_small, make_scenario):
...
+
+    @pytest.fixture
+    def configspace():
+        return ConfigurationSpace({
+            "cat": [True, False, None],
+            "othercat": [A(1), f],
+            "constant": Constant("constant", (24, 25)),
+        })
+    try:
+        model = DummyModel(configspace_small)
```
benjamc: is `configspace` used here /  what is `configspace_small`?

benjamc (changes requested):

tests/test_model/test_abstract_model.py
timruhkopf: it is merely a [fixture ](https://github.com/automl/SMAC3/blob/9d194754a5fed3ec48be06987cfc24ee99b76af5/tests/fixtures/configspace.py#L13) that i need to attempt instantiating the model. As long as it is a ConfigurationSpace object it doesn't really matter much what it contains.

benjamc: yes therefore in that case we do not need the redefinition, right?
and the import statements should go outside the function def (unless there is a reason which I do not see rn)

in addition, what is the expected error? specifically test for this with sth similar to L99  
```python
with pytest.raises(ValueError, match="Feature mismatch.*"):
        model.predict_marginalized(X_test)
```

timruhkopf: huups that snuck in there :)

timruhkopf: was from my if name main first demo that i copied over into the test :disappointed:

PR #1236, opened 2025-06-12
Feature/batch bo issue 1229 (QEI)

daphne12345:
I am not sure if it is correctly implemented. Please check!


smac/main/config_selector.py
```diff
@@ -325,6 +365,89 @@ def _get_evaluated_configs(self) -> list[Configuration]:
...
+            return self._model.predict_marginalized(X_running)[0]  # type: ignore[union-attr]
+        elif estimation_strategy == "sample":
+            # https://papers.nips.cc/paper_files/paper/2012/file/05311655a15b75fab86956663e1819cd-Paper.pdf
+            # since this requires a multi-variant gaussian distribution for the candidates, we need to restrict the
+            # model to be a gaussian process
+            assert isinstance(
+                self._model, GaussianProcess
+            ), "Sample based estimate strategy only allows GP as surrogate model!"
+            return self._model.sample_functions(X_test=X_running, n_funcs=1)
+        elif estimation_strategy == "q_ei":
```
dengdifan: Hmm... I think if we use `q_ei`, there is no need to estimate the running configurations. i.e., we use `q_ei` to provide multiple candidates at once.

smac/acquisition/function/expected_improvement.py
```diff
@@ -286,3 +286,99 @@ def calculate_f() -> np.ndarray:
...
+            std_copy = np.copy(std)
+            std[std_copy == 0.0] = 1.0  # prevent division by zero
+
+        # Monte Carlo sampling from log-normal distribution
+        normal_samples = np.random.normal(loc=m.T, scale=std.T, size=(self.n_samples, X.shape[0]))
+
+        if not self._log:
+            f_samples = normal_samples  # in original (normal) space
+            f_min_sample = np.min(f_samples, axis=1)
+            improvement = np.maximum(self._eta - self._xi - f_min_sample, 0.0)
```
dengdifan, line 369: Not quite sure here, but if the surrogate model is a GP, we can directly sample from the joint posterior distribution. As shown in botorch: https://botorch.readthedocs.io/en/stable/acquisition.html#botorch.acquisition.monte_carlo.qExpectedImprovement

smac/main/config_selector.py
daphne12345: I removed this part.

smac/acquisition/function/expected_improvement.py
daphne12345: I replaced it with the surrogate model, if it is a GP.

PR #1274, opened 2025-11-11
1184 prioracquisitionfunction adapts sampling

LukasFehring:


LukasFehring:
I slightly adapted the random design to use a uniform copy of the configuration space. To that end i added functionality to make the configspace uniform, and tested it

daphne12345 (approved):

PR #1276, opened 2025-11-13
966 output trafos

lhennig0103:


mfeurer:
Hey, is there a reason why the scientific source for the Copula transform in Bayesian optimization is not cited in the PR? I think it would be great to mention the paper by [Salinas et al.](http://proceedings.mlr.press/v119/salinas20a.html). Sorry for only bringing this up now, but I haven't been following github repositories lately.

PR #1279, opened 2025-11-28
1264 bug reset after initialization

LukasFehring:


examples/1_basics/3_ask_and_tell.py
```diff
@@ -71,6 +71,9 @@ def train(self, config: Configuration, seed: int = 0) -> float:
 
         smac.tell(info, value)
 
+    smac.optimizer.reset()
+    #smac.optimizer.reset()
```
mwever: ?

mwever (commented):
Except for the example which has commented code and the same code line again uncommented. Not sure whether in the example there should be a reset at all.

examples/1_basics/3_ask_and_tell.py
LukasFehring: Thanks for pointing that out. Yes ithese lines should have been deleted before pushing :)

mwever (approved):

PR #1296, opened 2026-02-24
added a warning when n_trials budget is depleted but ask still return trials

### LINKED ISSUE #1197: Ask and Tell continues providing configuraitons after n_trilas is exceeded
#### Description
Even after `n_trials` is depleted, the `ask and tell` interface of SMAC allows to sample additional configurations without notifying the user. This behavior warrants documentation, and probably a warning.

sagar1sharma:
…s trials

> Thanks for contributing a pull request! Please ensure you have taken a look at the contribution guidelines: https://github.com/automl/SMAC3/blob/main/CONTRIBUTING.md

Previously, ask() could continue returning trials after budget exhaustion without a strict guard in ask/tell usage, with stopping primarily enforced in the optimize() loop.
This PR adds explicit post-budget handling in ask() (warning/exception modes), updates docs, and adds tests for the same.

Added dedicated exception class: AskAndTellBudgetExhaustedError.
Added/updated warn mode plumbing through facade to optimizer.
Updated ask-and-tell docs to describe post-budget behavior and warn modes.
Added/updated tests in both:
tests/test_ask_and_tell/test_ask_and_tell_intensifier.py
tests/test_ask_and_tell/test_ask_and_tell_successive_halving.py

[x] New feature (non-breaking change which adds functionality)
[ ] Breaking change (fix or feature that would cause existing functionality to change)

[x] My change required a change of the documentation, which has been done.
[ ] I checked that the documentation can be build, visualizes everything as expected, and does not contain any warnings.
[x] I have added/adapted tests to cover my changes.
[x] The tests can be executed successfully.
[ ] I have added a description of the changes to CHANGELOG.rst.


kalifou:
Status update about the PR:

- The acurate reference Issue is Nr.1197 #1197

> #### Reference Issues/PRs
> 
> > Fixes: #1296

-  Updated checklist of remaining todos:
 [V] Improve the ask_and_tell code example
 [V ] add a description of the changes to CHANGELOG.rst.
 [V ] fix the formatting of the codebase and docstring
 [V] branch is outdated: merge main/dev into current branch
 [V ] check that the updated branch passes all tests
 [V] checked that the documentation can be built, visualizes everything as expected, and does not contain any warnings.

> #### Checklist
> 
> > [x] My change is based on the latest stage of the develop branch.
> > [x] My change required a change of the documentation, which has been done.
> > [ ] I checked that the documentation can be built, visualizes everything as expected, and does not contain any warnings.
> > [x] I have added/adapted tests to cover my changes.
> > [x] The tests can be executed successfully.
> > [ ] I have added a description of the changes to CHANGELOG.rst.

smac/facade/abstract_facade.py
```diff
@@ -43,6 +44,39 @@
 __license__ = "3-clause BSD"
 
 
+class AskExhaustedWarnMode(str, Enum):
+    WARN_ONCE = "warn_once"
```
daphne12345: I think it is a bit strange to have another class in this file. I would suggest to convert this to a function and also change the name, because normalize might be misleading, maybe to validate_warn_string().

smac/main/smbo.py
```diff
@@ -76,10 +78,17 @@ def __init__(
         self._runner = runner
         self._overwrite = overwrite
 
+        allowed_warn_modes = {"warn_once", "warn_never", "warn_always", "exception"}
```
daphne12345: The same code should not be copied in two places.

smac/main/smbo.py
```diff
@@ -106,7 +115,9 @@ def intensifier(self) -> AbstractIntensifier:
     @property
     def remaining_walltime(self) -> float:
         """Subtracts the runtime configuration budget with the used wallclock time."""
```
daphne12345, line 119: Why did you change it to this? Is this function never used anywhere else?

smac/main/smbo.py
kalifou: This is dictionnary is now stored in the main constant file of smac and imported every time as:

from smac.constants import ASK_AND_TELL_VALID_WARN_MODES

smac/facade/abstract_facade.py
kalifou: okay, done.
now in a dedicated util file and can be imported like this:

from smac.utils.ask_and_tell import validate_warn_mode

smac/main/smbo.py
kalifou: Thanks, this was added by the author of the initial PR draft. From my understanding, the assumption of the main branch is that you do not need to call this function remaining_walltime(), outside of the general optimize() function, which sets internally the value of self._start_time to time.time(). So you would not encounter self._start_time being None before.

This conflicts with the ask-and-tell branch, where remaining_walltime() is now also called outside of optimize(), facing self._start_time possibly never defined and None. I have reverted this change, and now define self._start_time in the __init__() and also in reset() functions. 

Let me know if it is better now

daphne12345 (approved):

Issue #954, opened 2023-03-30
Implement parallel optimization of acquisition function


benjamc:
Optimization takes time, could be improved by that.
Emerged from this issue: #774

benjamc:
Further improvements as proposed by @mfeurer 
https://github.com/automl/SMAC3/issues/501#issue-418761985

> 1. Use configurations with highest EI from the random search happening before
> 
>     2. Scale standard deviation of neighborhood sampling as in [Regis and Shoemaker (2013)](https://www.researchgate.net/publication/254230327_Combining_radial_basis_function_surrogates_and_dynamic_coordinate_search_in_high-dimensional_expensive_black-box_optimization)
> 
>     3. Use multiple-coordinate local search instead of single-coordinate.
> 
>     4. When doing plateau walking, there are multiple hyperparameters which do not need to be searched any further: Categorical and Ordinal, Integer with less possible values than requested neighbors, and potentially Integers with only a few legal values (for example, an integer with 9 legal neighbors which is queried for 8 neighbors does probably not need to be queried more than twice (right now it is queried 11 times)).

benjamc:
- [ ] check PR of Jeroen: acquisition maximizer diff
https://github.com/automl/SMAC3/pull/1222/files

this could speed up the maximization, roughly test that

benjamc:
- [ ] find out from which task dimensionality it makes sense to parallelize
- [ ] when everything is ready: benchmark following the steps in SMAC3-benchmarking

PR #1298, opened 2026-02-25
Add batch parallelization to local search #954

JannisKst:
This PR introduces batch parallelization to LocalSearch.

- Add user-tunable `n_jobs` parameter to control number of parallel workers
- Split starting points into `n_jobs` batches so each worker performs search on one batch
- Handle all edge-case `n_jobs` values (negative, zero, or too large)

These changes were benchmarked in the [SMAC3-Benchmarking](https://github.com/automl/SMAC3-benchmarking/tree/954-acqmax/experiments/954-acqmax) repository, demonstrating improved runtime and consistent acquisition performance across multiple experiments.


mwever (commented):
Looks nice and clean

PR #1303, opened 2026-03-05
Add adaptive stdev scaling to local search neighborhood sampling #954

JannisKst:
This PR adds adaptive scaling of the neighborhood sampling standard deviation for continuous hyperparameters based on local search progress.

The scaling follows these rules:
- Each local search starts with an initial stdev of 0.05.
- After 3 consecutive improvements, the stdev is doubled, up to a maximum of stdev_init * 8.
- After every third plateau walk, the stdev is halved.

These changes were benchmarked in the [SMAC3-Benchmarking](https://github.com/automl/SMAC3-benchmarking/tree/954-acqmax/experiments/954-acqmax) repository. The repository also contains additional details explaining the design and implementation choices.

### DIFF
--- a/smac/acquisition/maximizer/local_search.py
+++ b/smac/acquisition/maximizer/local_search.py
@@ -53,7 +53,18 @@ class LocalSearch(AbstractAcquisitionMaximizer):
         reduce the overhead of SMAC.
     seed : int, defaults to 0
     n_jobs: int, defaults to 1
-        Number of parallel workers to use for local search evaluation
+        Number of parallel workers to use for local search evaluation.
+    stdev_init: float, defaults to 0.05
+        Initial standard deviation used when sampling neighbors for continuous hyperparameters during local search.
+    stdev_min: float, defaults to 5e-3
+        Minimum allowed standard deviation for neighborhood sampling.
+    stdev_max: float | None, defaults to None
+        Maximum allowed standard deviation for neighborhood sampling.
+        If None, the value is automatically set to ``stdev_init * 2**downscale_interval``.
+    upscale_thresh: int, defaults to 3
+        Number of consecutive improvements required before increasing the standard deviation.
+    downscale_interval: int, defaults to 3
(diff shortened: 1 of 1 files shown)


smac/acquisition/maximizer/local_search.py
```diff
@@ -318,6 +318,13 @@ def _search(
 
         hp_names = list(start_points[0].config_space.keys())
 
+        # Initial standard deviation used when sampling continuous neighbors.
```
mwever: Please, no magic constants in our code :-)

Can you please add these parameters as actual parameters for this class? You can still set default values, no problem, but these parameters should not just simply pop up within an internal method.

smac/acquisition/maximizer/local_search.py
```diff
@@ -506,6 +519,17 @@ def _search(
...
 
                 if new_neighborhood[i]:
+
+                    if improved[i]:
+                        improvement_count[i] += 1
+                    else:
+                        improvement_count[i] = 0
+
+                    # Increase exploration radius if the search improves consistently.
+                    if improvement_count[i] >= 3 and n_no_plateau_walk[i] == 0:
```
mwever: Also here please no magic constants in between. The limit for scaling out or zooming in should be a parameter of this class.

smac/acquisition/maximizer/local_search.py
```diff
@@ -514,13 +538,19 @@ def _search(
                             for s in visited_values[i].values():
                                 s.clear()
                         n_no_plateau_walk[i] += 1
+                        # Reduce exploration radius during plateau walking to refine the
+                        # search locally around the current candidate.
+                        if n_no_plateau_walk[i] % 3 == 0:
```
mwever: See above

mwever (commented):
Please fix the comments made below.

mwever (approved):
LG2M

PR #1308, opened 2026-04-21
Refactor RunHistory to separate data storage from cost computation

JannisKst:
Refactors RunHistory into a pure data container by removing all cost computation logic.
Cost aggregation, normalization, and scalarization are now handled by CostTransformer.
Existing code has been updated to use raw cost retrieval + transformation, and new unit tests for CostTransformer have been added


mwever (approved):
LG2M

PR #1339, opened 2026-08-06
Validate instance features for non-finite values before optimization

JannisKst:
This PR validate instance features when creating a `Scenario`.

Previously, non-finite instance features (e.g. `NaN` or `±inf`) were only detected later by downstream preprocessing or surrogate models, often resulting in cryptic warnings or errors. Now, SMAC validates instance features upfront and raises a clear `ValueError` if non-finite values are present, allowing users to correct their data before the optimization starts.


mwever (approved):
Looks good.

Issue #1314, opened 2026-06-23
[BUG] API for scikit-learn outdated in code


benjamc:
#### Description
Scikit-learn API in code is outdated with newest scikit-learn version (1.9).

#### Steps/Code to Reproduce

Python 3.13., uv, SMAC version 2.4.0
Installing SMAC from source (main branch).

Running: `python examples/1_basics/1_quadratic_function.py` returns following error code:

```
Traceback (most recent call last):
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/examples/1_basics/1_quadratic_function.py", line 16, in <module>
    from smac.facade.hyperparameter_optimization_facade import HyperparameterOptimizationFacade as HPOFacade
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/smac/__init__.py", line 27, in <module>
    from smac.callback.callback import Callback
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/smac/callback/__init__.py", line 2, in <module>
    from smac.callback.metadata_callback import MetadataCallback
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/smac/callback/metadata_callback.py", line 9, in <module>
    from smac.main.smbo import SMBO
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/smac/main/smbo.py", line 13, in <module>
    from smac.acquisition.function.abstract_acquisition_function import (
        AbstractAcquisitionFunction,
    )
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/smac/acquisition/function/__init__.py", line 9, in <module>
    from smac.acquisition.function.prior_acquisition_function import (
        PriorAcquisitionFunction,
    )
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/smac/acquisition/function/prior_acquisition_function.py", line 18, in <module>
    from smac.model.random_forest.abstract_random_forest import AbstractRandomForest
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/smac/model/random_forest/__init__.py", line 2, in <module>
    from smac.model.random_forest.random_forest import RandomForest
  File "/home/carolin/Documents/repos/CARP-S/lib/SMAC3/smac/model/random_forest/random_forest.py", line 14, in <module>
    from sklearn.tree._tree import DTYPE
[...]

Issue #1306, opened 2026-04-21
[bug] Some generated configs don't satisfy forbiddens


bbudescu:
I think this happens because when sampling the initial configs from a `RandomInitialDesign`, `configspace.create_uniform_configspace_copy` is called to instantiate a modified copy of the search space (which replaces each hyperparameter with uniform versions thereof), while conditions are [explicitly set](https://github.com/automl/SMAC3/blob/b8f10493df802def001ec32c666ed03533f3bb06/smac/utils/configspace.py#L280) on the newly created space, setting forbiddens is omitted.

mwever:
Resolved by #1315

PR #1347, opened 2026-09-09
V2.4.1

LukasFehring:
# 2.4.1
## Improvements
- Include a new multi-objective method based on hypervolume
- Added configurable `warn_mode` for ask-and-tell budget exhaustion behavior (#1197)
- Batch model predictions when selecting the best predicted configuration.

## Examples
- An example on the new multi-objective method

## Bugfixes
- Scikit-learn 1.9 deprecated an alias for np.float32 called DTYPE. Change removes references to deprecated alias (#1314)
- Fix `create_uniform_configspace_copy` not copying forbidden clauses, causing `RandomInitialDesign` to sample configs that violate the original search space's constraints (#1306)
- Tests were failing due to an incompatibility between pytest and pytest-cases.  Pytest-cases is a dead stub, so simplest fix is to remove it.


smac/main/config_selector.py
```diff
@@ -215,26 +261,31 @@ def __iter__(self) -> Iterator[Configuration]:
...
             challengers = self._acquisition_maximizer.maximize(
                 previous_configs,
+                # n_points=self._retrain_after, #TODO MERGE check
                 random_design=self._random_design,
             )
 
+            if self._retrain_wallclock_ratio is not None:
+                # TODO: CB: What does this actually do? Delete/clear the iterator?
+                #  --> JG: To easily measure the time needed to perform maximise, difficult otherwise due to yield
+                len(list(challengers))  # Forces actual computation of the acquisition function maximizer
```
kalifou, line 271: Hey!

Generally looks good. Minor question: are we sure that line 271 works as intended?

Thanks

kalifou (approved):
LG2M

